PR #237, opened 2014-06-12
Minor amendment

bafco:
\+ update gitignore


antoinesd:
The phase you corrected has to be clarified, but as a decorator is not tied to a class but to a bean (type and qulifier) your correction is also wrong. Please fill a ticket in JIRA to discuss the right wording with EG.

PR #241, opened 2015-04-14
CDI-160 : Introducing SE and EE part

antoinesd:
This is the fist step of spliting the spec between Core and Java EE in CDI spec.
New parts have been created. All EJB content has bee removed from Core part to go to Java EE.

There're are still ref to Java EE in part regarding EL, Servlet/JSP and JSF.


spec/src/main/doc/implementation.asciidoc
```diff
@@ -899,7 +661,7 @@ If any other injection point declares a +@Named+ annotation that does not specif
 
 _The @New qualifier was deprecated in CDI 1.1. CDI applications are encouraged to inject @Dependent scoped beans instead._
 
-For each managed bean, and for each session bean, a second bean exists which:
+For each bean, a second bean exists which:
```
jharting: For each *managed* bean

jharting:
In many places an expression "managed bean or session bean" is changed to "bean". This is of course incorrect. Instead of reducing the target set to managed beans only the "bean" expression also references producer methods, producer fields and perhaps extension-provided beans.

spec/src/main/doc/injectionelresolution.asciidoc
```diff
@@ -346,23 +346,14 @@ The behavior of all methods declared by +java.lang.Object+, except for +toString
...
-From time to time the container instantiates beans and other Java EE component classes supporting injection. The resulting instance may or may not be a _contextual instance_ as defined by <<contextual_instance>>.
+From time to time the container instantiates beans and other component classes supporting injection. The resulting instance may or may not be a _contextual instance_ as defined by <<contextual_instance>>.
 
-The container is required to perform dependency injection whenever it creates one of the following contextual objects:
+The container is required to perform dependency injection whenever it creates one of the following objects:
 
-* contextual instances of session beans, and
-* contextual instances of managed beans.
+* contextual instances of managed beans, and
+* non-contextual instances of managed beans.
```
jharting: what is a non-contextual managed bean?

spec/src/main/doc/injectionelresolution.asciidoc
```diff
@@ -384,16 +375,16 @@ When the container creates a new instance of a managed bean, session bean, or of
...
-* Any +@PostConstruct+ callback declared by a class X in the type hierarchy of the bean is called after all initializer methods declared by X or by superclasses of X have been called, after all injected fields declared by X or by superclasses of X have been initialized, and after all Java EE component environment resource dependencies declared by X or by superclasses of X have been injected.
-* Any servlet +init()+ method is called after all initializer methods have been called, all injected fields have been initialized and all Java EE component environment resource dependencies have been injected.
+* Initializer methods declared by a class X in the type hierarchy of the bean are called after all injected fields declared by X or by superclasses of X have been initialized, and after all component environment resource dependencies declared by X or by superclasses of X have been injected.
```
jharting: Is the container supposed to support resource injection or not? The removal of resource producers seems to indicate that not but this wording seems to indicate that it is supported.

spec/src/main/doc/injectionelresolution.asciidoc
```diff
@@ -384,16 +375,16 @@ When the container creates a new instance of a managed bean, session bean, or of
...
+* Any +@PostConstruct+ callback declared by a class X in the type hierarchy of the bean is called after all initializer methods declared by X or by superclasses of X have been called, after all injected fields declared by X or by superclasses of X have been initialized, and after all component environment resource dependencies declared by X or by superclasses of X have been injected.
+* Any servlet +init()+ method is called after all initializer methods have been called, all injected fields have been initialized and all component environment resource dependencies have been injected.
```
jharting: Should be removed from the SE part probably

spec/src/main/doc/javaee/Interceptors_ee.asciidoc
```diff
@@ -0,0 +1,11 @@
...
+
+== Interceptor bindings in Java EE
+
+EJB session and message-driven beans support interception as defined in <<interceptors>>.
+
+[[interceptor_resolution_ee]]
+
+=== Interceptor resolution in Java EE
+
+For a custom implementation of the +Interceptor+ interface defined in <<interceptor>>, the container also calls +intercepts()+ to determine if the interceptor intercepts a given EJB timeout.
```
jharting: to determine if the interceptor intercepts an EJB timeout method invocation

spec/src/main/doc/javaee/implementation_ee.asciidoc
```diff
@@ -0,0 +1,297 @@
...
+
+Declaring a producer field in an EJB session bean follow the rules defined in <<declaring_producer_field>> with the following addition:
+
+* If a non-static field of an EJB session bean class is annotated +@Produces+, the container automatically detects the problem and treats it as a definition error.
+
+[[disposer_method_ee]]
+
+=== Disposer methods on EJB session bean
+
+A disposer method defined in an EJB session bean follow the rules defined in <<disposer_method> with the following addition:
```
jharting: follows

spec/src/main/doc/spi.asciidoc
```diff
@@ -241,9 +241,7 @@ public interface CDIProvider {
 }
 ----
 
-A Java EE container is required to provide a CDI provider that will allow access to the current container for any Java EE application or Java EE module which contains enabled beans.
-
-Java EE components may obtain an instance of +BeanManager+ from JNDI by looking up the name +java:comp/BeanManager+.
+Components may obtain an instance of +BeanManager+ from JNDI by looking up the name +java:comp/BeanManager+.
```
jharting: This should stay as "Java EE components" and the entire sentence should be moved to the Java EE part

spec/src/main/doc/injectionelresolution.asciidoc
antoinesd: Original text was:
>The container is required to perform dependency injection whenever it creates one of the following contextual objects:
>
>* contextual instances of session beans, and
>* contextual instances of managed beans.
>
>
>The container is also required to perform dependency injection whenever it instantiates any of the following non-contextual objects:
>
>* non-contextual instances of session beans (for example, session beans obtained by the application from JNDI or injected using +@EJB+),
>* non-contextual instances of managed beans, and
>* instances of any other Java EE component class supporting injection.

it became this text (didn't invented the *non-contextual instances of managed beans* concept, it's introduced in [InjectionTarget](http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#bm_obtain_unmanaged_instance) section) and this text in Java EE part:

>When running in Java EE, the container must extend the rules defined in <<injection>> and is also required to perform dependency injection whenever it creates one of the following objects:
>
>* contextual instances of EJB session beans,
>* non-contextual instances of EJB session beans (for example, session beans obtained by the application from JNDI or injected using +@EJB+), and
>* instances of any other Java EE component class supporting injection.

I split the list of non-managed between Core and EE, but perhaps lost something in the operation...

spec/src/main/doc/core/lifecycle.asciidoc
```diff
@@ -10,11 +10,11 @@ Every bean in the system is represented by an instance of the +Bean+ interface d
...
 
-If the application requires more control over instantiation of a contextual instance, a producer method or field may be used. Any Java object may be returned by a producer method or field. It is not required that the returned object be a contextual reference for a bean. However, if the object is not a contextual reference for another bean, the object will be contextual instance of the producer method bean, and therefore available for injection into other objects and use in EL expressions, but the other capabilities listed in <<capabilities>> will not be available to the object.
+If the application requires more control over instantiation of a contextual instance, a producer method or field may be used. Any Java object may be returned by a producer method or field. It is not required that the returned object be a contextual reference for a bean. However, if the object is not a contextual reference for another bean, the object will be contextual instance of the producer method bean, and therefore available for injection into other objects and use in name resolution, but the other capabilities listed in <<capabilities>> will not be available to the object.
```
tremes: Missing "will" in: "It is not required that the returned object be a contextual reference for a bean."

spec/src/main/doc/core/scopescontexts.asciidoc
```diff
@@ -170,10 +169,10 @@ The +@Dependent+ scope is always active.
 
 ==== Dependent objects
 
-Many instances of beans with scope +@Dependent+ belong to some other bean or Java EE component class instance and are called _dependent objects_.
+Many instances of beans with scope +@Dependent+ belong to some other bean are called _dependent objects_.
```
tremes: I dislike this sentence somehow. Shouldn't be there "belonging" instead of "belong" ?

spec/src/main/doc/core/spi.asciidoc
```diff
@@ -1062,9 +1047,9 @@ If any +ProcessInjectionPoint+ method is called outside of the observer method i
 
 ==== +ProcessInjectionTarget+ event
 
-The container must fire an event for every Java EE component class supporting injection that may be instantiated by the container at runtime, including every managed bean declared using +@ManagedBean+, EJB session or message-driven bean, bean, interceptor or decorator.
+The container must fire an event for every component class supporting injection that may be instantiated by the container at runtime, including every managed bean declared using +@ManagedBean+, bean, interceptor or decorator.
```
tremes: What about to define FQN of ManagedBean annotation? Just for sure...

spec/src/main/doc/javaee/implementation_ee.asciidoc
```diff
@@ -0,0 +1,319 @@
...
+* Session beans
+* Resources (Java EE resources, persistence contexts, persistence units, remote EJBs and web services)
+
+Java EE and embeddable EJB containers are required by the Java EE and EJB specifications to support EJB session beans and the Java EE component environment. Other containers are not required to provide support for injection or lifecycle management of session beans or resources.
+
+[[what_classes_are_beans_ee]]
+
+==== Which Java classes are managed beans in Java EE?
+
+When running in Java EE, A top-level Java class is a managed bean if it meets requirement describe in <<what_classes_are_beans>> and if it is defined to be a managed bean by any other Java EE specification and if
```
tremes: "if it meets requirement described". Missing d in describe

spec/src/main/doc/core/scopescontexts.asciidoc
antoinesd: The sentence is bad, you're right. But it's because I accidentally removed "and". It should be:

> Many instances of beans with scope +@Dependent+ belong to some other bean __and__ are called _dependent objects_.

spec/src/main/doc/core/lifecycle.asciidoc
antoinesd: Didn't change this sentence. I don't know why Gihtub shows modification.Anyway this "mistake is in the original text (so it's out of scope of this split) and I'm not sure it's an error. We should check with a native english speaker.

spec/src/main/doc/core/spi.asciidoc
antoinesd: Out of scope for this ticket. But your remark should be noted in an other ticket. It's the only mention of `@ManagedBean` in the whole spec. Strange...

spec/src/main/doc/core/injectionandresolution.asciidoc
```diff
@@ -281,27 +279,26 @@ A bean must declare _all_ of the qualifiers that are specified at the injection
...
 
-The process of matching a bean to a name used in EL is called _name resolution_. Since there is no typing information available in EL, the container may consider only the bean name. Name resolution usually occurs at runtime, during EL expression evaluation.
+The process of matching a bean to a name is called _name resolution_. Since there is no typing information available during name resolution, the container may consider only the bean name. Name resolution usually occurs at runtime.
 
-An EL name resolves to a bean if:
+A name resolves to a bean if:
 
-* the bean has the given bean name, and
-* the bean is available for injection in the war containing the JSP or JSF page with the EL expression.
+* the bean has the given bean name.
```
jharting: We should not eliminate the accessibility requirement. Now that we moved from EL name resolution to a more general "name resolution" the accessibility requirement should not only stay but should be made more general, for example:
"the bean is available for injection in the given module"

spec/src/main/doc/core/injectionandresolution.asciidoc
```diff
@@ -281,27 +279,26 @@ A bean must declare _all_ of the qualifiers that are specified at the injection
...
 
 * there is exactly one bean remaining, the container will select this bean, and the ambiguous dependency is called _resolvable_.
 * all the beans left are alternatives with a priority, then the container will determine the highest priority value, and eliminate all beans, except for producer methods and fields of beans that are alternatives with the highest priority value. If there is exactly one bean remaining, the container will select this bean, and the ambiguous dependency is called _resolvable_.
 
-All unresolvable ambiguous EL names are detected by the container when the application is initialized. Suppose two beans are both available for injection in a certain war, and either:
+All unresolvable ambiguous names are detected by the container when the application is initialized. Suppose two beans are both available for injection in a certain application, and either:
```
jharting: this should say "module", not "application"

spec/src/main/doc/core/implementation.asciidoc
```diff
@@ -730,7 +494,7 @@ A bean constructor may have any number of parameters. All parameters of a bean c
 
 === Injected fields
 
-An _injected field_ is a non-static, non-final field of a bean class or of any Java EE component class supporting injection.
+An _injected field_ is a non-static, non-final field of a bean class.
```
jharting: The expression "Java EE component class supporting injection" is not treated consistently in this PR. Here, the expression is eliminated whereas in `injectionandresolution.asciidoc` for example this expression is replaced by "other classes supporting injection". The expression should be dealt with consistently across the entire document.

spec/src/main/doc/core/injectionandresolution.asciidoc
```diff
@@ -384,16 +375,15 @@ When the container creates a new instance of a managed bean, session bean, or of
...
+* Initializer methods declared by a class X in the type hierarchy of the bean are called after all injected fields declared by X or by superclasses of X have been initialized.
+* Any +@PostConstruct+ callback declared by a class X in the type hierarchy of the bean is called after all initializer methods declared by X or by superclasses of X have been called, after all injected fields declared by X or by superclasses of X have been initialized.
 
 
 [[dependent_objects_destruction]]
 
 ==== Destruction of dependent objects
 
-When the container destroys an instance of a bean or of any Java EE component class supporting injection, the container destroys all dependent objects, as defined in <<dependent_destruction>>, after the +@PreDestroy+ callback completes and after the servlet +destroy()+ method is called.
+When the container destroys an instance of a bean, the container destroys all dependent objects, as defined in <<dependent_destruction>>, after the +@PreDestroy+ callback completes and after the servlet +destroy()+ method is called.
```
jharting: Servlet reference

spec/src/main/doc/core/scopescontexts.asciidoc
```diff
@@ -437,116 +415,27 @@ Portable extensions are encouraged to fire an event with qualifier +@Initialized
 
 ==== Request context lifecycle
 
-The _request context_ is provided by a built-in context object for the built-in scope type +@RequestScoped+. The request scope is active:
-
-* during the +service()+ method of any servlet in the web application, during the +doFilter()+ method of any servlet filter and when the container calls any +ServletRequestListener+ or +AsyncListener+,
-* during any Java EE web service invocation,
-* during any remote method invocation of any EJB, during any asynchronous method invocation of any EJB, during any call to an EJB timeout method and during message delivery to any EJB message-driven bean, and
-* during +@PostConstruct+ callback of any bean.
```
jharting: Is the "request scope available during @PostConstruct" feature intentionally restricted to Java EE only or is this an oversight?

spec/src/main/doc/core/spi.asciidoc
```diff
@@ -42,7 +42,7 @@ public interface Bean<T> extends Contextual<T>, BeanAttributes<T> {
 }
 ----
 
-* +getBeanClass()+ returns the bean class of the managed bean or session bean or of the bean that declares the producer method or field.
+* +getBeanClass()+ returns the bean class of the bean or of the bean that declares the producer method or field.
```
jharting: should be "+getBeanClass()+ returns the bean class of the managed bean"

spec/src/main/doc/core/spi.asciidoc
```diff
@@ -174,9 +174,9 @@ public interface InjectionTarget<T>
 }
 ----
 
-* +inject()+ performs dependency injection upon the given object. The container performs Java EE component environment injection, according to the semantics required by the Java EE platform specification, sets the value of all injected fields, and calls all initializer methods, as defined in <<fields_initializer_methods>>.
-* +postConstruct()+ calls the +@PostConstruct+ callback, if it exists, according to the semantics required by the Java EE platform specification.
-* +preDestroy()+ calls the +@PreDestroy+ callback, if it exists, according to the semantics required by the Java EE platform specification.
+* +inject()+ performs dependency injection upon the given object. The container sets the value of all injected fields, and calls all initializer methods, as defined in <<fields_initializer_methods>>.
+* +postConstruct()+ calls the +@PostConstruct+ callback, if it exists.
+* +preDestroy()+ calls the +@PreDestroy+ callback.
```
jharting: callback, if it exists

spec/src/main/doc/core/spi.asciidoc
```diff
@@ -1034,9 +1019,9 @@ If any +ProcessAnnotatedType+ method is called outside of the observer method in
 
 ==== +ProcessInjectionPoint+ event
 
-The container must fire an event for every injection point of every Java EE component class supporting injection that may be instantiated by the container at runtime, including every managed bean declared using +@ManagedBean+, EJB session or message-driven bean, bean, interceptor or decorator.
+The container must fire an event for every injection point of every component class supporting injection that may be instantiated by the container at runtime, including every managed bean declared using +@ManagedBean+, bean, interceptor or decorator.
```
jharting: `@ManagedBean` reference should be moved to the EE part

spec/src/main/doc/core/spi.asciidoc
jharting: @ManagedBean reference should be moved to the EE part

spec/src/main/doc/core/spi.asciidoc
```diff
@@ -1076,19 +1061,19 @@ public interface ProcessInjectionTarget<X> {
...
 * +getInjectionTarget()+ returns the +InjectionTarget+ object that will be used by the container to perform injection.
 * +setInjectionTarget()+ replaces the +InjectionTarget+.
 * +addDefinitionError()+ registers a definition error with the container, causing the container to abort deployment after bean discovery is complete.
 
 
-Any observer of this event is permitted to wrap and/or replace the +InjectionTarget+. The container must use the final value of this property, after all observers have been called, whenever it performs injection upon the managed bean, session bean or other Java EE component class supporting injection.
+Any observer of this event is permitted to wrap and/or replace the +InjectionTarget+. The container must use the final value of this property, after all observers have been called, whenever it performs injection upon the bean or other component class supporting injection.
 
-For example, this observer decorates the +InjectionTarget+ for all servlets.
+For example, this observer decorates the +InjectionTarget+ for all Maps.
```
jharting: A Map is not a good substitute here. Map is a generic type and raw references to generic types are deprecated in Java since version 5.

spec/src/main/doc/javaee/definition_ee.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+* _stateless objects_, such as servlets and stateless session beans, which do not contain client-visible state, or
+* objects that must be explicitly created and destroyed by their client, such as JavaBeans and stateful session beans, whose state is shared by explicit reference passing between clients.
+
+CDI scopes add to Java EE these missing well-defined lifecycle context as defined in <<scopes>>.
+
+[[builtin_scopes_ee]]
+
+==== Built-in scope types in Java EE
+
+When running in Java EE, the implementation of the @RequestScoped, @ApplicationScoped and @SessionScoped annotations provided by the container, represents the standard scopes defined by the Java Servlets specification.
```
jharting: s/implementation/implementations
s/represents/represent

spec/src/main/doc/javaee/definition_ee.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+
+
+[[default_bean_discovery_ee]]
+=== Default bean discovery mode for Java EE
+
+When running in Java EE, the container must extend the rules defined in <<default_bean_discovery>> with:
+
+If the _bean discovery mode_ is `annotated`, then:
+
+* bean classes of EJB sessions beans, are not discovered, and
```
jharting: bean classes of EJB session beans *should be discovered* when in `annotated` mode

spec/src/main/doc/javaee/definition_ee.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+* producer methods that are not on an EJB session bean are not discovered, and
```
jharting: This is confusing as it kind of implies that in `annotated` archive in Java EE a producer/disposer/observer has to be defined on an EJB session bean

spec/src/main/doc/javaee/implementation_ee.asciidoc
```diff
@@ -0,0 +1,319 @@
...
+* Session beans
+* Resources (Java EE resources, persistence contexts, persistence units, remote EJBs and web services)
+
+Java EE and embeddable EJB containers are required by the Java EE and EJB specifications to support EJB session beans and the Java EE component environment. Other containers are not required to provide support for injection or lifecycle management of session beans or resources.
+
+[[what_classes_are_beans_ee]]
+
+==== Which Java classes are managed beans in Java EE?
+
+When running in Java EE, A top-level Java class is a managed bean if it meets requirement described in <<what_classes_are_beans>> and if it is defined to be a managed bean by any other Java EE specification and if
```
jharting: this should be: "requirements" OR "defined to be a managed bean by any other Java EE specification", not AND

spec/src/main/doc/javaee/implementation_ee.asciidoc
```diff
@@ -0,0 +1,319 @@
...
+
+A disposer method declaration in an EJB session bean follows the rules defined in <<declaring_disposer_method>> with the following addition:
+
+* If a non-static method of an EJB session bean class has a parameter annotated +@Disposes+, and the method is not a business method of the session bean, the container automatically detects the problem and treats it as a definition error.
+
+[[javaee_components]]
+
+=== Java EE components
+
+Most Java EE components support injection and interception, as defined in the Java Platform, Enterprise Edition Specification 7, table EE.5-1, but are not considered beans (as defined by this specification). EJBs, as defined in <<session_beans>> are an exception.
```
jharting: Only *EJB session beans* are the exception

spec/src/main/doc/javaee/injectionandresolution_ee.asciidoc
```diff
@@ -0,0 +1,104 @@
...
+
+When resolving a name in an EL expression, the container considers the bean name and selected alternatives.
+
+[[selection_ee]]
+
+=== Modularity in Java EE
+
+In the Java EE module architecture, any Java EE module or library is a module. The Java EE module is a bean archive if it contains a +beans.xml+ file, as defined in <<bean_archive>>.
+
+When running in Java EE, the container must extend the accessibility rules for beans and alternatives defined in <<selection>> to JSP/JSF pages using EL resolution.
```
jharting: The container should not really extended the accessibility rules. Instead, the container should follow the accessibility rules and make sure that only beans available from injection in the module that defines the JSP/JSF pages are resolved.

spec/src/main/doc/javaee/injectionandresolution_ee.asciidoc
```diff
@@ -0,0 +1,104 @@
...
+
+When running in Java EE, the container must extend the rules defined for managed beans in <<dependent_objects_destruction>> to any other Java EE component class supporting injection.
+
+[[injection_point_ee]]
+
+==== Injection point metadata in Java EE
+
+When running in Java EE, the container must extend the rules defined in <<injection_point>> and is also required to ensure that:
+
+If a Java EE component class supporting injection that is not a bean has an injection point of type +InjectionPoint+ and qualifier +@Default+, the container automatically detects the problem and treats it as a definition error.
```
jharting: This part is implied by <<injection_point>> and redundant

spec/src/main/doc/javaee/packagingdeployment_ee.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+
+== Packaging and deployment in Java EE
+
+
+[[bean_archive_ee]]
+=== Bean archive with EJB Session Beans
+
+When running in Java EE, the container must extend the rules defined in <<bean_archive>> with:
+
+* An implicit bean archive may also contains EJB session beans, and
```
jharting: s/contains/contain

spec/src/main/doc/javaee/packagingdeployment_ee.asciidoc
```diff
@@ -0,0 +1,57 @@
...
+* if an archive meet the requirements to be an _implicit bean archive_ but doesn't contain any bean class with a bean defining annotation and have at least one EJB sessions beans, it's an _implicit bean archive_.
```
jharting: s/meet/meets

jharting: s/have/it has
s/sessions beans/session bean

jharting: "if an archive meet the requirements to be an _implicit bean archive_ but doesn't contain any bean class with a bean defining annotation" is kind of a confusing definition given that "containing classes with bean defining annotation" is itself a half of the original implicit bean archive definition.

spec/src/main/doc/javaee/spi_ee.asciidoc
```diff
@@ -0,0 +1,91 @@
...
+
+== Portable extensions in Java EE
+
+[[interceptor_ee]]
+
+==== The +Interceptor+ interface in Java EE
+
+When Running in Java EE, the container must extend the rules defined in <<interceptor>> and must also ensure that
+ 
+`PRE_PASSIVATE`, `POST_ACTIVATE` and `AROUND_TIMEOUT` InterceptorType value are linked to EJB lifecycle callback or method timeout.
```
jharting: s/value/values

jharting: s/method timeout/timeout method

spec/src/main/doc/preface.asciidoc
```diff
@@ -9,30 +9,26 @@ include::license-{license}.asciidoc[]
...
+A full changelog will be found link:https://issues.jboss.org/secure/ReleaseNote.jspa?version=12323655&styleName=Html&projectId=12311062[here^].
+
+Starting with version 2.0 CDI targets Java SE and Java EE platforms.
+CDI in a Java EE container is mostly a superset of CDI in Java SE, except for features explicitly stating otherwise.
+
+=== Organisation of this document
+
+This document is organized in 4 parts:
+
+* An introduction (this part), whose not part of the specification but introduce CDI concepts and give examples.
```
jharting: "whose not part of" - probably a mistake

spec/src/main/doc/preface.asciidoc
```diff
@@ -9,30 +9,26 @@ include::license-{license}.asciidoc[]
 
 === Foreword
 
-Contexts and Dependency Injection 1.2 is a maintenance release of Contexts and Dependency Injection 1.1 (link:https://jcp.org/en/jsr/detail?id=346[JSR-346^]).
+Contexts and Dependency Injection 2.0 (link:https://jcp.org/en/jsr/detail?id=365[JSR-365^]) is an update to Contexts and Dependency Injection 1.2 (link:https://jcp.org/en/jsr/detail?id=346[JSR-346^]).
+
+A full changelog will be found link:https://issues.jboss.org/secure/ReleaseNote.jspa?version=12323655&styleName=Html&projectId=12311062[here^].
+
+Starting with version 2.0 CDI targets Java SE and Java EE platforms.
+CDI in a Java EE container is mostly a superset of CDI in Java SE, except for features explicitly stating otherwise.
```
jharting: "features explicitly stating otherwise" sounds weird

spec/src/main/doc/core/implementation.asciidoc
antoinesd: You're right, they should be removed. My only concern is for `InjectionTarget`s created at runtime with `BeanManager#CreateInjectionTarget` or indirectly via `Unmanaged`. Some of this rules may apply to the classes used to create these InjectionTarget. wdyt ?

spec/src/main/doc/core/scopescontexts.asciidoc
antoinesd: It is intentional. In core, all normal scope are only said to be required and must be defined depending of the platform. Thus we can define how these scope are implemented in EE and how they be implemented in SE.

spec/src/main/doc/javaee/injectionandresolution_ee.asciidoc
antoinesd: yes in <<injection_point>> we have

>If a class supporting injection that is not a bean has an injection point of type +InjectionPoint+ and qualifier +@Default+, the container automatically detects the problem and treats it as a definition error.

But it should be removed in favor of this mention IMO. A class can support injection without being a bean (`Unmanaged` class or class that could become an `InjectionTarget` at runtime) but eh container cannot detect it automatically. Now what happens if the class I used to create an `InjectionTarget` inject on `InjectionPoint` ?

spec/src/main/doc/core/implementation.asciidoc
antoinesd: So I removed all "* supporting injection" in core except in injectionandresolution.asciidoc (line 347 and 357) and in spi.asciidoc (line 785).
I let them because you could imagine having a non contextual object instantiated by the BM thru `CreateInjectionTarget` or `Unmanaged` and thus supporting injection.
Let me know what you think about this @jharting.

spec/src/main/doc/javaee/injectionandresolution_ee.asciidoc
antoinesd: Remove the mention in core. As I said the container cannot handle automatically non bean classes that contain @Inject. So the mention should be in EE alone.

spec/src/main/doc/core/implementation.asciidoc
jharting: Cannot we use "other classes supporting injection" everywhere where "Java EE component class supporting injection" was used previously?

spec/src/main/doc/core/scopescontexts.asciidoc
jharting: I see

spec/src/main/doc/javaee/injectionandresolution_ee.asciidoc
jharting: If there is a class supporting injection (e.g. `InjectionTarget` or `Unmanaged`) and this class has an `InjectionPoint` injection point then this should be recognized as a problem. Therefore, removing the constraint from core is probably not right.

antoinesd: Yes but it should be an exception at runtime not at boot time. When I read "the container automatically detects the problem" I understand boot time.
So from what I understand the case of having `@Inject InjectionPoint ip;`  in this kind of non contextual object, is not specified...

spec/src/main/doc/core/implementation.asciidoc
antoinesd: Hmm... I'm gonna double check that. I'll put them back in all places where non contextual object instantiated by the container can be used. Like in this specific place ;).

spec/src/main/doc/javaee/injectionandresolution_ee.asciidoc
jharting: All the class definition constraints in the spec are specified this way.

spec/src/main/doc/core/implementation.asciidoc
antoinesd: After check it was the only place. Other deal with automatic injection (when these unmanaged object have manual injection) or automatic check or SPI events firing that are not related to these objects as well.

jharting: I am going to add a comment to places where I think "other classes supporting injection" should be preserved.

spec/src/main/doc/javaee/injectionandresolution_ee.asciidoc
antoinesd: Yes but nearly all these contraints are checked at boot time (automatic). We have specific rules for `Instance#get` since constraint check are done at runtime but nothing . So I think I should open a ticket to specify how contraints should be treated for unmanaged objects. Wdyt?

spec/src/main/doc/core/implementation.asciidoc
```diff
@@ -755,7 +519,7 @@ If an injected field is annotated +@Produces+, the container automatically detec
 
 === Initializer methods
 
-An _initializer method_ is a default-access, public, protected or private, non-abstract, non-static, non-generic method of a bean class or of any Java EE component class supporting injection. If the bean is a session bean, the initializer method is _not_ required to be a business method of the session bean.
+An _initializer method_ is a default-access, public, protected or private, non-abstract, non-static, non-generic method of a bean class.
```
jharting: "other classes supporting injection" should be preserved here

spec/src/main/doc/core/injectionandresolution.asciidoc
```diff
@@ -346,23 +344,17 @@ The behavior of all methods declared by +java.lang.Object+, except for +toString
...
-* contextual instances of session beans, and
 * contextual instances of managed beans.
 
+The container is also required to perform dependency injection whenever it instantiates the following non-contextual objects:
 
-The container is also required to perform dependency injection whenever it instantiates any of the following non-contextual objects:
-
-* non-contextual instances of session beans (for example, session beans obtained by the application from JNDI or injected using +@EJB+),
-* non-contextual instances of managed beans, and
-* instances of any other Java EE component class supporting injection.
```
jharting: "other classes supporting injection" should be preserved here

spec/src/main/doc/core/injectionandresolution.asciidoc
```diff
@@ -370,13 +362,13 @@ The object injected by the container may not be a direct reference to a contextu
...
 ==== Injection using the bean constructor
 
-When the container instantiates a managed bean or session bean with a constructor annotated +@Inject+, the container calls this constructor, passing an injectable reference to each parameter. If there is no constructor annotated +@Inject+, the container calls the constructor with no parameters.
+When the container instantiates a managed bean with a constructor annotated +@Inject+, the container calls this constructor, passing an injectable reference to each parameter. If there is no constructor annotated +@Inject+, the container calls the constructor with no parameters.
 
 [[fields_initializer_methods]]
 
 ==== Injection of fields and initializer methods
 
-When the container creates a new instance of a managed bean, session bean, or of any other Java EE component class supporting injection the container must:
```
jharting: "other classes supporting injection" should be preserved here

spec/src/main/doc/core/injectionandresolution.asciidoc
antoinesd: From what I understand If you request such an instance from the container from an `InjectionTarget` or `Unmanaged` object, dependency injection is not performed by the container at instantiation but when you request it with the `inject()` method. So it shouldn't be there IMO.

spec/src/main/doc/core/implementation.asciidoc
antoinesd: ok. I corrected it there.

spec/src/main/doc/core/injectionandresolution.asciidoc
antoinesd: Same as above. This is not done at instantiation but when `inject()` is called.

spec/src/main/doc/javaee/injectionandresolution_ee.asciidoc
jharting: Let's have a class that is not scanned automatically but instead used to obtain an InjectionTarget implementation from the container. If this class has definition errors then these should be found and reported by the container when `createInjectionTargte()` is called. This is required by the `createInjectionTarget()` method contract. Therefore, the original constraint:
> If a Java EE component class supporting injection that is not a bean has an injection point of type +InjectionPoint+ and qualifier +@Default+, the container automatically detects the problem and treats it as a definition error.

makes sense in core and should stay untouched IMHO.

spec/src/main/doc/core/injectionandresolution.asciidoc
jharting: ok

spec/src/main/doc/javaee/injectionandresolution_ee.asciidoc
antoinesd: you mean

>If a *class* supporting injection that is not a bean has an injection point of type +InjectionPoint+ and qualifier +@Default+, the container automatically detects the problem and treats it as a definition error

and removing the mention for Java EE component class in EE. Ok for me.

jharting: Yes, sorry. Keep the sentence you quoted in core and remove this specialized "Java EE version" from EE part as it is redundant (implied by the core part).

antoinesd:
Close in favor of #251 targetedt to the right branch

PR #242, opened 2015-04-19
CDI-4, CDI-521 Draft for event observer ordering

mp911de:
First draft for event observer ordering without API changes and not affecting extensions

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/Bean.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/Bean.java
@@ -32,7 +32,7 @@ import javax.enterprise.context.spi.CreationalContext;
  * @author David Allen
  * @param <T> the class of the bean instance
  */
-public interface Bean<T> extends Contextual<T>, BeanAttributes<T>, Prioritized {
+public interface Bean<T> extends Contextual<T>, BeanAttributes<T> {
 
     /**
      * The bean {@linkplain Class class} of the managed bean or session bean or of the bean that declares the producer method or
--- a/api/src/main/java/javax/enterprise/inject/spi/Bean.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/Bean.java
@@ -32,7 +32,7 @@ import javax.enterprise.context.spi.CreationalContext;
  * @author David Allen
  * @param <T> the class of the bean instance
  */
-public interface Bean<T> extends Contextual<T>, BeanAttributes<T> {
+public interface Bean<T> extends Contextual<T>, BeanAttributes<T>, Prioritized {
 
     /**
      * The bean {@linkplain Class class} of the managed bean or session bean or of the bean that declares the producer method or
--- a/api/src/main/java/javax/enterprise/inject/spi/Prioritized.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/Prioritized.java
@@ -1,0 +1,21 @@
+package javax.enterprise.inject.spi;
+
+/**
+ * <p>
+ * Provides access to the priority value of prioritized elements.
+ * </p>
+ *
+ * @author Mark Paluch
+ * @since 1.1
+ */
+public interface Prioritized {
+
+    /**
+     * <p>
+     * Returns the priority value of this prioritized element.
+     * </p>
+     *
+     * @return the priority value
+     */
+    public int getPriority();
+}
--- a/api/src/main/java/javax/enterprise/inject/spi/Prioritized.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/Prioritized.java
@@ -6,7 +6,7 @@ package javax.enterprise.inject.spi;
  * </p>
  *
  * @author Mark Paluch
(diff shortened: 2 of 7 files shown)


jharting:
* Observers with no `@Priority` should be treated as having a default priority. That way it is easier to define the relation between an observer with `@Priority` and the one without it.
* We should not add `@Priority(APPLICATION)` to all examples in the spec as it may lead to reader's understanding that the priority is something they need to specify (whereas explicit priority definition should be optional)
* An extension API for setting / altering priority of an observer method will be needed

antoinesd:
@mp911de, thanks for your PR, it's always nice to see new contributor proposing content for the spec.

I agree with points raised by @jharting.

- We shouldn’t put a priority on all example but have a paragraph introduce them.
- In this paragraph we should also state that an observer without a priority has a default priority of APPLICATION level
- We should also state that observers with the same priority are executed in a non predictable order

In a second time we should also think about SPI

- ObserverMethod should have a new getPriority() method
- We should also add something regarding extension and ordering of event in chapter 11. I propose that for the moment we add a mention saying that priority isn’t taken into account for extension observer. We’ll be able to go back tot hat mention later should we decide to have ordering of SPI event.

johnament:
Are you thinking we should add a `get/setPriority` to the [ObserverMethod](http://docs.oracle.com/javaee/6/api/javax/enterprise/inject/spi/ObserverMethod.html) interface?

antoinesd:
@johnament. For the moment `ObserverMethod` is a readonly metadata so a `getPriority()` as I wrote in my comment. Makes more senses.
If we want to add ObserverMethod mutability I think we should add a `setObserverMethod()` to `ProcessObserverMethod` for the sake of consistency.

johnament:
@antoinesd to do something like that, we'd need to have some kind of builder or easy to recreate functionality to override, I'd hate for something like this to be a DeltaSpike only type of solution...

```
ObserverMethod<?> om = ObserverMethodBuilder.from(processObserverMethod.getObserverMethod()).withPriority(N);
```

We'd also have to make sure `getPriority` is a default method, since technically people can implement `ObserverMethod` themselves.

antoinesd:
yes @johnament, but we are on another subject : builder for meta-data like `AnnotatedType`, `BeanAttributes`, `InjectionPoint` and others. I think we have a ticket for these (jira is not responding right now).

jharting:
Also if/once https://issues.jboss.org/browse/CDI-484 gets resolved it would simplify overriding CDI metadata a lot.

mp911de:
Thanks for the feedback. I incorporated parts of your feedback into the PR. Open points are:
* `setPriority` in `ObserverMethod`
* `setObserverMethod` in `ProcessObserverMethod`

mp911de:
Added changes from our last meeting. See https://www.dropbox.com/s/vnkh7agw4pz1waa/cdi-spec.html?dl=0 for the HTML version.

jharting:
We should do something with the [BeanManager.resolveObserverMethods()](http://red.ht/1bxpcqY) method for it to express the order of resolved observers.

spec/src/main/doc/events.asciidoc
```diff
@@ -411,6 +411,7 @@ When an event is fired by the application, the container must:
 
 * determine the observer methods for that event according to the rules of observer resolution defined by <<observer_resolution>>, then,
 * for each observer method, either invoke the observer method immediately, or register the observer method for later invocation during the transaction completion phase, using a JTA +Synchronization+.
+* honor the priority if it is defined using the +@Priority+ annotation.
```
jharting: This should be updated to reflect that each observer now has a priority

mp911de:
I'll take care of it.

api/src/main/java/javax/enterprise/inject/spi/Bean.java
```diff
@@ -32,7 +32,7 @@
  * @author David Allen
  * @param <T> the class of the bean instance
  */
-public interface Bean<T> extends Contextual<T>, BeanAttributes<T> {
+public interface Bean<T> extends Contextual<T>, BeanAttributes<T>, Prioritized {
```
jharting: Currently only a small subset of beans has a priority. Even not all alternatives, interceptors and decorators declared priority (are instead enabled and ordered in beans.xml). Therefore, we should not force this on all Bean implementations but rather make it optional. Similarly, a container should provide a Prioritized Bean impl only if it creates a Bean for a class that has @Priority.

api/src/main/java/javax/enterprise/inject/spi/Prioritized.java
```diff
@@ -0,0 +1,21 @@
+package javax.enterprise.inject.spi;
+
+/**
+ * <p>
+ * Provides access to the priority value of prioritized elements.
```
jharting: I would add some more information e.g. a reference to the priority ranges. Other than that this looks good to me.

antoinesd:
This PR is replaced by #248 which target the right branch

PR #243, opened 2015-04-21
CDI-26 Bootstrap support

johnament:
Squashed commits:
[d005b5a] CDI-26 Cleaned up docs after EG review.
[4a86f00] CDI-26 Simplified try with resources example.
[e248ce4] CDI-26 Added docs for SE.
[a731d53] CDI-26 Added docs for SE.
[8d1c9d4] CDI-26 Close shouldn't have been throwing an exception.
[dd81e7a] CDI-26 Minor javadoc clean up.
[3115343] CDI-509 Updating pom file to use latest felix plugin.
[9db0748] CDI-26 Bootstrap API for CDI.

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/CDIProvider.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/CDIProvider.java
@@ -1,5 +1,8 @@
 package javax.enterprise.inject.spi;
 
+import java.util.Collections;
+import java.util.Map;
+
 /**
  * Interface implemented by a CDI provider to provide access to the current container
  * 
@@ -14,5 +17,32 @@ public interface CDIProvider {
      * @return the CDI instance for the current container
      */
     public CDI<Object> getCDI();
+
+    /**
+     * Determines whether or not this CDIProvider has been initialized or not.
+     * @return true if initialized, false if not.
+     */
+    boolean isInitialized();
+
+    /**
+     * Initializes a CDI Container.  Cannot be called within an application server or if the container is already started.
+     * If called in either of these ways, an {@link java.lang.IllegalStateException} is thrown.
+     *
+     * @return the {@link CDI} instance associated with the container.  This is the same instance returned by using
+     * {@link CDI.current()}
+     */
+    default CDI<Object> initialize() {
+        return initialize(Collections.EMPTY_MAP);
+    }
+
+    /**
+     * Initializes a CDI Container.  Cannot be called within an application server or if the container is already started.
+     * If called in either of these ways, an {@link java.lang.IllegalStateException} is thrown.
+     *
+     * @param params optional parameters, may be null or empty.  May also be immutable.
+     * @return the {@link CDI} instance associated with the container.  This is the same instance returned by using
+     * {@link CDI.current()}
+     */
+    CDI<Object> initialize(Map<String,Object> params);
 
 }
--- a/api/src/main/java/javax/enterprise/inject/spi/CDIProvider.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/CDIProvider.java
@@ -16,7 +16,7 @@ public interface CDIProvider {
      * 
      * @return the CDI instance for the current container
      */
-    public CDI<Object> getCDI();
+    CDI<Object> getCDI();
 
     /**
      * Determines whether or not this CDIProvider has been initialized or not.
@@ -25,8 +25,8 @@ public interface CDIProvider {
     boolean isInitialized();
 
     /**
-     * Initializes a CDI Container.  Cannot be called within an application server or if the container is already started.
-     * If called in either of these ways, an {@link java.lang.IllegalStateException} is thrown.
+     * Initializes a CDI Container.  Cannot be called within an application server.
+     * If called in this way, an {@link java.lang.UnsupportedOperationException} is thrown.
      *
      * @return the {@link CDI} instance associated with the container.  This is the same instance returned by using
      * {@link CDI.current()}
@@ -36,8 +36,8 @@ public interface CDIProvider {
     }
 
     /**
-     * Initializes a CDI Container.  Cannot be called within an application server or if the container is already started.
-     * If called in either of these ways, an {@link java.lang.IllegalStateException} is thrown.
+     * Initializes a CDI Container.  Cannot be called within an application server.
+     * If called in this way, an {@link java.lang.UnsupportedOperationException} is thrown.
      *
      * @param params optional parameters, may be null or empty.  May also be immutable.
      * @return the {@link CDI} instance associated with the container.  This is the same instance returned by using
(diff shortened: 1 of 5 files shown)


spec/src/main/doc/cdi-se.asciidoc
```diff
@@ -0,0 +1,51 @@
...
+    MyBean myBean = cdi.select(MyBean.class).get();
+    myBean.doWork();
+    // when done
+    cdi.shutdown();
+}
+----
+
+The `CDI` utility class provides static lookup for `CDIProvider` instances based on `ServiceLoader` (this doesn't preclude the notion of manually instantiating the provider in a non-portable manner).  By default, the provider will be looked up via a ServiceLoader but can be over-written by calling `setCDIProvider` on the utility.  Once a `CDIProvider` has been located, it should not be instantiated again, instead the same one returned again.
+
+The `CDIProvider` instance returned is capable of starting a `CDI` instance.  By invoking either of the  `initialize` methods it can start up a container to retriever beans, fire events, gain access to the `BeanManager`.  The `initialize` methods on the `CDIProvider` interface must not be invoked when in an EE container.  If they are invoked, an `UnsupportedOperationException` may be thrown.  Likewise, if `shutdown` or `close` were invoked on the `CDI` class while in an EE container, an `UnsupportedOperationException` may also be thrown.
```
tremes: I am not native speaker nor some english language master but I wouldn't use words like "can", "should", "may". So I suggest:
"it will start up a container to retrieve beans...."
"UnsupportedOperationException is thrown." - this is current spec wording.

spec/src/main/doc/cdi-se.asciidoc
```diff
@@ -0,0 +1,51 @@
...
+        MyBean myBean = cdi.select(MyBean.class).get();
+        myBean.doWork();
+    }
+    // shuts down automatically after the try with resources block.
+}
+----
+
+=== Classpath Scanning
+
+All classpath entries in a Java SE application are considered to be at the same level.  All rules that apply to bean resolution based on `beans.xml` declaration continue to apply the same way in Java SE.
```
mkouba: I'm not sure about using class path entries. First this will not cover all use-cases - the class path is not the only place where classes are loaded from (WebStart, Java extensions, ...). And there are other problems. E.g. class path wildcards - how would you process `foo/*` where foo is a directory with several jar files?

johnament: @mkouba actually what happens when you do foo/* is that it's expanded to the full list of entries.  foo/* doesn't become a CP entry.

This section is mostly to represent what we do today in EE, emulated in SE.  Later on after EDR1 I would expect we're going to talk more about bean discovery mode in SE.

mkouba: @johnament I see, so `foo/*` becomes `foo/bar.jar;foo/baz.jar`. Still there are other use-cases (except for WebStart and extensions) where this will probably not work - e.g. weld and owb arquillian embedded containers. My only concern is that if we stick with this solution from the beginning it might be harder to change/replace it later on.

spec/src/main/doc/cdi-se.asciidoc
```diff
@@ -0,0 +1,51 @@
...
+
+=== Bootstrapping a container
+
+In Java EE, CDI comes enabled by default in your application, but if you're working with a Java SE application you can bootstrap a CDI container yourself to enable dependency injection.
+
+[source,java]
+----
+public static void main(String... args) {
+    CDIProvider provider = CDI.getCDIProvider();
+    CDI<Object> cdi = container.initialize();
```
antoinesd: Where does `container` comes from? Shouldn't it be `provider` here?

spec/src/main/doc/cdi-se.asciidoc
antoinesd: I agree with @tremes. Wording feels too different from the rest of the spec.

spec/src/main/doc/cdi-se.asciidoc
```diff
@@ -0,0 +1,51 @@
...
+
+The `CDIProvider` instance returned is capable of starting a `CDI` instance.  By invoking either of the  `initialize` methods it can start up a container to retriever beans, fire events, gain access to the `BeanManager`.  The `initialize` methods on the `CDIProvider` interface must not be invoked when in an EE container.  If they are invoked, an `UnsupportedOperationException` may be thrown.  Likewise, if `shutdown` or `close` were invoked on the `CDI` class while in an EE container, an `UnsupportedOperationException` may also be thrown.
+
+Only a single `CDI` instance may be initialized within an SE application at a time, based on the provider used.  If a container has already been initialized and you call `initialize` again, an `IllegalStateException` is thrown.  If you manually instantiate a `CDIProvider` and initialize it, non-portable behavior may occur.
```
antoinesd: I thought we decided to let the possibility to have multiple container in one application. I agree that it is useless without a new bean discovery mechanism but nothing in your proposed API prevent this possibility. Am I wrong?

spec/src/main/doc/cdi-se.asciidoc
gastaldi: I think we should adopt (or stick) with the RFC 2119 wording: https://www.ietf.org/rfc/rfc2119.txt

jharting:
I dislike the strict wording forbidding multiple parallel containers. Furthermore, I still think `CDIProvider` should stay an SPI used internally by `CDI` for communication with the implementation. We should not expose that class to application code. Ideally the initialize methods exposed to application would be static methods of the `CDI` class (which would use CDIProvider internally). This is temporary anyway since we'll most likely need some kind of builder API for building the application programmatically (as opposed to automated scanning).

Btw it is better not to squash together commits with unrelated changes (e.g. fixing maven-bundle-plugin).

spec/src/main/doc/cdi-se.asciidoc
johnament: What would you guys recommend here.  Is it simple enough to swap "may" for "shall"?

spec/src/main/doc/cdi-se.asciidoc
johnament: I would think if we add webstart support, it would be through some kind of extension to load those classes.  How does weld/owb deal with webstart today?

spec/src/main/doc/cdi-se.asciidoc
johnament: I think we need to work through how to "multi boot" (so to speak) a bit more.  I would prefer to define how it behaves by default (global) and then go back to clean up for sub deployments.  To me, you cannot initialize two of these global CDI instances at once.

api/src/main/java/javax/enterprise/inject/spi/CDIProvider.java
```diff
@@ -15,4 +18,31 @@
...
     public CDI<Object> getCDI();
 
+    /**
+     * Determines whether or not this CDIProvider has been initialized or not.
+     * @return true if initialized, false if not.
+     */
+    boolean isInitialized();
+
+    /**
+     * Initializes a CDI Container.  Cannot be called within an application server or if the container is already started.
```
antoinesd: Update the javadoc to allow multiple call like you allowed it in the spec.

````
Initializes a CDI Container.  If called within an application server, an {@link java.lang.IllegalStateException} is thrown.
`````

File: api/src/main/java/javax/enterprise/inject/spi/CDIProvider.java [hunk 6]
```diff
@@ -15,4 +18,31 @@
...
+     * @return the {@link CDI} instance associated with the container.  This is the same instance returned by using
+     * {@link CDI.current()}
+     */
+    default CDI<Object> initialize() {
+        return initialize(Collections.EMPTY_MAP);
+    }
+
+    /**
+     * Initializes a CDI Container.  Cannot be called within an application server or if the container is already started.
+     * If called in either of these ways, an {@link java.lang.IllegalStateException} is thrown.
```
antoinesd: same as javadoc above

antoinesd:
PR merged in EDR1

antoinesd:
This PR has been retarget to the right branch in PR #249

PR #250, opened 2015-06-02
CDI-499 Async events

antoinesd:
First draft for Async events proposal


jharting:
This proposal lacks certain parts:
* invocation context definition for async delivery
* thread-safety guarantees for events and observers
* does not deal with possibly multiple exceptions being thrown in async event delivery

antoinesd:
Yes @jharting, aspect are missing. That's why it's a draft. Now the true challenge is to bring proposal for these lacking aspects ;).

Regarding Exception, `CompletionStage` provides `handle` and `exceptionally` methods. That's why I wrote:

> Exception can be handled with standard `CompletionStage` methods.

I could have mentioned `handle` and `exceptionally` in my text. But beyond that I don't think useful to reinvent exception handling in async operation unless we don't use JDK 8 API...

jharting:
@antoinesd yes, not doubts about that. My point was that the this proposal, unless I misinterpret it, allows multiple observers resolved for a particular event to be notified of this event in parallel threads. In such case it may happen that these parallel executions yield more than one exception. The proposal however only talks about a single exception as if the observers were invoked sequentially. It IMHO misses the case of more than one exception.

antoinesd:
Ok I get it @jharting. Using handle will only treat the first exception happening in the pipeline. We should have an `handle()` after each `thenApply()`. I see 2 solutions to deal with that:

1. add a signature for fireAsync() allowing to pass a consumer method provided by user. This method will be used internally to `handle()` each observer execution. Something like `thenApply().handle().thenApply()...` (or `exceptionally`) internally

2. provide a hook (via extension?) to event bus allowing addition of these handle methods. Seems a lot more complex and fuzzy than 1...

spec/src/main/doc/events.asciidoc
```diff
@@ -73,6 +73,26 @@ else {
...
+----
+@Inject Event<LoggedInEvent> loggedInEvent;
+
+public void login() {
+    ...
+    loggedInEvent.fireAsync( new LoggedInEvent(user) );
+}
+----
+
+When an event is fired asynchronously, all the observers are called in one or more different thread than the one in which `fireAsync()` was called.
```
mp911de: s/thread/threads

spec/src/main/doc/events.asciidoc
```diff
@@ -292,14 +318,23 @@ The event parameter may be an array type whose component type contains a type va
...
 ----
 public void afterLogin(@Observes LoggedInEvent event) { ... }
+
+public void asyncAfterLogin(@ObservesAsync LoggedInEvent event) { ... }
 ----
 
-If a method has more than one parameter annotated +@Observes+, the container automatically detects the problem and treats it as a definition error.
+If a method has more than one parameter annotated +@Observes+ or `@ObservesAsync`, the container automatically detects the problem and treats it as a definition error.
+
+if a method has one parameter annotated +@Observes+ and one parameter annotated `@ObservesAsync`, the container automatically detects the problem and treats it as a definition error.
```
mp911de: s/if/If

spec/src/main/doc/events.asciidoc
```diff
@@ -425,6 +460,7 @@ Any observer method called before completion of a transaction may call +setRollb
 Observer methods may throw exceptions:
 
 * If the observer method is a transactional observer method, any exception is caught and logged by the container.
+* If the observer method is asynchronous, the exception aborts processing of the observer. Exception can be handled with standard `CompletionStage` methods.
```
mp911de: s/Exception/Exceptions

spec/src/main/doc/events.asciidoc
```diff
@@ -84,6 +104,8 @@ The +Event+ interface provides a method for firing events with a specified combi
 public interface Event<T> {
       
     public void fire(T event);
+    public CompletionStage<void> fireAsync(T event);
```
mp911de: primitive `void` is not allowed here. You need to use `Void` type.

api/src/main/java/javax/enterprise/event/Event.java
```diff
@@ -100,6 +100,30 @@
 
     /**
      * <p>
+     * Fires an event with the specified qualifiers and notifies observers.
```
jharting: This should be expanded.

api/src/main/java/javax/enterprise/event/Event.java
```diff
@@ -100,6 +100,30 @@
...
+     * </p>
+     *
+     * @param event the event object
+     * @throws IllegalArgumentException if the runtime type of the event object contains a type variable
+     * @throws FireAsyncException   if a notified observer throws a checked exception, it will be wrapped and rethrown as an
+     *                                  (unchecked) {@link FireAsyncException}
+     */
+    public void fireAsync(T event);
```
jharting: the method signature does not match the one introduced in the specification document

api/src/main/java/javax/enterprise/event/Event.java
```diff
@@ -100,6 +100,30 @@
 
     /**
      * <p>
+     * Fires an event with the specified qualifiers and notifies observers.
+     * </p>
+     *
+     * @param event the event object
+     * @throws IllegalArgumentException if the runtime type of the event object contains a type variable
+     * @throws FireAsyncException   if a notified observer throws a checked exception, it will be wrapped and rethrown as an
```
jharting: This is wrong. An async observer method may throw an exception after this method has already returned.

api/src/main/java/javax/enterprise/event/Event.java
```diff
@@ -100,6 +100,30 @@
...
+     *                                  (unchecked) {@link FireAsyncException}
+     */
+    public void fireAsync(T event);
+
+    /**
+     * <p>
+     * Fires an event with the specified qualifiers and notifies observers.
+     * </p>
+     *
+     * @param event the event object
```
jharting: executor param missing

api/src/main/java/javax/enterprise/event/Event.java
```diff
@@ -100,6 +100,30 @@
...
+     * @throws IllegalArgumentException if the runtime type of the event object contains a type variable
+     * @throws FireAsyncException   if a notified observer throws a checked exception, it will be wrapped and rethrown as an
+     *                                  (unchecked) {@link FireAsyncException}
+     */
+    public void fireAsync(T event, Executor executor);
```
jharting: same signature mismatch as above

api/src/main/java/javax/enterprise/event/FireAsyncException.java
```diff
@@ -0,0 +1,34 @@
+package javax.enterprise.event;
+
+/**
+ * <p>
+ * This Exception all exceptions that occurred during notification of async observers of an event.
```
jharting: a verb is missing in the first part of the sentence

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
```diff
@@ -229,6 +230,41 @@
...
+     * @throws IllegalArgumentException if two instances of the same qualifier type are given
+     * @throws IllegalArgumentException if an instance of an annotation that is not a qualifier type is given
+     * @throws IllegalArgumentException if the runtime type of the event object is assignable to the type of a container
+     *                                  lifecycle event
+     * @throws FireAsyncException       if a notified observer throws an exception, it will be caught and added
+     *                                  as a suppressed exception in a {@link FireAsyncException}.
+     *                                  This exception will be launch at the end of the notification pipeline
+     * @since 2.0
+     */
+    public void fireAsyncEvent(Object event, Annotation... qualifiers);
```
mp911de: Missing return value `CompletionStage<U>`

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
```diff
@@ -229,6 +230,41 @@
...
+     * @throws IllegalArgumentException if two instances of the same qualifier type are given
+     * @throws IllegalArgumentException if an instance of an annotation that is not a qualifier type is given
+     * @throws IllegalArgumentException if the runtime type of the event object is assignable to the type of a container
+     *                                  lifecycle event
+     * @throws FireAsyncException       if a notified observer throws an exception, it will be caught and added
+     *                                  as a suppressed exception in a {@link FireAsyncException}.
+     *                                  This exception will be launch at the end of the notification pipeline
+     * @since 2.0
+     */
+    public void fireAsyncEvent(Object event, Executor executor, Annotation... qualifiers);
```
mp911de: Missing return value `CompletionStage<Object>`

api/src/main/java/javax/enterprise/event/Event.java
antoinesd: Yes. I obviously lost a commit in my rebases :-( . Looking for it right now

spec/src/main/doc/events.asciidoc
```diff
@@ -122,7 +148,9 @@ If two instances of the same qualifier type are passed to +select()+, an +Illega
 
 If an instance of an annotation that is not a qualifier type is passed to +select()+, an +IllegalArgumentException+ is thrown.
 
-The method +fire()+ fires an event with the specified qualifiers and notifies observers, as defined by <<observer_notification>>. If the container is unable to resolve the parameterized type of the event object, it uses the specified type to infer the parameterized type of the event types.
+The methods +fire()+ and `fireAsync()` fire an event with the specified qualifiers and notifies observers, as defined by <<observer_notification>>. If the container is unable to resolve the parameterized type of the event object, it uses the specified type to infer the parameterized type of the event types.
```
jharting: s/notifies/notify

spec/src/main/doc/events.asciidoc
```diff
@@ -122,7 +148,9 @@ If two instances of the same qualifier type are passed to +select()+, an +Illega
...
+
+The method `fireAsync()` may be called with a specific `Executor` object to be used to invoke observers method. The container should provide a default `Executor` when `fireAsync()` is called without specifying one.
```
jharting: We should specify a case when the given executor is not capable of executing observer notifications (e.g. shut down). Should the fireAsync method throw something right away?

spec/src/main/doc/events.asciidoc
```diff
@@ -288,18 +318,31 @@ The event parameter type may contain a type variable or wildcard.
 
 The event parameter may be an array type whose component type contains a type variable or a wildcard.
 
+Modifications made to the event parameter in an observer method are propagated to following observers and notification point.
```
jharting: what is "notification point"?

spec/src/main/doc/events.asciidoc
```diff
@@ -288,18 +318,31 @@ The event parameter type may contain a type variable or wildcard.
...
+The container is not required to guarantee a consistent state for an event parameter modified by asynchronous observers.
```
jharting: This kind of contradicts the previous statement or am I missing something?

spec/src/main/doc/events.asciidoc
```diff
@@ -288,18 +318,31 @@ The event parameter type may contain a type variable or wildcard.
...
 ----
 public void afterLogin(@Observes LoggedInEvent event) { ... }
+
+public void asyncAfterLogin(@ObservesAsync LoggedInEvent event) { ... }
 ----
 
-If a method has more than one parameter annotated +@Observes+, the container automatically detects the problem and treats it as a definition error.
+If a method has more than one parameter annotated +@Observes+ or `@ObservesAsync`, the container automatically detects the problem and treats it as a definition error.
+
+If a method has one parameter annotated +@Observes+ and one parameter annotated `@ObservesAsync`, the container automatically detects the problem and treats it as a definition error.
```
jharting: This is already included in the preceeding statement, isn't it?

spec/src/main/doc/events.asciidoc
```diff
@@ -425,11 +472,36 @@ Any observer method called before completion of a transaction may call +setRollb
 Observer methods may throw exceptions:
 
 * If the observer method is a transactional observer method, any exception is caught and logged by the container.
+* If the observer method is asynchronous, the exception aborts processing of the observer. Exception management during an asynchronous event is defined in <<async_exception>>.
```
jharting: What about other asynchronous observers in such case? Must, may or mustn't they be aborted also?

spec/src/main/doc/events.asciidoc
```diff
@@ -425,11 +472,36 @@ Any observer method called before completion of a transaction may call +setRollb
...
 
 
 For a custom implementation of the +ObserverMethod+ interface defined in <<observer_method>>, the container must call +getTransactionPhase()+ to determine if the observer method is transactional observer method, and +notify()+ to invoke the method.
 
+[[async_exception]]
+
+==== Handling multiple exceptions thrown during an asynchronous event
+
+If an event is asynchronous, and one or more exception occurred in notified observers, the container must throw a `FireAsyncException`.
+`FireAsyncException` contains all suppressed exception launch by observers. They can be accessed as an array of `throwable` with the `getSuppressed` method.
```
jharting: s/exception/exceptions
s/launch/thrown
s/handle/handled
s/method/methods

spec/src/main/doc/events.asciidoc
```diff
@@ -451,6 +523,10 @@ The order of more than one observer with the same priority is undefined and the
 
 The transaction context, client security context and lifecycle contexts active when an observer method is invoked depend upon what kind of observer method it is.
 
+* If the observer method is asynchronous, it is called:
+ ** in the same client security context as invocation of `Event.fireAsync()` or `BeanManager.fireAsync()`, and
+ ** in an unspecified transaction context, and
+ ** within a new lifecycle for request context, session context and conversation context.
```
jharting: Are these supposed to be some dummy session and coversation contexts serving the single event delivery only? Are these really needed?

jharting:
How can an extension-provided observer method identify itself as sync/async?

spec/src/main/doc/events.asciidoc
antoinesd: Previous line was for synchronous observer, this one is to state that container doesn't provide thread safety for payload modified in async observer.
The idea is to say "if user manage to have a thread safe payload" modification will be taken into account".
Better formulation is of course welcome.

spec/src/main/doc/events.asciidoc
antoinesd: the returns of fireAsync(). I remove the mention.

spec/src/main/doc/events.asciidoc
antoinesd: yes, I remove this last one.

spec/src/main/doc/events.asciidoc
antoinesd: Yes, we can say they are dummy. Do you suggest to forbid (exception thrown) injection of bean in these scopes?

antoinesd:
@jharting, thanks for the precise feedback. I'm adding a `isAsync` method to `ObserverMethod`. I probably have other points to add to spi chapter.

spec/src/main/doc/events.asciidoc
mkouba: I think that session and conversation contexts should be probably inactive during async delivery...

api/src/main/java/javax/enterprise/event/Event.java
```diff
@@ -100,6 +106,36 @@
 
     /**
      * <p>
+     * Fires an event asynchronously with the specified qualifiers and notifies synchronous and asynchronous observers.
+     * </p>
+     *
+     * @param event the event object
+     * @throws IllegalArgumentException if the runtime type of the event object contains a type variable
+     * @throws FireAsyncException   if a notified observer throws an exception, it will be caught and added
```
jharting: Is this exception supposed to be thrown out of the fireAsync method or is the completion stage supposed to fail with it?

api/src/main/java/javax/enterprise/event/FireAsyncException.java
```diff
@@ -0,0 +1,34 @@
+package javax.enterprise.event;
+
+/**
+ * <p>
+ * This Exception list all exceptions that occurred during notification of async observers of an event.
```
jharting: s/Exception/exception
s/list/?

jharting: If this exception is relevant to exceptions thrown out of async observers only, we should define what happens with exceptions raised by sync observers.

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
```diff
@@ -229,6 +231,41 @@
...
+     * @throws IllegalArgumentException if two instances of the same qualifier type are given
+     * @throws IllegalArgumentException if an instance of an annotation that is not a qualifier type is given
+     * @throws IllegalArgumentException if the runtime type of the event object is assignable to the type of a container
+     *                                  lifecycle event
+     * @throws FireAsyncException       if a notified observer throws an exception, it will be caught and added
+     *                                  as a suppressed exception in a {@link FireAsyncException}.
+     *                                  This exception will be thrown at the end of the notification pipeline
+     * @since 2.0
+     */
+    public CompletionStage<Object> fireAsyncEvent(Object event, Annotation... qualifiers);
```
jharting: We can probably be more type-safe by propagating the event type to the return type.

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
```diff
@@ -229,6 +231,41 @@
     public void fireEvent(Object event, Annotation... qualifiers);
 
     /**
+     * Fire an asynchronous event and notify observers.
```
jharting: Do we need these methods on BeanManager as well? I know that fireEvent() is already duplicated since CDI 1.0 though I am wondering if we want to follow this. The BeanManager already has way too many methods. In addition, the CDI 1.0 fireEvent() method makes sense on BeanManager since it allows events to be fired from extensions (where Event built-in bean cannot be injected). With fireAsync() there may not be the need.

api/src/main/java/javax/enterprise/event/Event.java
antoinesd: From what I understand, CompletionStage doesn't fail, it allows to handle exception with handle() or exceptionally(). As describe in spec I think the last step in the async event pipeline building would be to check if we filled suppressedException in a FireAsyncExcetion created for the execution and if it's the case throw the exception. User will be able to handle this exception and dig thru the suppressed exceptions.

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
antoinesd: I think it can be useful for code outside CDI. For instance user can get beanmanager from JNDI and fire an event asynchronously to trigger code in the bean graph without having to add CDI code to the calling module.

spec/src/main/doc/events.asciidoc
antoinesd: and why not request @mkouba?

mkouba: Well, the request context is not HTTP-specific. E.g. it's active during any asynchronous method invocation of any EJB..

api/src/main/java/javax/enterprise/event/Event.java
jharting: Yes, I used a wrong term. I meant "completeExceptionally" instead of fail.

So if my understanding is correct then `fireAsync` never throws an exception (even though Javadoc says so). Instead, if any of the synchronous or asynchronous observers throws an exception then the resulting CompletionStage is completed exceptionally with FireAsyncException that wraps all the exceptions raised by observers. If no exception is thrown by observers then the resulting CompletionStage is completed normally with the event payload.

If this is a correct understanding then it should be specified in the spec and Javadoc more explicitly.

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
jharting: I see. Perhaps having a `BeanManager.event()` or `getEvent()` method that returns an `Event<Object>` object could prevent API swell up?

api/src/main/java/javax/enterprise/event/Event.java
antoinesd: Yes that's exactly the expected behavior. I'll try to rework that to be more precise on that point.

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
antoinesd: Good idea. I remove the fireAsyncEvent and add a the method. It would deprecate beanManager#fireEvent as well.

nykolaslima:
Hello guys,

This feature is going to be released only in CDI 2.0?

antoinesd:
Hi Nykolas,


Yes, for 2 reasons:

- In Java EE new features are never back ported in earlier version
- this feature rely on Java 8 and other CDI version are but on Java 6 and 7.

Why this question ?

Antoine

nykolaslima:
I'm developing a new feature in my system and async events would help a lot. There is a expected release date for it?

PR #251, opened 2015-06-11
CDI 160 Core, SE, EE split for EDR1

antoinesd:
This PR is the continuation of pr #241 but targeted to the right branch. Please refer to discussion on PR #241 for history.


spec/src/main/doc/core/injectionandresolution.asciidoc
```diff
@@ -346,23 +344,17 @@ The behavior of all methods declared by +java.lang.Object+, except for +toString
...
-
-* non-contextual instances of session beans (for example, session beans obtained by the application from JNDI or injected using +@EJB+),
-* non-contextual instances of managed beans, and
-* instances of any other Java EE component class supporting injection.
-
-A Java EE 5 container is not required to support injection for non-contextual objects.
+* non-contextual instances of managed beans.
 
-The container interacts with instances of beans and other Java EE component classes supporting injection by calling methods and getting and setting field values.
+The container interacts with instances of beans or objects supporting injection by calling methods and getting and setting field values.
```
tremes: Not 100% sure here but I would rather use "beans and classes supporting injection".

spec/src/main/doc/javaee/implementation_ee.asciidoc
```diff
@@ -0,0 +1,323 @@
...
+
+=== Disposer methods on EJB session bean
+
+A disposer method defined in an EJB session bean follows the rules defined in <<disposer_method> with the following addition:
+
+* A disposer method defined in an EJB session bean must be either a business method of the EJB or a static method of the bean class.
+
+[[declaring_disposer_method_ee]]
+
+==== Declaring a disposer method
```
tremes: "on EJB session bean" is missing - to distinguish the chapter names.

spec/src/main/doc/core/injectionandresolution.asciidoc
antoinesd: In CDI core these objects were instantiated via `UnManaged` or an `InjectionTarget`, so I think it's ore object than classes.

spec/src/main/doc/javaee/implementation_ee.asciidoc
antoinesd: thx

spec/src/main/doc/javaee/scopescontext_ee.asciidoc
```diff
@@ -0,0 +1,199 @@
...
+When running in Java EE, the container must extend the rules defined in <<passivation_capable_dependency>>, and must also guarantee that:
+
+* all EJB stateless session beans are passivation capable dependencies,
+* all EJB singleton session beans are passivation capable dependencies,
+* all passivation capable EJB stateful session beans are passivation capable dependencies, and
+* all Java EE resources are passivation capable dependencies.
+
+[[passivation_validation_ee]]
+
+==== Validation of passivation capable beans and dependencies
```
tremes: "in Java EE" is missing - to distinguish chapter names.

spec/src/main/doc/javaee/spi_ee.asciidoc
```diff
@@ -0,0 +1,110 @@
...
+
+[[process_injection_point_ee]]
+
+==== `ProcessInjectionPoint` event and EJB
+
+When running in Java EE, the container must also fire an event for every injection point of every Java EE component class supporting injection that may be instantiated by the container at runtime, including every managed bean declared using +@ManagedBean+, EJB session or message-driven bean.
+
+[[process_injection_target_ee]]
+
+==== `ProcessInjectionTarget` event
```
tremes: "and EJB" is missing - to distinguish chapter names.

spec/src/main/doc/javase/scopescontext_se.asciidoc
```diff
@@ -0,0 +1,58 @@
...
+
+[[request_context_se]]
+
+==== Request context lifecycle in Java SE
+
+When running in Java SE, the container must extend the rules defined in <<request_context>> and is also required to ensure the following rules.
+
+The request scope is active:
+
+* during any method invocation.
```
jharting: So request context isn't shared between method invocations? I really cannot tell from the proposal as the wording is very vague. Has any thought been given on how this is supposed to be implemented?

antoinesd: No. As you said in other mail thread this is temporary for EDR1 and will be
changed with CDI-530. I let it intentionally vague to let you choose the
easiest solution in this temp impl. Now if you think it should be more
precise and have suggestion, I'm all ears...
Le mar. 23 juin 2015 à 12:55, Jozef Hartinger <notifications@github.com> a
écrit :

jharting: OK, so my understanding is that you haven't really thought about this. You just put it here vaguely for us to figure out and make work somehow. And then it will be changed anyway. Correct?

In that case my suggestion would be to figure out the semantics (not necessarily the implementation) first and only then to put it into the spec.

antoinesd: Yes, the idea is to make your life easy for EDR1 impl. It's useless to give you hard work for a feature that will be obviously changed...
If you prefer we can decide that request context behaves like application context in SE. My goal is to make reviewer work as simple as possible to test and evaluate EDR1 using existing lib or own code.

jharting: Then I do not understand why we are adding this in the first place. What libraries are not going to work?

antoinesd: DeltaSpike for instance. Jozef, what is your suggestion here?
Le mar. 23 juin 2015 à 14:19, Jozef Hartinger <notifications@github.com> a
écrit :

jharting: Which part of DeltaSpike?

antoinesd: Come on the meeting tonight and ask Mark directly it will be faster than playing riddle on Github.

PR #256, opened 2015-10-05
CDI-458, CDI-522 deactivate an observer and add method setObserver() in ProcessObserverMethod

antoinesd:
CDI-458 Give the possibility to deactivate an observer in ProcessObserverMethod
CDI-522 Create add method setObserver() to ProcessObserverMethod SPI event


tremes:
Proposal looks good to me!

antoinesd:
It also solves CDI-522

PR #261, opened 2015-10-18
CDI-564: Async events simplification

antoinesd:
CDI-564: Async events simplification


spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -396,14 +395,12 @@ public interface EventMetadata {
     public Set<Annotation> getQualifiers();
     public InjectionPoint getInjectionPoint();
     public Type getType();
-    public boolean isAsync();
```
mkouba: Why this modification?

antoinesd: Because you know if you are in a synchronous or asynchronous observer, therefore you know that Event was fired asynchronously or synchronously since there's no more way to trigger async observer with sync event.

mkouba: Yep, you're right.

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -106,8 +106,7 @@ public void login() {
 Event fired with the `fireAsync()` method is fired asynchronously.
 All resolved synchronous observers are called in the same thread in which `fireAsync()` was called, and all the resolved asynchronous observers are called in one or more different threads.
```
tremes: I think this should be changed as well. No SYNC observers are called on "fireAsync()"

tremes:
This seems to solve CDI-543 as well.

antoinesd:
Right @tremes. I link them

PR #263, opened 2015-11-04
CDI-569 Add @ObservesAsync where the spec mention @Observes

antoinesd:
CDI-569 Add @ObservesAsync where the spec mention @Observes


tremes:
Looks good to me.

PR #266, opened 2015-12-01
CDI-563 Event.fireAsync() - clarify the usage of the returned CompletionStage

antoinesd:
CDI-563 Event.fireAsync() - clarify the usage of the returned CompletionStage

### DIFF
--- a/api/src/main/java/javax/enterprise/event/Event.java
+++ b/api/src/main/java/javax/enterprise/event/Event.java
@@ -106,12 +106,13 @@ public interface Event<T> {
 
     /**
      * <p>
-     * Fires an event asynchronously with the specified qualifiers and notifies synchronous and asynchronous observers.
+     * Fires an event asynchronously with the specified qualifiers and notifies asynchronous observers.
      * </p>
      *
      * @param event the event object
      * @return a {@link CompletionStage} allowing further pipeline composition on the asynchronous operation.
-     *         if any of the synchronous or asynchronous observers notified by this event throws an exception
+     *         Default asynchronous execution facility is container specific.
+     *         If any observer notified by this event throws an exception
      *         then the resulting CompletionStage is completed exceptionally with {@link java.util.concurrent.CompletionException}
      *         that wraps all the exceptions raised by observers as suppressed exception.
      *         If no exception is thrown by observers then the resulting CompletionStage is completed normally with the event payload.
@@ -123,14 +124,15 @@ public interface Event<T> {
 
     /**
      * <p>
-     * Fires an event asynchronously with the specified qualifiers and notifies synchronous and asynchronous observers.
+     * Fires an event asynchronously with the specified qualifiers and notifies asynchronous observers.
      * A custom {@link Executor} will be used to make asynchronous calls 
      * </p>
      *
      * @param event the event object
      * @param executor a custom executor to execute asynchronous event
      * @return a {@link CompletionStage} allowing further pipeline composition on the asynchronous operation.
-     *         if any of the synchronous or asynchronous observers notfied by this event throws an exception
+     *         Default asynchronous execution facility is container specific.
+     *         If any observer notified by this event throws an exception
      *         then the resulting CompletionStage is completed exceptionally with {@link java.util.concurrent.CompletionException}
      *         that wraps all the exceptions raised by observers as suppressed exception.
      *         If no exception is thrown by observers then the resulting CompletionStage is completed normally with the event payload.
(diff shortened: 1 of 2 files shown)


api/src/main/java/javax/enterprise/event/Event.java
```diff
@@ -123,14 +126,16 @@
...
      * A custom {@link Executor} will be used to make asynchronous calls 
      * </p>
      *
      * @param event the event object
      * @param executor a custom executor to execute asynchronous event
      * @return a {@link CompletionStage} allowing further pipeline composition on the asynchronous operation.
-     *         if any of the synchronous or asynchronous observers notfied by this event throws an exception
+     *         Default asyncrhonous execution facility of this stage will be based on a container default {@link Executor}
+     *         specific to implementation.
+     *         if any of the synchronous or asynchronous observers notified by this event throws an exception
```
tremes: Typo in "Default asynchronous execution" and mentioning of synchronous observers should be removed.

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -168,6 +168,8 @@ If the container is unable to resolve the parameterized type of the event object
 The method `fireAsync()` may be called with a specific `Executor` object to be used to invoke observers method.
 The container should provide a default `Executor` when `fireAsync()` is called without specifying one.
```
tremes: I would change following sentence 
> The container should provide a default `Executor` ...
to 
> The container must provide a default `Executor` ...
WDYT?

antoinesd: Well, as it is implementation specific I'm not sure...

mkouba: +1 for MUST. I think async events must always work even if an `Executor` is not specified. Although the CDI spec does not use key words from [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) we should probably follow these rules where possible.

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -166,7 +166,13 @@ The methods `fire()` and `fireAsync()` fire an event with the specified qualifie
...
 
 The method `fireAsync()` may be called with a specific `Executor` object to be used to invoke observers method.
-The container should provide a default `Executor` when `fireAsync()` is called without specifying one.
+
+The following elements are container specific:
+
+* the default `Executor` used by the container when `fireAsync()` is called without specifying an `Executor`,
+* the `CompletionStage` returned by `fireAsync` methods,
+* all dependent stages of this initial CompletionStage, and
+* the default asynchronous execution facility used for this stages.
```
tremes: This last point seems to me unnecesarry IMO. It just repeats CompletionStage javadoc.

mkouba: used for **these** stages

api/src/main/java/javax/enterprise/event/Event.java
```diff
@@ -123,14 +124,15 @@
...
+     * Fires an event asynchronously with the specified qualifiers and notifies asynchronous observers.
      * A custom {@link Executor} will be used to make asynchronous calls 
      * </p>
      *
      * @param event the event object
      * @param executor a custom executor to execute asynchronous event
      * @return a {@link CompletionStage} allowing further pipeline composition on the asynchronous operation.
-     *         if any of the synchronous or asynchronous observers notfied by this event throws an exception
+     *         Default asynchronous execution facility is container specific.
+     *         if any observer notified by this event throws an exception
```
tremes: Uppercase "If"?. Same for line 115.

spec/src/main/asciidoc/core/events.asciidoc
antoinesd: @tremes I'm mentioning that we'll have a specific (unspecified) execution facility for the dependent stages, how does it repeat the CS Javadoc?

tremes: It appears to me as it doesn't bring any new information. We are speaking about "default execution facility" which nobody knows what it is (at least it's my feeling) and adittionaly we are saying that it's container specific. 
Anyway it looks like "testable=false" so don't hesitate to ignore me. :)

mkouba: The truth is if we say the `CompletionStage` is container specific then its *default asynchronous execution facility* is also implicitly container specific.

mkouba: First three items are ok.

antoinesd: Ok I remove it.

PR #268, opened 2015-12-16
CDI-566 container events explicitly mentioned as synchronous.

tremes:


manovotn:
Sorry to nag like this, but from grammar point of view, I think it  isn't really common to put an adverb after a verb.
E.g.:
"Portable extensions are encouraged to **fire synchronously** an event with ..." 
Wouldn't it be better to use "**to synchronously fire** ..." or "Portable extensions are encouraged to **fire a synchronous event** with ..." 

The way it is now it just seems a bit weird to me (in all three sentences). But then again, sorry to bother with details :)

tremes:
Actually I was thinking about this. I will update it as you suggested.

antoinesd:
Seems good to me

PR #270, opened 2016-01-05
CDI-558 Standardize the Meta-Data builders API (BeanBuilder, etc...)

antoinesd:
CDI-558 Standardize the Meta-Data builders API (BeanBuilder, etc...)

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
@@ -79,7 +79,7 @@ public interface AfterBeanDiscovery {
      * @throws IllegalStateException if called outside of the observer method invocation
      * @since 2.0
      */
-    public BeanConfigurator<?> addBean();
+    public <T> BeanConfigurator<T> addBean();
 
     /**
      * Fires an event of type {@link javax.enterprise.inject.spi.ProcessObserverMethod} containing the given
@@ -105,7 +105,7 @@ public interface AfterBeanDiscovery {
      * @throws IllegalStateException if called outside of the observer method invocation
      * @since 2.0
      */
-    public ObserverMethodConfigurator<?> addObserverMethod();
+    public <T> ObserverMethodConfigurator<T> addObserverMethod();
 
 
     /**
--- a/api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
@@ -17,6 +17,8 @@
 package javax.enterprise.inject.spi;
 
 import javax.enterprise.context.spi.Context;
+import javax.enterprise.inject.spi.builder.BeanConfigurator;
+import javax.enterprise.inject.spi.builder.ObserverMethodConfigurator;
 
 /**
  * <p>
@@ -41,6 +43,7 @@ import javax.enterprise.context.spi.Context;
  * </p>
  * 
  * @author David Allen
+ * @author Antoine Sabot-Durand
  */
 public interface AfterBeanDiscovery {
     /**
@@ -64,6 +67,21 @@ public interface AfterBeanDiscovery {
     public void addBean(Bean<?> bean);
 
     /**
+     *
+     * Obtains a new {@link BeanConfigurator} to configure  a new {@link Bean}  and add it at the end of the observer invocation.
+     * It will then fire an event of type {@link javax.enterprise.inject.spi.ProcessBean} containing the built
+     * {@link javax.enterprise.inject.spi.Bean} from this configuration and then register it with the
+     * container, thereby making it available for injection into other beans.
+     *
+     * Each call returns a new BeanConfigurator.
+     *
+     * @return a non reusable {@link BeanConfigurator} to configure the bean to add
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public BeanConfigurator<?> addBean();
+
+    /**
      * Fires an event of type {@link javax.enterprise.inject.spi.ProcessObserverMethod} containing the given
      * {@link javax.enterprise.inject.spi.ObserverMethod} and then registers the
      * {@link javax.enterprise.inject.spi.ObserverMethod} with the container, thereby making it available for event
@@ -73,6 +91,22 @@ public interface AfterBeanDiscovery {
      * @throws IllegalStateException if called outside of the observer method invocation
      */
     public void addObserverMethod(ObserverMethod<?> observerMethod);
+
+    /**
+     * obtains a new {@link ObserverMethodConfigurator} to configure a new {@link ObserverMethod} and add it at the end of the observer invocation.
+     * It will then fire an event of type {@link javax.enterprise.inject.spi.ProcessObserverMethod} containing the built
+     * {@link javax.enterprise.inject.spi.ObserverMethod} from this configuration and then registers the
+     * {@link javax.enterprise.inject.spi.ObserverMethod} with the container, thereby making it available for event
+     * notifications.
+     *
+     * Each call returns a new ObserverMethodConfigurator.
+     *
+     * @return a non reusable {@link ObserverMethodConfigurator} instance
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public ObserverMethodConfigurator<?> addObserverMethod();
+
 
     /**
      * Registers a custom {@link javax.enterprise.context.spi.Context} object with the container.
@@ -106,5 +140,4 @@ public interface AfterBeanDiscovery {
      * @since 1.1
      */
     public <T> Iterable<AnnotatedType<T>> getAnnotatedTypes(Class<T> type);
-
 }
--- a/api/src/main/java/javax/enterprise/inject/spi/AfterTypeDiscovery.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/AfterTypeDiscovery.java
@@ -17,6 +17,7 @@
 
 package javax.enterprise.inject.spi;
 
+import javax.enterprise.inject.spi.builder.AnnotatedTypeConfigurator;
 import java.util.List;
 
 /**
@@ -32,6 +33,7 @@ import java.util.List;
  * </p>
  *
  * @author Pete Muir
+ * @author Antoine Sabot-Durand
  * @since 1.1
  */
 public interface AfterTypeDiscovery {
@@ -75,5 +77,33 @@ public interface AfterTypeDiscovery {
      * @throws IllegalStateException if called outside of the observer method invocation
      */
     public void addAnnotatedType(AnnotatedType<?> type, String id);
+
+    /**
+     * <p>
+     * Obtains a new {@link AnnotatedTypeConfigurator} to configure a new {@link javax.enterprise.inject.spi.AnnotatedType} and
+     * add it to the set of types which will be scanned during bean discovery at the end of the observer invocation.
+     * Calling this method multiple times will return a new AnnotatedTypeConfigurator.
+     * </p>
+     *
+     * <p>
+     * This method allows multiple annotated types, based on the same underlying type, to be defined. {@link AnnotatedType}s
+     * discovered by the container use the fully qualified class name of {@link AnnotatedType#getJavaClass()} to identify the
+     * type.
+     * </p>
+     *
+     * <p>
+     * {@link AfterBeanDiscovery#getAnnotatedType(Class, String)} and {@link AfterBeanDiscovery#getAnnotatedTypes(Class)} allows
+     * annotated types to be obtained by identifier.
+     * </p>
+     *
+     * Each call returns a new AnnotatedTypeConfigurator.
+     *
+     *
+     * @param id The id of the annotated type
+     * @return a non reusable {@link AnnotatedTypeConfigurator} to configure the new AnnotatedType
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public AnnotatedTypeConfigurator<?> addAnnotatedType(String id);
 
 }
--- a/api/src/main/java/javax/enterprise/inject/spi/AfterTypeDiscovery.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/AfterTypeDiscovery.java
@@ -104,6 +104,6 @@ public interface AfterTypeDiscovery {
      * @throws IllegalStateException if called outside of the observer method invocation
      * @since 2.0
      */
-    public AnnotatedTypeConfigurator<?> addAnnotatedType(String id);
+    public <T> AnnotatedTypeConfigurator<T> addAnnotatedType(String id);
 
 }
--- a/api/src/main/java/javax/enterprise/inject/spi/BeforeBeanDiscovery.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/BeforeBeanDiscovery.java
@@ -198,7 +198,7 @@ public interface BeforeBeanDiscovery {
      * @throws IllegalStateException if called outside of the observer method invocation
      * @since 2.0
      */
-    public AnnotatedTypeConfigurator<?> addAnnotatedType(String id);
+    public <T> AnnotatedTypeConfigurator<T> addAnnotatedType(String id);
 
 
 }
--- a/api/src/main/java/javax/enterprise/inject/spi/BeforeBeanDiscovery.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/BeforeBeanDiscovery.java
@@ -16,13 +16,13 @@
  */
 package javax.enterprise.inject.spi;
 
-import java.lang.annotation.Annotation;
-
 import javax.enterprise.context.NormalScope;
 import javax.enterprise.inject.Stereotype;
+import javax.enterprise.inject.spi.builder.AnnotatedTypeConfigurator;
 import javax.inject.Qualifier;
 import javax.inject.Scope;
 import javax.interceptor.InterceptorBinding;
+import java.lang.annotation.Annotation;
 
 /**
  * <p>
@@ -32,6 +32,7 @@ import javax.interceptor.InterceptorBinding;
  * 
  * @author Pete Muir
  * @author David Allen
+ * @author Antoine Sabot-Durand
  */
 public interface BeforeBeanDiscovery {
     /**
@@ -171,4 +172,33 @@ public interface BeforeBeanDiscovery {
      * @since 1.1
      */
     public void addAnnotatedType(AnnotatedType<?> type, String id);
+
+    /**
+     * <p>
+     * Obtains a new {@link AnnotatedTypeConfigurator} to configure a new {@link javax.enterprise.inject.spi.AnnotatedType} and
+     * add it to the set of types which will be scanned during bean discovery at the end of the observer invocation
+     * </p>
+     *
+     * <p>
+     * This method allows multiple annotated types, based on the same underlying type, to be defined with a builder.
+     * {@link AnnotatedType}s discovered by the container use the fully qualified class name of
+     * {@link AnnotatedType#getJavaClass()} to identify the type.
+     * </p>
+     *
+     * <p>
+     * {@link AfterBeanDiscovery#getAnnotatedType(Class, String)} and {@link AfterBeanDiscovery#getAnnotatedTypes(Class)} allows
+     * annotated types to be obtained by identifier.
+     * </p>
+     *
+     * Each call returns a new AnnotatedTypeConfigurator
+     *
+     *
+     * @param id The id of the annotated type
+     * @return a non reusable {@link AnnotatedTypeConfigurator} to configure the new AnnotatedType
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public AnnotatedTypeConfigurator<?> addAnnotatedType(String id);
+
+
 }
--- a/api/src/main/java/javax/enterprise/inject/spi/ProcessAnnotatedType.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/ProcessAnnotatedType.java
@@ -16,6 +16,8 @@
  */
 package javax.enterprise.inject.spi;
 
+import javax.enterprise.inject.spi.builder.AnnotatedTypeConfigurator;
+
 /**
  * <p>
  * The container fires an event of this type for each Java class or interface it discovers in a bean archive, before it reads
@@ -42,6 +44,7 @@ package javax.enterprise.inject.spi;
  * </p>
  * 
  * @author David Allen
+ * @author Antoine Sabot-Durand
  * @see AnnotatedType
  * @param <X> The class being annotated
  */
@@ -62,6 +65,18 @@ public interface ProcessAnnotatedType<X> {
      * @throws IllegalStateException if called outside of the observer method invocation
      */
     public void setAnnotatedType(AnnotatedType<X> type);
+
+    /**
+     * Returns an {@link AnnotatedTypeConfigurator} initialized with the {@link AnnotatedType} processed by this event
+     * to configure a new AnnotatedType that will replace the original one at the end of the observer invocation.
+     *
+     * Each call returns the same AnnotatedTypeConfigurator.
+     *
+     * @return a non reusable {@link AnnotatedTypeConfigurator} to configure the replacing AnnotatedType
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public AnnotatedTypeConfigurator<X> configureAnnotatedType();
 
     /**
      * Forces the container to ignore this type.
--- a/api/src/main/java/javax/enterprise/inject/spi/ProcessBeanAttributes.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/ProcessBeanAttributes.java
@@ -18,6 +18,7 @@
 package javax.enterprise.inject.spi;
 
 import javax.enterprise.inject.New;
+import javax.enterprise.inject.spi.builder.BeanAttributesConfigurator;
 
 /**
  * <p>
@@ -33,6 +34,7 @@ import javax.enterprise.inject.New;
  * </p>
  * 
  * @author Pete Muir
+ * @author Antoine Sabot-Durand
  * @param <T> The class of the bean
  * @since 1.1
  */
@@ -58,6 +60,18 @@ public interface ProcessBeanAttributes<T> {
      * @throws IllegalStateException if called outside of the observer method invocation
      */
     public void setBeanAttributes(BeanAttributes<T> beanAttributes);
+
+    /**
+     * returns a {@link BeanAttributesConfigurator} initialized with the {@link BeanAttributes} processed by this event
+     * to configure a new BeanAttributes that will replace the original one at the end of the observer invocation.
+     *
+     * Each call returns the same BeanAttributesConfigurator.
+     *
+     * @return a non reusable {@link BeanAttributesConfigurator} to configure the replacing BeanAttributes
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public BeanAttributesConfigurator<T> configureBeanAttributes();
 
     /**
      * Registers a definition error with the container, causing the container to abort deployment after bean discovery is
--- a/api/src/main/java/javax/enterprise/inject/spi/ProcessInjectionPoint.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/ProcessInjectionPoint.java
@@ -17,6 +17,8 @@
 
 package javax.enterprise.inject.spi;
 
+import javax.enterprise.inject.spi.builder.InjectionPointConfigurator;
+
 /**
  * <p>
  * The container fires an event of this type for every injection point of every Java EE component class supporting injection
@@ -36,6 +38,7 @@ package javax.enterprise.inject.spi;
  * 
  * @see InjectionPoint
  * @author Pete Muir
+ * @author Antoine Sabot-Durand
  * @param <X> the declared type of the injection point.
  * @param <T> the bean class of the bean that declares the injectoion point
  */
@@ -54,6 +57,18 @@ public interface ProcessInjectionPoint<T, X> {
      * @throws IllegalStateException if called outside of the observer method invocation
      */
     public void setInjectionPoint(InjectionPoint injectionPoint);
+
+    /**
+     * Returns an {@link InjectionPointConfigurator} initialized with the {@link InjectionPoint} processed by this event
+     * to configure a new InjectionPoint that will replace the original one at the end of the observer invocation.
+     *
+     * Each call returns the same InjectionPointConfigurator
+     *
+     * @return a non reusable {@link InjectionPointConfigurator} to configure the replacing InjectionPoint
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public InjectionPointConfigurator configureInjectionPoint();
 
     /**
      * Registers a definition error with the container, causing the container to abort deployment after bean discovery is
--- a/api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeBuilder.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeBuilder.java
@@ -28,7 +28,7 @@ import javax.enterprise.inject.spi.AnnotatedType;
  *
  * @param <T> the class represented by the AnnotatedType to build
  * @author Antoine Sabot-Durand
- * @see Builders#annotatedType()
+ * @see Builders#annotatedType(Class)
  * @since 2.0
  */
 public interface AnnotatedTypeBuilder<T> {
--- a/api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeBuilder.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeBuilder.java
@@ -1,0 +1,50 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package javax.enterprise.inject.spi.builder;
+
+import javax.enterprise.inject.spi.AnnotatedType;
+
+/**
+ * This API is an helper to configure and build a new {@link AnnotatedType} instance.
+ * <p/>
+ * The same AnnotatedTypeBuilder can be used multiple times.
+ * <p/>
+ * This builder is not thread safe and shall not be used concurrently.
+ *
+ * @param <T> the class represented by the AnnotatedType to build
+ * @author Antoine Sabot-Durand
+ * @see Builders#annotatedType()
+ * @since 2.0
+ */
+public interface AnnotatedTypeBuilder<T> {
+
+    /**
+     * Give access to the builder configuration
+     *
+     * @return an AnnotatedTypeConfigurator to configure the builder
+     */
+    AnnotatedTypeConfigurator<T> configure();
+
+    /**
+     * Build the AnnotatedType
+     *
+     * @return the built AnnotatedType
+     */
+    AnnotatedType<T> build();
+
+}
--- a/api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
@@ -307,6 +307,16 @@ public interface AnnotatedTypeConfigurator<T> {
     AnnotatedTypeConfigurator<T> removeFromAll(Class<? extends Annotation> annotationType);
 
     /**
+     * Remove the provided annotation from the type, and all of it's members. If the
+     * annotation appears on the type declaration, or any of it's members it will be removed.
+     *
+     * @param annotation the annotation to remove
+     * @throws IllegalArgumentException if the annotation is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromAll(Annotation annotation);
+
+    /**
      * Reads in from an existing AnnotatedType. Any elements not present are
      * added. The javaClass will be read in. If the annotation already exists on
      * that element in the configurator the read annotation will be used.
--- a/api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
@@ -1,0 +1,422 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package javax.enterprise.inject.spi.builder;
+
+import javax.enterprise.inject.spi.AnnotatedConstructor;
+import javax.enterprise.inject.spi.AnnotatedField;
+import javax.enterprise.inject.spi.AnnotatedMethod;
+import javax.enterprise.inject.spi.AnnotatedParameter;
+import javax.enterprise.inject.spi.AnnotatedType;
+import javax.enterprise.inject.spi.ProcessAnnotatedType;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+
+
+/**
+ * This API is an helper to configure a new {@link AnnotatedType} instance.
+ * The CDI container must provides an implementation of this interface.
+ *
+ * AnnotatedTypeConfigurator is not reusable.
+ *
+ * This configurator is not thread safe and shall not be used concurrently.
+ *
+ * @see javax.enterprise.inject.spi.BeforeBeanDiscovery#addAnnotatedType(String)
+ * @see javax.enterprise.inject.spi.AfterTypeDiscovery#addAnnotatedType(String)
+ * @see ProcessAnnotatedType#configureAnnotatedType()
+ * @param <T> the class represented by the configured AnnotatedType
+ * @author Antoine Sabot-Durand
+ * @since 2.0
+ */
+public interface AnnotatedTypeConfigurator<T> {
+
+    /**
+     * Add an annotation to the type declaration.
+     *
+     * @param annotation the annotation instance to add
+     * @throws IllegalArgumentException if the annotation is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> addToType(Annotation annotation);
+
+    /**
+     * Remove an annotation class from the type
+     *
+     * @param annotationType the annotation type to remove
+     * @throws IllegalArgumentException if the annotationType is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromType(Class<? extends Annotation> annotationType);
+
+    /**
+     * Remove an annotation instance from the type
+     *
+     * @param annotation the annotation to remove
+     * @throws IllegalArgumentException if the annotation is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromType(Annotation annotation);
+
+    /**
+     * Add an annotation to the specified field. If the field is not already
+     * present, it will be added.
+     *
+     * @param field      the field to add the annotation to
+     * @param annotation the annotation to add
+     * @throws IllegalArgumentException if the annotation is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> addToField(Field field, Annotation annotation);
+
+    /**
+     * Add an annotation to the specified field. If the field is not already
+     * present, it will be added.
+     *
+     * @param field      the field to add the annotation to
+     * @param annotation the annotation to add
+     * @throws IllegalArgumentException if the annotation is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> addToField(AnnotatedField<? super T> field, Annotation annotation);
+
+    /**
+     * Remove an annotation from the specified field.
+     *
+     * @param field          the field to remove the annotation from
+     * @param annotationType the annotation type to remove
+     * @throws IllegalArgumentException if the annotationType is null or if the
+     *                                  field is not currently declared on the type
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromField(Field field, Class<? extends Annotation> annotationType);
+
+    /**
+     * Remove an annotation from the specified field.
+     *
+     * @param field          the field to remove the annotation from
+     * @param annotationType the annotation type to remove
+     * @throws IllegalArgumentException if the annotationType is null or if the
+     *                                  field is not currently declared on the type
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromField(AnnotatedField<? super T> field,
+                                                 Class<? extends Annotation> annotationType);
+
+    /**
+     * Add an annotation to the specified method. If the method is not already
+     * present, it will be added.
+     *
+     * @param method     the method to add the annotation to
+     * @param annotation the annotation to add
+     * @throws IllegalArgumentException if the annotation is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> addToMethod(Method method, Annotation annotation);
+
+    /**
+     * Add an annotation to the specified method. If the method is not already
+     * present, it will be added.
+     *
+     * @param method     the method to add the annotation to
+     * @param annotation the annotation to add
+     * @throws IllegalArgumentException if the annotation is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> addToMethod(AnnotatedMethod<? super T> method, Annotation annotation);
+
+    /**
+     * Remove an annotation from the specified method.
+     *
+     * @param method         the method to remove the annotation from
+     * @param annotationType the annotation type to remove
+     * @throws IllegalArgumentException if the annotationType is null or if the
+     *                                  method is not currently declared on the type
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromMethod(Method method, Class<? extends Annotation> annotationType);
+
+    /**
+     * Remove an annotation from the specified method.
+     *
+     * @param method         the method to remove the annotation from
+     * @param annotationType the annotation type to remove
+     * @throws IllegalArgumentException if the annotationType is null or if the
+     *                                  method is not currently declared on the type
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromMethod(AnnotatedMethod<? super T> method,
+                                                  Class<? extends Annotation> annotationType);
+
+    /**
+     * Add an annotation to the specified method parameter. If the method is not
+     * already present, it will be added. If the method parameter is not already
+     * present, it will be added.
+     *
+     * @param method     the method to add the annotation to
+     * @param position   the position of the parameter to add
+     * @param annotation the annotation to add
+     * @throws IllegalArgumentException if the annotation is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> addToMethodParameter(Method method, int position, Annotation annotation);
+
+    /**
+     * Remove an annotation from the specified method parameter.
+     *
+     * @param method         the method to remove the annotation from
+     * @param position       the position of the parameter to remove
+     * @param annotationType the annotation type to remove
+     * @throws IllegalArgumentException if the annotationType is null, if the
+     *                                  method is not currently declared on the type or if the
+     *                                  parameter is not declared on the method
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromMethodParameter(Method method,
+                                                           int position, Class<? extends Annotation> annotationType);
+
+    /**
+     * Add an annotation to the specified constructor. If the constructor is not
+     * already present, it will be added.
+     *
+     * @param constructor the constructor to add the annotation to
+     * @param annotation  the annotation to add
+     * @throws IllegalArgumentException if the annotation is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> addToConstructor(Constructor<T> constructor, Annotation annotation);
+
+    /**
+     * Add an annotation to the specified constructor. If the constructor is not
+     * already present, it will be added.
+     *
+     * @param constructor the constructor to add the annotation to
+     * @param annotation  the annotation to add
+     * @throws IllegalArgumentException if the annotation is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> addToConstructor(AnnotatedConstructor<T> constructor, Annotation annotation);
+
+    /**
+     * Remove an annotation from the specified constructor.
+     *
+     * @param constructor    the constructor to add the annotation to
+     * @param annotationType the annotation to add
+     * @throws IllegalArgumentException if the annotationType is null or if the
+     *                                  constructor is not currently declared on the type
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromConstructor(Constructor<T> constructor,
+                                                       Class<? extends Annotation> annotationType);
+
+    /**
+     * Remove an annotation from the specified constructor.
+     *
+     * @param constructor    the constructor to add the annotation to
+     * @param annotationType the annotation to add
+     * @throws IllegalArgumentException if the annotationType is null, if the
+     *                                  annotation does not exist on the type or if the constructor is
+     *                                  not currently declared on the type
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromConstructor(AnnotatedConstructor<T> constructor,
+                                                       Class<? extends Annotation> annotationType);
+
+    /**
+     * Add an annotation to the specified constructor parameter. If the
+     * constructor is not already present, it will be added. If the constructor
+     * parameter is not already present, it will be added.
+     *
+     * @param constructor the constructor to add the annotation to
+     * @param position    the position of the parameter to add
+     * @param annotation  the annotation to add
+     * @throws IllegalArgumentException if the annotation is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> addToConstructorParameter(Constructor<T> constructor,
+                                                           int position,
+                                                           Annotation annotation);
+
+    /**
+     * Remove an annotation from the specified constructor parameter.
+     *
+     * @param constructor    the constructor to remove the annotation from
+     * @param position       the position of the parameter to remove
+     * @param annotationType the annotation type to remove
+     * @throws IllegalArgumentException if the annotationType is null, if the
+     *                                  constructor is not currently declared on the type or if the
+     *                                  parameter is not declared on the constructor
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromConstructorParameter(Constructor<T> constructor,
+                                                                int position,
+                                                                Class<? extends Annotation> annotationType);
+
+    /**
+     * Remove an annotation from the specified parameter.
+     *
+     * @param parameter      the parameter to remove the annotation from
+     * @param annotationType the annotation type to remove
+     * @throws IllegalArgumentException if the annotationType is null, if the
+     *                                  callable which declares the parameter is not currently declared
+     *                                  on the type or if the parameter is not declared on either a
+     *                                  constructor or a method
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromParameter(AnnotatedParameter<? super T> parameter,
+                                                     Class<? extends Annotation> annotationType);
+
+    /**
+     * Add an annotation to the specified parameter. If the callable which
+     * declares the parameter is not already present, it will be added. If the
+     * parameter is not already present on the callable, it will be added.
+     *
+     * @param parameter  the parameter to add the annotation to
+     * @param annotation the annotation to add
+     * @throws IllegalArgumentException if the annotation is null or if the
+     *                                  parameter is not declared on either a constructor or a method
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> addToParameter(AnnotatedParameter<? super T> parameter, Annotation annotation);
+
+    /**
+     * Remove annotations from the type, and all of it's members. If an
+     * annotation of the specified type appears on the type declaration, or any
+     * of it's members it will be removed.
+     *
+     * @param annotationType the type of annotation to remove
+     * @throws IllegalArgumentException if the annotationType is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromAll(Class<? extends Annotation> annotationType);
+
+    /**
+     * Reads in from an existing AnnotatedType. Any elements not present are
+     * added. The javaClass will be read in. If the annotation already exists on
+     * that element in the configurator the read annotation will be used.
+     *
+     * @param type the type to read from
+     * @throws IllegalArgumentException if type is null
+     * @return self
+     */
+    <U extends T> AnnotatedTypeConfigurator<T> read(AnnotatedType<U> type);
+
+    /**
+     * Reads in from an existing AnnotatedType. Any elements not present are
+     * added. The javaClass will be read in if overwrite is true. If the
+     * annotation already exists on that element in the configurator, overwrite
+     * determines whether the original or read annotation will be used.
+     *
+     * @param type      the type to read from
+     * @param overwrite if true, the read annotation will replace any existing
+     *                  annotation
+     * @throws IllegalArgumentException if type is null
+     * @return self
+     */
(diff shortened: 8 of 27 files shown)


antoinesd:
Currently, this PR only contains the SPI/API part. I'd like to gather feedback before adding content to the spec.

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodBuilder.java
```diff
@@ -0,0 +1,231 @@
...
+     *
+     * @param consumer to call for the event notification
+     * @return self
+     */
+    ObserverMethodBuilder<T> notifyWith(Consumer<T> consumer);
+
+    /**
+     *
+     * Make the observer to build asynchronous.
+     * When no specified the observer is asynchronous.
```
tremes: Isn't this wrong?
>When not specified the observer is synchronous

tremes:
It seems to me ok...besides the naming of "Helper" and "HelperProvider".

api/src/main/java/javax/enterprise/inject/spi/builder/Helper.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.ServiceLoader;
+
+/**
+ * Provide access to CDI helpers.
+ *
+ * @author Antoine Sabot-Durand
+ * @since 2.0
+ */
+public class Helper {
```
mp911de: Not sure about the future of the `Helper` class but, for now, it's a bit expressionless. How about `Builders` or `BuilderFactory`?

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointBuilder.java
```diff
@@ -0,0 +1,243 @@
...
+
+    /**
+     * Read the InjectionPoint information from the given {@link Constructor}.
+     * All relevant information is overwritten.
+     *
+     * @param constructor defining the InjectionPoint.
+     * @param param       the parameter's index in the Constructor defining the InjectionPoint
+     * @return self
+     */
+    InjectionPointBuilder read(Constructor<?> constructor, int param);
```
mp911de: It might be worth to have another `read` method accepting the parameter name for the constructor argument

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointBuilder.java
```diff
@@ -0,0 +1,243 @@
...
+
+    /**
+     * Read the InjectionPoint information from the given {@link Method}.
+     * All relevant information is overwritten.
+     *
+     * @param method defining the InjectionPoint.
+     * @param param  the parameter's index in the Method defining the InjectionPoint
+     * @return self
+     */
+    InjectionPointBuilder read(Method method, int param);
```
mp911de: It might be worth to have another `read` method accepting the parameter name for the method argument

api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
```diff
@@ -107,4 +109,24 @@
...
 
+    /**
+     *
+     * Gives access to a {@link BeanBuilder} to help building
+     * a {@link Bean} instance
+     *
+     * @return a {@link BeanBuilder}
+     * @since 2.0
+     */
+    public <T> BeanBuilder<T> beanBuilder();
```
mkouba: I think we should add _"@throws IllegalStateException if called outside of the observer method invocation"_. The same applies to other builder-producing methods, e.g. `ProcessBeanAttributes.beanAttributesBuilder()`.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeBuilder.java
```diff
@@ -0,0 +1,413 @@
...
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+
+
+
+/**
+ * This API is an helper to build a new {@link AnnotatedType} instance.
+ * CDI container must provides you an implementation of this interface accessible by calling {@link Helper#getAnnotatedTypeBuilder()}
```
mkouba: CDI container must provide~~s you~~ an implementation...

api/src/main/java/javax/enterprise/inject/spi/builder/BeanAttributesBuilder.java
```diff
@@ -0,0 +1,241 @@
...
+import javax.enterprise.inject.spi.BeanAttributes;
+import javax.enterprise.inject.spi.ProcessBeanAttributes;
+import javax.enterprise.util.TypeLiteral;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.Set;
+
+/**
+ * This API is an helper to build a new {@link BeanAttributes} instance.
+ * CDI container must provides an implementation of this interface accessible by calling
```
mkouba: I would remove this sentence with `@see ProcessBeanAttributes#beanAttributesBuilder()` javadoc tag.

api/src/main/java/javax/enterprise/inject/spi/builder/BeanAttributesBuilder.java
```diff
@@ -0,0 +1,241 @@
...
+     * Add types to the bean types
+     *
+     * @param types types to add
+     * @return self
+     */
+    BeanAttributesBuilder<T> addTypes(Set<Type> types);
+
+    /**
+     * Adds an unrestricted set of bean types for the given type as if it represented a bean class of a managed bean.
+     * Illegal bean types are omitted. See also
```
mkouba: See also what? ;-)

api/src/main/java/javax/enterprise/inject/spi/builder/BeanBuilder.java
```diff
@@ -0,0 +1,369 @@
...
+import java.lang.reflect.Type;
+import java.util.Set;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * This API is an helper to build a new {@link Bean} instance.
+ * CDI container must provides an implementation of this interface accessible by calling
```
mkouba: Again, I would remove this sentence with `@see AfterBeanDiscovery#beanBuilder()` javadoc tag.

api/src/main/java/javax/enterprise/inject/spi/builder/BeanBuilder.java
```diff
@@ -0,0 +1,369 @@
...
+ * The same BeanBuilder can be used multiple times in one AfterBeanDiscovery observer but it shouldn't be used
+ * after the end of observer invocation.
+ *
+ *
+ * @author Martin Kouba
+ * @author Antoine Sabot-Durand
+ * @param <T> the class of the bean instance
+ * @since 2.0
+ */
+public interface BeanBuilder<T> {
```
mkouba: `BeanBuilder` could possibly extend `BeanAttributesBuilder`?

api/src/main/java/javax/enterprise/inject/spi/builder/Helper.java
mkouba: +1 for `Builders` and making the class final.

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointBuilder.java
mkouba: AFAIK class files do not store parameter names by default - the source needs to be compiled with the `-parameters` option. So such a method might not always work...

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodBuilder.java
```diff
@@ -0,0 +1,231 @@
...
+ public interface ObserverMethodBuilder<T> {
+
+    /**
+     *
+     * Read observer meta data from a existing {@link java.lang.reflect.Method}
+     *
+     * @param method to read meta data from
+     * @return self
+     */
+    ObserverMethodBuilder<T> read (Method method);
```
mkouba: Fix the formatting to `ObserverMethodBuilder<T> read(Method method)`...

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodBuilder.java
mkouba: +1 Maybe I would change the wording to something similar:
_Make the observer method asynchronous. By default, the resulting observer method is synchronous._

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeBuilder.java
```diff
@@ -0,0 +1,413 @@
...
+     *
+     */
+    Class<T> getJavaClass();
+
+    /**
+     * set the class represented by the AnnotatedType to build
+     *
+     * @return self
+     */
+    AnnotatedTypeBuilder<T> setJavaClass(Class<T> javaClass);
```
johnament: setJavaClass is distinct from the others.  Why not just javaClass(javaClass) ?

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointBuilder.java
johnament: Right, most Di providers look at type and qualifiers, not names.

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodBuilder.java
johnament: This is confusing.  I would think async() returns the boolean, not toggles.  Why can't I read the async property? Most other builders provide read.

antoinesd: @johnament How would call reader / toggler methods ?

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeBuilder.java
johnament: And maybe "The CDI container must.." ?

api/src/main/java/javax/enterprise/inject/spi/builder/Helper.java
antoinesd: +1 for Builders. I didn't make the class final to mimic CDI. I was happy that this one wasn't final when I had to create a test with it: https://github.com/cdi-spec/cdi/blob/2.0-EDR2/api/src/test/java/org/jboss/cdi/api/test/CDITest.java
@mkouba what would be the benefit to have it final?

api/src/main/java/javax/enterprise/inject/spi/builder/BeanBuilder.java
tremes: It seems so. What about to reflect this entire hierarchy and introduce InterceptorBuilder and DecoratorBuilder as well? Does it make sense?

api/src/main/java/javax/enterprise/inject/spi/builder/Helper.java
mkouba: Well, I think "library" classes should be always final and no one should ever want to customize the functionality (except for implementing a service provider). `CDI` is a bit different because it can be used as a regular object - e.g. `CDI.current().getBeanManager()` and `Instance` methods (and honestly for me `CDI`  is not a very intelligible code ;-). For tests we should probably use package-private members.

api/src/main/java/javax/enterprise/inject/spi/builder/BeanBuilder.java
mkouba: I think it makes sense for interceptors... we already have a Weld issue for this:
https://issues.jboss.org/browse/WELD-2008

antoinesd: In my 1st version BeanBuilder inherited BeanAttributesBuilder, but since we need the method to return the current builder, I had to override all the methods. So it wasn't helping more than having an @Override on common method. WDYT ?

mkouba: Maybe we could use generics and have something similar to TCK's `ArchiveBuilder`:
https://github.com/cdi-spec/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/shrinkwrap/ArchiveBuilder.java#L165
https://github.com/cdi-spec/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/shrinkwrap/WebArchiveBuilder.java
Although it might look like over-engineering. I'm not sure whether there is some better solution...

api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
antoinesd: ok

api/src/main/java/javax/enterprise/inject/spi/builder/BeanAttributesBuilder.java
antoinesd: Why ?

api/src/main/java/javax/enterprise/inject/spi/builder/BeanBuilder.java
antoinesd: I'm not sure to get it. Could you detail please?

api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
mkouba: I can see the `IllegalStateException` mentioned in javadoc for all methods of `AfterBeanDiscovery`.

api/src/main/java/javax/enterprise/inject/spi/builder/BeanAttributesBuilder.java
mkouba: `@see` is usually well arranged in the rendered output. At first sight one can see what this class relates to. If not removing the sentence then at least add the tag.

api/src/main/java/javax/enterprise/inject/spi/builder/BeanBuilder.java
mkouba: Something like this should work:
```
interface BeanAttributesBuilder<T, B extends BeanAttributesBuilder<T, B>> {
  B read(BeanAttributes<?> beanAttributes);
  // This method would be necessary for BeanAttributesBuilder impl to return the correct subclass
  B self();
}

interface BeanBuilder<T> extends BeanAttributesBuilder<T, BeanBuilder<T>> {
  <U extends T> BeanBuilder<U> producing(U instance);
}
```
But again it might be overkill...

api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
antoinesd: ok

api/src/main/java/javax/enterprise/inject/spi/builder/BeanAttributesBuilder.java
antoinesd: Sorry I understood you were suggesting to **remove** the @link when you were suggesting to **replace** it by a @see

mkouba: Actually, I did not mean `@link` at all. I was talking about the sentence starting with "CDI container must.."  :wink:

struberg:
I think we should add to all the builders that they are _not_ supposed to be thread safe. 
The getter methods for the builders should also return a NEW builder instance for every invocation.

mkouba:
+1 for non-thread-safe builders
+1 for explicit note about methods returning new instances

I think the builders should be reusable though.

We could also add convenient methods to simplify the usage - see also https://github.com/weld/api/blob/master/weld/src/main/java/org/jboss/weld/experimental/ExperimentalAfterBeanDiscovery.java#L34

api/src/main/java/javax/enterprise/inject/spi/builder/BeanBuilder.java
antoinesd: Ok, got it except the exact use of self().

mkouba: This method should be implemented by `BeanBuilderImpl` and used in `BeanAttributesBuilderImpl` to return a correct builder type, i.e. `BeanBuilder`...

antoinesd:
@mkouba added mention of non thread safe and new instances

I'm not convinced by the addBean() method returning a BeanBuilder and building / registering it automatically: it breaks the builder pattern IMO.

mkouba:
> it breaks the builder pattern IMO.

No problem. On the other hand, the `build()` contract is clear now and so I don't see how this could break the pattern. The point was users will mostly add the result of the builder.

```
event.addBean().addType(Configuration.class).producing(new Configuration(1));
// VS
event.addBean(event.beanBuilder().addType(Configuration.class).producing(new Configuration(1)).build());
```

> added mention of non thread safe and new instances

Mark's comment was about new instances of builders (e.g. `event.beanBuilder()`), not about the results of `build()` methods.

antoinesd:
>```
>event.addBean().addType(Configuration.class).producing(new Configuration(1));
/>/ VS
>event.addBean(event.beanBuilder().addType(Configuration.class).producing(new >Configuration(1)).build());
>```

I like the DSL style but what happens if user write?

```
event.addBean().addType(Configuration.class).producing(new Configuration(1)).build();
```

mkouba:
> I like the DSL style but what happens if user write?

I'm not sure what the Weld prototype does. But we could disallow it and possibly throw some exception.

antoinesd:
Ok, I'll add it.

api/src/main/java/javax/enterprise/inject/spi/ProcessAnnotatedType.java
```diff
@@ -42,10 +45,11 @@
  * </p>
  * 
  * @author David Allen
+ * @author Antoine Sabot-Durand
  * @see AnnotatedType
  * @param <X> The class being annotated
  */
-public interface ProcessAnnotatedType<X> {
+public interface ProcessAnnotatedType<X> extends AnnotatedTypeBuilderProvider {
```
tremes: What will happen in the following case?:
` void observe(@Observes AnnotatedTypeBuilderProvider event)`
ISE or is it ok?

api/src/main/java/javax/enterprise/inject/spi/builder/Builders.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+public final class Builders {
+
+    protected static volatile BuildersProvider buildersProvider = null;
+
+    public static <T> AnnotatedTypeBuilder<T> annotatedType(Class<T> type) {
+        return provider().getAnnotatedTypeBuilder(type);
+    }
+
+    public static AnnotatedTypeBuilder<?> annotatedType() {
```
tremes: I think the name of this method could be confusing. It's true that `annotatedTypeBuilder` is longer but perhaps little bit more clear?

tremes:
I think the name of the method `javax.enterprise.inject.spi.builder.Builders#annotatedType()` could be confusing. It's true that annotatedTypeBuilder is longer but perhaps little bit more clear?

antoinesd:
@tremes yes I hesitated on the name. I chose this one for the following reasons

- Keep the spirit introduced in `ProcessAnnotatedType#setAnnotatedType()` (vs `ProcessAnnotatedType#setAnnotatedTypeWithConfigurator()`
- Shorter
- As it will be used with `Builders.annotatedType()` I found a bit heavy to have `Builders.annotatedTypeBuilders()` and confusion limiting to have the `Builders` mention at the begining of the call

Now we can decide to be more explicit. It's a matter of taste ;)

antoinesd:
I have still questions regarding this new APis. you'll have probably more ;):

## Configurators contract

What happen if I call `AfterBeanDiscovery#setBean()` and do nothing with the returned `BeanConfigurator`. I think the container should raised an exception.

But the more general question is "what would be the minimum to do to configure a AnnotatedType, Bean, BeanAttribute, InjectionPoint and ObserverMethod configurator".
Is this contract should be documented on configurators' side or on the method returning them in lifecycle events?

Same question for the builders at the time `build()` method is called.

## Using Builders outside container lifecycle events

While it could be interesting to build a `Bean` from standard java code or unrelated CDI code (e.g. build the bean from an AnnotatedType in ProcessAnnotatedType to see what will be created from this type), I don't measure the complexity or possible confusion with such a feature.

My feeling here is we should allow the same feature level someone could achieve today with its own implementation of `Bean`, but I may have overlooked some difficulty regarding the generic custom bean the implementations will have to provide (BTW should it be mentioned somewhere or is it implied when I write that container must provide implementation for Configurators / builder) 

## Interest of `Builder` interface

Builder interface transform configurators in true builders, but is it interesting to have this interface alone vs adding the `build()` method to every builder.
Is the allowed polymorphic approach or defining it as functional interface could be interesting? In both case I'll have to make it publicly visible (right now it is restricted to the package visibility)

mkouba:
@antoinesd Although I quite like the idea of configurators there's one major issue with usability - once you invoke a configurator method on a builder instance you will not be able to call `build()` method anymore because you will get a configurator type. You would have to cast the result:
````
((AnnotatedTypeBuilder)Builders.annotatedType().removeFromType(Foo.class)).build()
````
Or am I missing something? 

Also I'm not so sure we will be always able to initialize all builders outside of container lifecycle event notifications. Let me check again...

**Comment by antoinesd** [comment]
> @antoinesd Although I quite like the idea of configurators there's one major issue with usability - once you invoke a configurator method on a builder instance you will not be able to call `build()` method anymore because you will get a configurator type. You would have to cast the result:
> ````
> ((AnnotatedTypeBuilder)Builders.annotatedType().removeFromType(Foo.class)).build()
> ````
> Or am I missing something?

@mkouba not you're not missing something. It's me. I'll probably have to do the same trick that I did for  `BeanAttributesConfigurator` and `BeanConfigurator`. Or if you have a better idea...

> Also I'm not so sure we will be always able to initialize all builders outside of container lifecycle event notifications. Let me check again...

Yes that's one of my open questions.

**Comment by antoinesd** [comment]
Push a new version to solve problems of type raised by @mkouba and raw type issues.

File: api/src/main/java/javax/enterprise/inject/spi/builder/BeanAttributesConfigurator.java [hunk 15]
```diff
@@ -0,0 +1,238 @@
...
+import javax.enterprise.util.TypeLiteral;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.Set;
+
+/**
+ * This API is an helper to configure a new {@link BeanAttributes} instance.
+ * CDI container must provides an implementation of this interface.
+ *
+ * BeanConfigurator is not reusable.
```
mp911de: s/BeanConfigurator/BeanAttributesConfigurator

api/src/main/java/javax/enterprise/inject/spi/builder/BeanConfigurator.java
```diff
@@ -0,0 +1,371 @@
...
+     */
+    BeanConfigurator<T> alternative(boolean value);
+
+    /**
+     *
+     * Allow implementation to return its actual type
+     *
+     * @return the configurator implementation object
+     */
+    BeanConfigurator<T> self();
```
mkouba: This method is useless if you don't extend `BeanAttributesConfigurator`.

api/src/main/java/javax/enterprise/inject/spi/AfterTypeDiscovery.java
```diff
@@ -76,4 +78,29 @@
...
     public void addAnnotatedType(AnnotatedType<?> type, String id);
 
+    /**
+     * <p>
+     * Obtains a new {@link AnnotatedTypeConfigurator} to configure a new {@link javax.enterprise.inject.spi.AnnotatedType} and
+     * add it to the set of types which will be scanned during bean discovery at the end of the observer invocation
+     * </p>
+     *
+     * <p>
+     * This method allows multiple annotated types, based on the same underlying type, to be defined with a builder.
```
mkouba: `to be defined with a builder`? This is probably a leftover of the previous version.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+
+package javax.enterprise.inject.spi.builder;
+
+import javax.enterprise.inject.spi.AnnotatedType;
+
+/**
+ * This API is an helper to configure and build a new {@link AnnotatedType} instance.
+ * The CDI container must provides an implementation of this interface.
```
mkouba: `s/must provides/must provide` and I think the sentence is useless.

api/src/main/java/javax/enterprise/inject/spi/builder/Builders.java
```diff
@@ -0,0 +1,149 @@
...
+    public static <T> BeanBuilder<T> bean(Class<T> type) {
+        return provider().getBeanBuilder(type);
+    }
+
+    /**
+     * Obtains a generic {@link BeanBuilder}
+     *
+     * @return a reusable BeanBuilder
+     */
+    public static BeanBuilder bean() {
```
tremes: I don't understand to purpose of this non-generic method (and other similar here). I guess that user need to specify the type sooner or later. Or is it available due to reusability of the builders? Anyway it still seems to me bit useless or I just simple need some good example.

api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
```diff
@@ -64,6 +67,19 @@
     public void addBean(Bean<?> bean);
 
     /**
+     *
+     * Obtains a new {@link BeanConfigurator} to configure  a new {@link Bean}  and add at the end of the observer invocation.
```
tremes: > ...and add it at the of the observer invocation....

api/src/main/java/javax/enterprise/inject/spi/BeforeBeanDiscovery.java
```diff
@@ -171,4 +172,31 @@
...
+     *
+     * <p>
+     * This method allows multiple annotated types, based on the same underlying type, to be defined with a builder.
+     * {@link AnnotatedType}s discovered by the container use the fully qualified class name of
+     * {@link AnnotatedType#getJavaClass()} to identify the type.
+     * </p>
+     *
+     * <p>
+     * {@link AfterBeanDiscovery#getAnnotatedType(Class, String)} and {@link AfterBeanDiscovery#getAnnotatedTypes(Class)} allows
+     * annotated types to be obtained by identifier.
```
tremes: > by the identifier

manovotn:
Looking at the PR, there is one thing I wonder about. Given the following sample:
```java
public <T> void addContext(@Observes AfterBeanDiscovery event) {
        event.addObserverMethod().async(false).priority(1000).addQualifier(...). // etc. set tup the method
        event.addObserverMethod().async(true).priority(1).addQualifier(...). // call again with different values
    }
```
My questions are:
* You cannot call ```build()``` here and from the PR I understood it should be built once the observer method invocation is over. Is that correct? E.g. you needn't invoke it yourself.
* If so, what happens if you call the Configurator (in this code sample observer method conf.) twice within the method? Does it apply the latter one? Both?

api/src/main/java/javax/enterprise/inject/spi/ProcessBeanAttributes.java
```diff
@@ -60,6 +62,16 @@
     public void setBeanAttributes(BeanAttributes<T> beanAttributes);
 
     /**
+     * Obtains a new {@link BeanAttributesConfigurator} to configure
+     * a new {@link BeanAttributes} and replace the original on at the end of the observer invocation.
```
tremes: > .. original one ..

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -871,6 +872,7 @@ public interface BeforeBeanDiscovery {
 * `addInterceptorBinding()` declares an annotation type as an interceptor binding type, and specifies its meta-annotations.
 * `addAnnotatedType()` adds a given `AnnotatedType` to the set of types which will be scanned during bean discovery, with an optional identifier.
 The first version of the method is deprecated since version 1.1 of Contexts and Dependency Injection.
+The third version of the method, returns an `AnnotatedTypeConfigurator` as defined in <<annotated_type_configurator>> to easily configure the AnnotatedType which will be added at the end of observer invocation.
```
tremes: > .. which will bea dded at the end of the observer invocation.

api/src/main/java/javax/enterprise/inject/spi/builder/Builders.java
mkouba: I don't like it at all.

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -882,6 +884,29 @@ If any observer method of the `BeforeBeanDiscovery` event throws an exception, t
 
 If any `BeforeBeanDiscovery` method is called outside of the observer method invocation, an `IllegalStateException` is thrown.
 
+[[annotated_type_configurator]]
+
+===== `AnnotatedTypeConfigurator` interface
+
+CDI 2.0 introduced the `javax.enterprise.inject.spi.builder.AnnotatedTypeConfigurator` interfaces to help configuring a new `AnnotatedType` instance.
```
tremes: > CDI 2.0 introduced the `javax.enterprise.inject.spi.builder.AnnotatedTypeConfigurator<T>` interface to help

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -965,6 +996,50 @@ If any observer method of the `AfterBeanDiscovery` event throws an exception, th
...
 If any `AfterBeanDiscovery` method is called outside of the observer method invocation, an `IllegalStateException` is thrown.
 
+
+[[bean_configurator]]
+
+===== `BeanConfigurator` interface
+
+CDI 2.0 introduced the `javax.enterprise.inject.spi.builder.BeanConfigurator` interface to help configuring a new `Bean` instance.
+
+`BeanConfigurator` and `BeanAttributesConfigurator` (defined in <<bean_attributes_configurator>>) extends `GenericBeanAttributesConfigurator` interface to share the same methods related to bean's attributes.
```
tremes: `GenericBeanAttributesConfigurator` is likely leftover from previous proposals.

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1172,6 +1275,28 @@ If any observer method of a `ProcessBeanAttributes` event throws an exception, t
...
+
+* Add type with `addType` or `addTransitiveTypeClosure` methods.
+* Set multiple types with `types` methods.
+* Set scope with `scope` method.
+* Add a qualifier with `addQualifier` method.
+* Set or add multiple qualifiers with `addQualifiers` and `qualifiers` methods.
+* Add a stereotype with `addStereotype` method.
+* Set or add multiple stereotypes with `addStereotypes` and `stereotypes` methods.
+* Set the bean name `name` method.
+* Make the bean an alternative with `alternatives` methods.
```
manovotn: `alternatives` methods -> `alternative` methods

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1098,6 +1178,26 @@ If any observer method of a `ProcessInjectionPoint` event throws an exception, t
...
+With `InjectionPointConfigurator` you can perform the following operation:
+
+* Set the type of `InjectionPoint` with `type` method.
+* Add a qualifier with `addQualifier` method.
+* Set or add multiple qualifiers with `addQualifiers` and `qualifiers` methods.
+* Add a stereotype with `addStereotype` method.
+* Set or add multiple stereotypes with `addStereotypes` and `stereotypes` methods.
+* Set the bean that defines this injection point with `bean` method.
+* Make the injection point delegate with `delegate` methods.
+* Make the injection point a transient field with transientField.
```
manovotn: field with transientField. -> field with `transientField` methods.

mkouba:
@manovotn In your `@Observes AfterBeanDiscovery event` example, I believe each `event.addObserverMethod()` invocation returns a new instance of configurator and both observer methods should be added. `ProcessAnnotatedType.setAnnotatedType()` and similar methods are more problematic. I would say the last configurator will be used. Others will be ignored.

manovotn:
@mkouba good point. I just pasted and edited the first observer I came across and didn't realize that.
I updated my previous comment to work with `@Observes ProcessAnnotatedType event` instead.

api/src/main/java/javax/enterprise/inject/spi/builder/Builders.java
```diff
@@ -0,0 +1,149 @@
...
+        return provider().getBeanBuilder();
+    }
+
+    /**
+     * Obtains a {@link ObserverMethodBuilder} for the given event type
+     *
+     * @param type the type of the event for the built ObserverMethod
+     * @param <T> generic for type
+     * @return a reusable BeanBuilder
```
manovotn: reusable BeanBuilder -> reusable ObserverMethodBuilder

api/src/main/java/javax/enterprise/inject/spi/builder/Builders.java
```diff
@@ -0,0 +1,149 @@
...
+    public static  BeanAttributesBuilder beanAttributes() {
+        return provider().getBeanAttributesBuilder();
+    }
+
+    /**
+     * Obtains a {@link BeanBuilder} for the given type
+     *
+     * @param type the type of the instance of the built Bean
+     * @param <T> generic for type
+     * @return a reusable BeanAttributesBuilder
```
manovotn: reusable BeanAttributesBuilder -> reusable BeanBuilder

api/src/main/java/javax/enterprise/inject/spi/builder/Builders.java
antoinesd: You'll need these raw methods when working with unknown type.

How can know the parameter type for the AnnotatedTypeBuilder if you are in:

`public void example(@Observes ProcessAnnotatedType<?> pat)`

method?

The need is the same for `BeanAttributesBuilder` and `ObserverMethodBuilder` since they can be used in observers with wildcard in their parameter type and a `set` method

It is perhaps useless with `BeanBuilder` since it will be used only in AfterBeanDiscovery and no type parameter enforcement.

api/src/main/java/javax/enterprise/inject/spi/BeforeBeanDiscovery.java
antoinesd: this javadoc was pasted from old `addAnnotatedType(AnnotatedType<?> type, String id)` method

antoinesd:
@manovotn, @mkouba : my idea is when you are working on a `set*` method (like `setAnnotatedType` in `ProcessAnnotatedType`) you always get the same configurator since only one is meaningful. When you are working on `add*` method you get a new one each time.
Configurator are not reusable since the build process is not included while builder are reusable.
I agree that should be put in clearer way.

mkouba:
This pull request gets better and better. However, we went through the API this morning and found some inconsistencies and places which deserve attention ;-).

## Lifecycle events
I think that `ProcessAnnotatedType.setAnnotatedType()`, `ProcessBeanAttributes.setBeanAttributes()`, `ProcessInjectionPoint.setInjectionPoint()` and `ProcessObserverMethod.setObserverMethod()` without params should be renamed to `configureX()` and the returned configurators should be initialized with the event payload, as the usual use case is to modify/wrap the payload.

## Reusability
I think we should remove all the `Builders` methods which return raw types. It does not make sense to make builders completely reusable. Maybe we should only allow the `read` methods to accept subtypes.

## AnnotatedTypeConfigurator
* `read` methods should be probably parameterized with `<U extends T>`, similarly to `BeanConfigurator.read(AnnotatedType<U>)`
* `removeFromType()` variant which accepts `Annotation` is missing - currently it's only possible to remove all the annotations of the given type, e.g. it's not possible to only remove `@Transactional(TxType.REQUIRED)`

## InjectionPointConfigurator
* `delegate()` and `transientField()` seem redundant to me

## BeanConfigurator
* all "create callback" methods must return `BeanConfigurator<U>` (not `<T>`)

## ObserverMethodConfigurator
* `addStereotype()` methods should be removed - stereotypes are only intended for beans
* `read(ObserverMethod<?> method)` should be `read(ObserverMethod<T> method)`
* `async()` method seems redundant to me
* we should add `notify()` variant that accepts `EventMetadata` and `Instance<Object>`

antoinesd:
@mkouba I agree with all these points. Doing the modifications.

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodConfigurator.java
```diff
@@ -0,0 +1,197 @@
...
+
+   /**
+    *
+    * Define notification directly in the observer with event metadata and an Instance for beans
+    *
+    * @param metadata metadata of the event
+    * @param instance instance to retrieve all beans
+    * @return self
+     */
+   ObserverMethodConfigurator<T> notifyWith(EventMetadata metadata, Instance<Object> instance);
```
mkouba: This should be something like `notifyWith(BiFunction<EventMetadata, Instance<Object>, T> callback)`

api/src/main/java/javax/enterprise/inject/spi/builder/BeanAttributesConfigurator.java
```diff
@@ -0,0 +1,222 @@
...
+ * @since 2.0
+ */
+public interface BeanAttributesConfigurator<T> {
+    /**
+     * Read the information from the given annotated type. All relevant information is overwritten.
+     *
+     * @param type class to read information from
+     * @return self
+     */
+    <U extends T> BeanAttributesConfigurator<T> read(AnnotatedType<U> type);
```
mkouba: All these "read" methods should probably return `BeanAttributesConfigurator<U>`...

api/src/main/java/javax/enterprise/inject/spi/builder/BeanConfigurator.java
```diff
@@ -0,0 +1,355 @@
...
+     */
+    BeanConfigurator<T> disposeWith(Consumer<T> callback);
+
+    /**
+     * Read the information from the given annotated type. All relevant information is overwritten.
+     *
+     * @param type class to read information from
+     * @return self
+     */
+    <U extends T> BeanConfigurator<T> read(AnnotatedType<U> type);
```
mkouba: `BeanConfigurator<U>`

api/src/main/java/javax/enterprise/inject/spi/builder/BeanConfigurator.java
```diff
@@ -0,0 +1,355 @@
...
+
+    /**
+     *
+     * Set a {@link Consumer} to destroy a bean instance.
+     * If no dispose callback is specified, a NOOP dispose callback is automatically set.
+     *
+     * @param callback the Consumer to dispose the instance
+     * @return self
+     */
+    BeanConfigurator<T> disposeWith(Consumer<T> callback);
```
manovotn: Shouldn't ```disposeWith``` and ```destroyWith``` also return ```<U>``` rather than ```<T>```?
E.g. ```<U extends T> BeanConfigurator<U> disposeWith(Consumer<U> callback);```
I am rather asking than saying this is wrong. It just caught my attention :)

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodConfigurator.java
mkouba: Eh, not `BiFunction` but a consumer with three arguments. But it doesn't exist in JDK. Well, theoretically we could spit this in two methods `notifyWith(BiConsumer<EventMetadata, T> consumer)` and `notifyWith(BiConsumer<Instance<Object>, T> consumer)` (here the Instance could be used to obtain `EventMetadata`, although I'm not sure it would work).

api/src/main/java/javax/enterprise/inject/spi/builder/BeanConfigurator.java
mkouba: Probably not, because we don't want to allow to `disposeWith(Consumer<Bar> callback)` if `producing(Foo instance)` was defined previously (suppose that `Bar extends Foo`).

manovotn:
One more thing regarding reusability.
You removed methods which return raw type from ```Builders.java```.
But the very same issue remains in ```BuildersProvider.java```. Shouldn't we remove those as well?

api/src/main/java/javax/enterprise/inject/spi/builder/BeanAttributesBuilder.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * This API is an helper to configure and build a new {@link BeanAttributes} instance.
+ * CDI container must provides an implementation of this interface.
+ *
+ * The same BeanBuilder can be used multiple times.
+ *
+ * This builder is not thread safe and shall not be used concurrently.
+ *
+ * @see Builders#beanAttributes()
```
tremes: Additionaly this link to the method doesn't work anymore and need an update.:-) Same for all *Builder interfaces.

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodConfigurator.java
antoinesd: We could add our own `TriConsumer` interface (like https://github.com/ddickstein/Java-Library/blob/master/java8/function/TriConsumer.java for instance).

Regarding resolution of EventMetadata from `Instance` I'd be very surprise if it worked ;).

antoinesd:
Yes @manovotn, you're right thx.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
```diff
@@ -0,0 +1,422 @@
...
+    /**
+     * Remove annotations from the type, and all of it's members. If an
+     * annotation of the specified type appears on the type declaration, or any
+     * of it's members it will be removed.
+     *
+     * @param annotationType the type of annotation to remove
+     * @throws IllegalArgumentException if the annotationType is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromAll(Class<? extends Annotation> annotationType);
```
tremes: Variant of this method which accepts `Annotation` parameter could be helpful here as well. It's questionable in case of other remove* methods - it's probably not needed....

antoinesd: Added

tremes:
While writing new assertions in TCK I realized that `InjectionPointConfigurator` provides methods for handling stereotypes. It doesn't make sense.

antoinesd:
Cut and paste error. I'll remove them.
Le mar. 22 mars 2016 à 13:28, Tomas Remes <notifications@github.com> a
écrit :

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1172,6 +1275,28 @@ If any observer method of a `ProcessBeanAttributes` event throws an exception, t
...
+
+[[bean_attributes_configurator]]
+
+===== `BeanAttributesConfigurator` interface
+
+CDI 2.0 introduced the `javax.enterprise.inject.spi.builder.BeanAttributesConfigurator` interface to help configuring a new `BeanAttributes` instance.
+
+`BeanAttributesConfigurator` must be initialized from an `AnnotatedType`, or an existing `BeanAttributes` with one of its `read()` methods.
+
+With `BeanAttributesConfigurator` you can perform the following operation (defined in `GenericBeanAttributesConfigurator` interface that it extends):
```
tremes: @antoinesd :Remove this `GenericBeanAttributesConfigurator` as well please.

tremes: Maybe we should rather write:
> you can perform following operations

WDYT?

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1295,6 +1421,8 @@ public interface ProcessObserverMethod<T, X> {
 * `getObserverMethod()` returns the `ObserverMethod` object that will be used by the container to call the observer method.
 * `addDefinitionError()` registers a definition error with the container, causing the container to abort deployment after bean discovery is complete.
 * `setObserverMethod()` replaces the `ObserverMethod`.
+`configureObserverMethod()` returns an `ObserverMethodConfigurator` (as defined in <<observer_method_configurator>>) initialized with the `ObserverMethod` processed by the event to easily configure the  `ObserverMethod` which will be used to replace the original one at the end of observer invocation.
+The method always return the same `InjectionPointConfigurator`.
```
tremes: Another copy paste error. Method returns the same `ObserverMethodConfigurator`.

tremes: And fix formatting please. I cannot see list bullet for `configureObserverMethod()` in my html spec.

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1302,3 +1430,53 @@ public interface ProcessObserverMethod<T, X> {
...
+
+The container must provide implementation for all these builders and make them available through `ServiceLoader` for `javax.enterprise.inject.spi.builder.BuildersProvider` interface.
+
+The class `javax.enterprise.inject.spi.builder.Builders` provides static method to obtain each builder:
+
+[source, java]
+----
+public final class Builders {
+    public static <T> AnnotatedTypeBuilder<T> annotatedType(Class<T> type) {}
+    public static <T> AnnotatedTypeBuilder<T> annotatedType(AnnotatedType<T> type) {}
```
manovotn: This method is no longer a part the Builders class and hence this line should be removed.

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1302,3 +1430,53 @@ public interface ProcessObserverMethod<T, X> {
...
+
+The container must provide implementation for all these configurators and make them available in matching container lifecycle events as defined in <<init_events>>.
+
+[[builders]]
+
+=== Builders interfaces
+
+Builders interfaces are helpers to create instances for the following SPI elements:
+
+* `AnnotatedType` built with `javax.enterprise.inject.spi.builder.BeanAttributesBuilder`,
```
manovotn: `AnnotatedType` built with `javax.enterprise.inject.spi.builderAnnotatedTypeBuilder`

spec/src/main/asciidoc/core/spi.asciidoc
antoinesd: +1

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
```diff
@@ -0,0 +1,432 @@
...
+     *
+     * @param parameter      the parameter to remove the annotation from
+     * @param annotationType the annotation type to remove
+     * @throws IllegalArgumentException if the annotationType is null, if the
+     *                                  callable which declares the parameter is not currently declared
+     *                                  on the type or if the parameter is not declared on either a
+     *                                  constructor or a method
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromParameter(AnnotatedParameter<? super T> parameter,
```
tremes: Maybe I am today already out of order but what's this method good for? Which parameter? Same for below method `addToParameter`.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
```diff
@@ -0,0 +1,432 @@
...
+     * constructor is not already present, it will be added. If the constructor
+     * parameter is not already present, it will be added.
+     *
+     * @param constructor the constructor to add the annotation to
+     * @param position    the position of the parameter to add
+     * @param annotation  the annotation to add
+     * @throws IllegalArgumentException if the annotation is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> addToConstructorParameter(Constructor<T> constructor,
```
tremes: What about to add variant of this method with AnnotatedConstructor param? Imagine following use case. I already added some annotation to my **AnnotatedConstructor** (which I stored in some variable). Now I wanted to add some annotation to this constructor parameter and I cannot. I will have to get the same constructor via reflexion.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
tremes: @antoinesd Next idea wrt this method. What about annotations at parameters? I think it would make sense to include it to javadoc as well (similarly as https://github.com/cdi-spec/cdi/blob/2.0-EDR2/api/src/main/java/javax/enterprise/inject/spi/WithAnnotations.java#L36). E.g imagine use case when I wanted to remove all observers.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
```diff
@@ -0,0 +1,432 @@
...
+     * Remove an annotation from the specified constructor.
+     *
+     * @param constructor    the constructor to add the annotation to
+     * @param annotationType the annotation to add
+     * @throws IllegalArgumentException if the annotationType is null, if the
+     *                                  annotation does not exist on the type or if the constructor is
+     *                                  not currently declared on the type
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromConstructor(AnnotatedConstructor<T> constructor,
```
tremes: This is IMHO wrong. It should be `AnnotatedConstructor<? super T> constructor` as well.

api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
```diff
@@ -64,6 +67,21 @@
...
+     * {@link javax.enterprise.inject.spi.Bean} from this configuration and then register it with the
+     * container, thereby making it available for injection into other beans.
+     *
+     * Each call returns a new BeanConfigurator.
+     *
+     * @return a non reusable {@link BeanConfigurator} to configure the bean to add
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public BeanConfigurator<?> addBean();
```
mkouba: This method must be parameterized `<T> BeanConfigurator<T> addBean()`. Otherwise we wouldn't be able to pass lamdas to create callbacks. The same applies to `addObserverMethod()` and other `add()` methods.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
```diff
@@ -0,0 +1,432 @@
...
+
+    /**
+     * Override the declared type of a field
+     *
+     * @param field the field to override the type on
+     * @param type  the new type of the field
+     * @throws IllegalArgumentException if field or type is null
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> overrideFieldType(AnnotatedField<? super T> field, Type type);
```
tremes: I am not sure about usefulness of this method but if we have this then it will make sense to include something as `overrideMethodReturnType` or not? I know it's bit corner case but e.g I have `Foo` field and I overrided it to `Bar` and what about my getter method?

mkouba:
It's not clear whether the result of `BeanBuilder.build()/BeanConfigurator.configure()` is passivation capable or not. In Weld the resulting bean always implements `PassivationCapable` and it's possible to override the default id - see also https://github.com/weld/api/blob/master/weld/src/main/java/org/jboss/weld/experimental/BeanBuilder.java#L255.

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointConfigurator.java
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     *
+     * Change the transient status of the built InjectionPoint.
+     * By default the InjectionPoint is not transient.
+     *
+     * @param trans boolean to define or undefine the transient nature of the configured InjectionPoint
+     * @return self
+     */
+    InjectionPointConfigurator transientField(boolean trans);
```
mkouba: There's no need to abbreviate the name of the param IMHO.

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointConfigurator.java
```diff
@@ -0,0 +1,215 @@
...
+     * Read the InjectionPoint information from the given {@link InjectionPoint}.
+     * All relevant information is overwritten.
+     *
+     * @param delegate the InjectionPoint to get information from
+     * @return self
+     */
+    InjectionPointConfigurator read(InjectionPoint delegate);
+
+    /**
+     * Set the {@link Type} of the InjectionPoint to build.
```
mkouba: We should be more clear that this is **the required type** of injection point. I would also rename the param.

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointConfigurator.java
```diff
@@ -0,0 +1,215 @@
...
+    InjectionPointConfigurator read(AnnotatedParameter<?> param);
+
+    /**
+     * Read the InjectionPoint information from the given {@link InjectionPoint}.
+     * All relevant information is overwritten.
+     *
+     * @param delegate the InjectionPoint to get information from
+     * @return self
+     */
+    InjectionPointConfigurator read(InjectionPoint delegate);
```
mkouba: I would rename the `delegate` param. It's no delegate after all.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
```diff
@@ -0,0 +1,432 @@
...
+     */
+    Class<T> getJavaClass();
+
+    /**
+     * set the class represented by the AnnotatedType to configure
+     *
+     * @param type the Class represented by the AnnotatedType
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> setType(Class<T> type);
```
tremes: I guess I want to **change** the class so more likely `setType(Class<?> type)`?

mkouba: I think this method should be removed. Anyway, it's useless in its current form (you can only pass Class of the same type). Maybe we could make the method parameterized and allow to pass a subtype, i.e. `<U extends T>  AnnotatedTypeConfigurator<U> setType(Class<U> type)`. But this would for example break the contract of PAT imho.

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointConfigurator.java
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * InjectionPointConfigurator is not reusable.
+ *
+ * This builder is not thread safe and shall not be used concurrently.
+ *
+ * @see ProcessInjectionPoint#configureInjectionPoint()
+ * @author Antoine Sabot-Durand
+ * @since 2.0
+ */
+public interface InjectionPointConfigurator {
```
mkouba: I'm missing `member()` and `annotated()` methods. Is it intentional?

api/src/main/java/javax/enterprise/inject/spi/ProcessInjectionPoint.java
```diff
@@ -56,6 +59,18 @@
...
+     * Returns an {@link InjectionPointConfigurator} initialized with the {@link InjectionPoint} processed by this event
+     * to configure a new InjectionPoint that will replace the original one at the end of the observer invocation.
+     *
+     * Each call returns the same InjectionPointConfigurator
+     *
+     * @return a non reusable {@link InjectionPointConfigurator} to configure the replacing InjectionPoint
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public InjectionPointConfigurator configureInjectionPoint();
```
mkouba: All similar methods should define what happens if a configurator is in invalid state, e.g. the user breaks something. But this should be probably defined for `setInjectionPoint()` and friends as well.

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointConfigurator.java
```diff
@@ -0,0 +1,215 @@
...
+
+
+    /**
+     * Read the InjectionPoint information from the given {@link Field}.
+     * All relevant information is overwritten.
+     *
+     * @param field defining the InjectionPoint.
+     * @return self
+     */
+    InjectionPointConfigurator read(Field field);
```
mkouba: For all the read methods we should be probably more concrete what information is read. In this case, `member`, `qualifiers`, `type`, `isTransient` and `isDelegate`. We can't read the `Bean<?>` and `Annotated`.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
antoinesd: removing a annotation from a method parameter can be used to remove an `@Observes` (and deactivate an observer) or a qualifier on an initializer method parameter (or an observer).

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
antoinesd: I'm not saying it's wrong (my generic sience is limited ;) ). But most of these methods come from DeltaSpike (https://github.com/apache/deltaspike/blob/master/deltaspike/core/api/src/main/java/org/apache/deltaspike/core/util/metadata/builder/AnnotatedTypeBuilder.java#L341)

So can you double check, because that would mean DS as a bug too.

api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
antoinesd: ok, got it. I remove all wildcards.

antoinesd: Do you mean existing methods in 1.2 like `public void addBean(Bean<?> bean);`  should also be modified?

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodConfigurator.java
```diff
@@ -0,0 +1,183 @@
...
+     */
+    ObserverMethodConfigurator<T> notifyWith(Consumer<T> callback);
+
+    /**
+     * Defines a BiConsumer to call when event is notified
+     *
+     * @param callback a BiConsumer taking EventMetadata and event type as type parameters
+     * @return self
+     */
+    ObserverMethodConfigurator<T> notifyWith(BiConsumer<EventMetadata, T> callback);
```
mkouba: I think `BiConsumer<T,EventMetadata>` would be a bit nicer ;-)

api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
mkouba: No! You don't need type param there.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
```diff
@@ -0,0 +1,432 @@
...
+     *
+     * @param method         the method to remove the annotation from
+     * @param position       the position of the parameter to remove
+     * @param annotationType the annotation type to remove
+     * @throws IllegalArgumentException if the annotationType is null, if the
+     *                                  method is not currently declared on the type or if the
+     *                                  parameter is not declared on the method
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeFromMethodParameter(Method method,
```
mkouba: We're in Java 8. I think we should use `java.lang.reflect.Parameter` here.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
mkouba: Tomas is right that it's not clear what params should this method process. I guess it's intended for both cases - `AnnotatedConstructor` and `AnnotatedMethod`.

api/src/main/java/javax/enterprise/inject/spi/builder/BeanAttributesConfigurator.java
```diff
@@ -0,0 +1,222 @@
...
+ * @since 2.0
+ */
+public interface BeanAttributesConfigurator<T> {
+    /**
+     * Read the information from the given annotated type. All relevant information is overwritten.
+     *
+     * @param type class to read information from
+     * @return self
+     */
+    <U extends T> BeanAttributesConfigurator<U> read(AnnotatedType<U> type);
```
mkouba: We should probably also add `AnnotatedMember<U> member` variant for producer methods and fields?

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
tremes: +1 for remove.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
```diff
@@ -0,0 +1,432 @@
...
+    /**
+     * Reads in from an existing AnnotatedType. Any elements not present are
+     * added. The javaClass will be read in. If the annotation already exists on
+     * that element in the configurator the read annotation will be used.
+     *
+     * @param type the type to read from
+     * @throws IllegalArgumentException if type is null
+     * @return self
+     */
+    <U extends T> AnnotatedTypeConfigurator<T> read(AnnotatedType<U> type);
```
mkouba: I don't think these read methods make sense. The configurator should be always initialized with some AT (i.e. the class structure must be set) and then users should only be allowed to modify annotations. Maybe I'm wrong but I don't understand the purpose of these methods. Also `BeforeBeanDiscovery.addAnnotatedType(String)` should have a param (Class/AT) the configurator could be initialized from. Otherwise users could try to use `.addAnnotatedType(String).removeFromAll()` -> BANG!

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
mkouba: It's probably ok, because constructors are not members and are not inherited by subclasses. Also _"getConstructors() returns all default-access, public, protected or private constructors **declared for the type**"_ (11.4. Alternative metadata sources).

manovotn: +1 for @mkouba 
On top of what Martin said, it is just weird that you are currently customizing AnnotatedType <T> and while you are at it, you change the constructor of its subtype <U>. That just doesn't make much sense.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
tremes: I am not so sure here. How and with what do you initialize the configurator when you have e.g. `ProcessAnnotatedType<? extends Foo>` ?

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
antoinesd: @manovotn, I'm not changing the constructor by itself but changing its annotations. Removing `@Inject` from a constructor is the use case here.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
antoinesd: It's for both. I add the precision in the javadoc.

api/src/main/java/javax/enterprise/inject/spi/builder/BeanConfigurator.java
manovotn: One more thing regarding this function.
Normally, spec allows you do define Disposer methods which can have other parameters (those being injection points).
While I can see it is not (easily) possible to allow that, we might at least add something like 
```BeanConfigurator<T> disposeWith(BiConsumer<T, Instance<Object>> callback);```
@mkouba WDYT?

mkouba:
@antoinesd @tremes @manovotn I've tried to design a new API for `AnnotatedTypeConfigurator`:
https://github.com/mkouba/cdi/tree/next-atc-api

Few examples can be found here:
https://github.com/mkouba/cdi/blob/next-atc-api/api/src/main/java/javax/enterprise/inject/spi/builder/AnnnotatedConfiguratorsTest.java

Unlike the original version this API  allows the user to iterate/filter the existing methods/fields/constructor/params. The resulting code might sometimes look verbose but I find the API more clear and powerful.

api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
antoinesd: Done in `BeforeBeanDiscovery`, `AfterTypeDiscovery` and `AfterBeandiscovery`.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
antoinesd: Removing

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointConfigurator.java
antoinesd: Well, the InjectionPoint should be initialized by one of the the read method (an other IP or a field or  a parameter). `member()` and `annotated()` will be copied from there. Now, do you see some use case where it would be useful to override them ?

api/src/main/java/javax/enterprise/inject/spi/ProcessInjectionPoint.java
antoinesd: Or thrown during configuration. What happen if I set a `type` in the ip, not compatible with the field or parameter that has to be affected by the container?

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodConfigurator.java
antoinesd: +1

api/src/main/java/javax/enterprise/inject/spi/builder/BeanAttributesConfigurator.java
antoinesd: Adding it

antoinesd:
@mkouba your proposal is very interesting. ATC is not usable in its current state unles if we make it a raw type :-).

mkouba:
@antoinesd The new ATC proposal is ready. I've added some javadoc, removed "findFirst" methods, added corresponding `getAnnotated()` to each AT configurator and updated the examples (AnnnotatedConfiguratorsTest).

antoinesd:
This PR became too complicated (in its history). Code will be proposed again in a new PR soon

PR #271, opened 2016-01-16
CDI-527 add allowProxying rule

struberg:


spec/src/main/asciidoc/core/implementation.asciidoc
```diff
@@ -757,3 +757,24 @@ A bean type must be proxyable if an injection point resolves to a bean:
...
+
+The container has multiple ways to define a list of classes which should be considered proxyable despite of having final methods:
+
+* a comma (`','`) separated list of class names set via the java system property `javax.enterprise.inject.allowProxying.classes`, plus
+* a comma (`','`) separated list of class names set via the environment `JAVAX_ENTERPRISE_INJECT_ALLOWPROXYING_CLASSES`, plus
+* a `allowProxying` entry in `beans.xml` as shown in the following example
+
+[source, xml]
+----
+<beans xmlns="http://xmlns.jcp.org/xml/ns/javaee"
```
mp911de: What about the `beans_2_0.xsd`?

struberg: txs, will add the changes as well.
Will take till tomorrow afternoon as I’m hanging on some customer stuff right now :/

LieGrue,
strub

struberg: Which reminds me that I saw references to beans_1.0 and beans_1.1.xsd in the spec still…

LieGrue,
strub

tremes:
So this now contradicts "3.15. Unproxyable bean types". And what about case of Producers? I guess it doesn't solve anything in the case of Producers or does it? Sorry but I can see rather cons than pros.

struberg:
Did you even bother to build and read the changes? 
a.) I amended the section "Unproxyable bean types". Thus where does it contradict? I need facts, not bold statements...
b.) there is no section 3.15. For the rest: see a.) 
c.) bean types means ALL beans, not only Managed Beans, but also Producer Fields, Producer Methods, etc

tremes:
Ah sorry overlooked your amendment in the chapter.

struberg:
Np and thanks for the review!

spec/src/main/asciidoc/core/implementation.asciidoc
```diff
@@ -743,7 +743,7 @@ Certain legal bean types cannot be proxied by the container:
 
 * classes which don't have a non-private constructor with no parameters,
 * classes which are declared final,
-* classes which have non-static, final methods with public, protected or default visibility,
+* classes which have non-static, final methods with public, protected or default visibility except the classes are explicitly enlisted via <<allowProxying>>,
```
antoinesd: Could we imagine extends this Proxy Allowance to other cases listed here?

mkouba:
Correct me if I'm wrong, but the EG did not agree on the solution in CDI-527 discussion yet. I'm quite surprised how much time we spend with this issue. @struberg Does OWB provide a non-portable config to allow this right now? I went through `OpenWebBeansConfiguration` and did not find anything. I think the best solution would be to provide non-portable config and not to "pollute" the spec.

Other notes:
* I think a regex might be more powerful and less verbose than comma-separated list.
* System properties are ok for Java SE but environment variables? This is nothing system-specific IMHO.
* Again, I don't like adding `allowProxying` entry in `beans.xml` as beans.xml is not a global config file.

spec/src/main/asciidoc/core/implementation.asciidoc
mkouba: Definitely not for final classes. For non-private constructors it's possible using non-portable JVM APIs.

struberg:
Hi Martin!

answers inline

> Am 20.01.2016 um 09:53 schrieb Martin Kouba <notifications@github.com>:
> 
> Correct me if I'm wrong, but the EG did not agree on the solution in CDI-527 discussion yet. I’m quite surprised how much time we spend with this issue. 

Martin, this is a REAL WORLD NEED expressed by not only many users but also 3 independent parties of the EG. I’m rather surprised for how long we leave the real users alone in the dirt with such issues. 


> @struberg Does OWB provide a non-portable config to allow this right now? I went through OpenWebBeansConfiguration and did not find anything.

Yes, OWB implements this already
https://svn.apache.org/repos/asf/openwebbeans/trunk/webbeans-impl/src/main/java/org/apache/webbeans/config/DeploymentValidationService.java
It looks in system props, env, all openwebbeans.properties it can see and all beans.xml it can see.


> I think the best solution would be to provide non-portable config and not to „pollute" the spec.

Oh that’s a great idea to let users stay in the rain forever…
How does adding a mere 5 lines pollute the spec?


> 
> Other notes:
> 
> 	• I think a regex might be more powerful and less verbose than comma-separated list.

Thought about that as well, but we don’t have regexp for <class> in interceptors, nor in decorators, alternatives etc. 
It’s also only single classes which are affected by this problem. I don’t think any regexp will add much benefit but potentially might mud up the application config.


> 	• System properties are ok for Java SE but environment variables? This is nothing system-specific IMHO.
It’s ok for both SE and EE imo. Seen many things tweaked for the whole EE container that way. Although I agree that I personally also would rather not configure it that way. The only situation coming to my mind is if you move a whole container with all it’s EARs going untouched from Java6 to Java8. Which happens sometimes…


> 	• Again, I don’t like adding allowProxying entry in beans.xml as beans.xml is not a global config file.

As explained in the EG meeting yesterday. I explicitly omitted mentioning any scope or visibility of the configuration in beans.xml. 

In OWB we treat any allowProxying entry as global registry for all EE modules which can see this beans.xml. 
E.g. if you put an allowProxying in an ear/lib jar then it will account for the whole EAR. If you put it in just a single WAR of the EAR then it will only work there. 

But this is the way we impl it in OWB. Weld is free to only let it work in the JAR in which you put it in beans.xml. 
I honestly believe the per-jar handling is crippled and a burden for the users (do I have to mention CDI-18?). But both interpretations are allowed by the way it is worded. 

Of course if Weld folks think hard about it and come to the conclusion that the allowProxying config should rather follow Java visibility rules (like OWB does) then we can explicitly clarify this in the spec as well. 


LieGrue,
strub

mkouba:
Implementation note for JDK classes (e.g. HashMap): Even if we allow final methods we might not be able to create a proxy correctly - there are methods using package-private classes (e.g. `HashMap.newNode()`) and the generated proxy class is not allowed to use package name starting with the identifier `java` (reserved for JDK and throws `SecurityException` at runtime).

struberg:
Of course we will not be able to overwrite those final methods.  That’s the reason why I added the sentence that calling those methods will result in non-portable behaviour. 

LieGrue,
strub

mkouba:
Pls read my comment once again. These methods are not final...

mkouba:
> Yes, OWB implements this already

@struberg Ok. Was it added recently? I cannot see this change in the trunk branch on github.

struberg:
the github mirror seems to be out of sync.
Better check svn.

LieGrue,
strub

antoinesd:
I'm not discussing the usefulness of this, but the more I look, to this the more it looks like an ugly hack. In this form I'd prefer that both implementation agree on implementing it as a non portable behaviour the same way than adding this to the spec.
Maybe the best should be to call for a vote?

struberg:
What's the difference between agreeing and implementing it the same way with all impls and writing it to the spec? Except that the former will be useless for users as they cannot rely on it?

tremes:
That you don't need to necessarily implement it and write TCK test for it, which would basically fail by default and pass with another config - got it right? This doesn't really seem to me straighforward.

mkouba:
I still not convinced this should be specified this and I believe it would be better to leave it as non-portable workaround of implementations. From the user point of view - once [WELD-2094](https://issues.jboss.org/browse/WELD-2094) is finished (Weld 2.3.3) it would be easy to create "almost portable" app - just place two files in your app:
**weld.properties**
````
org.jboss.weld.proxy.ignoreFinalMethods=java.util.HashMap
````
**openwebbeans.properties**:
````
javax.enterprise.inject.allowProxying.classes=java.util.HashMap
````
Integrators (Java EE) may choose to add other ways of configuring this (relying on the impl SPI). And in Java SE (where usually a concrete impl is used) a user might also use system properties, bootstrap API or environment variables in case of OWB.

antoinesd:
@struberg the diff is that we can change this later with a better way. Today, this approach looks like putting duck tape on the spec regarding the needs in real life.
What we can do is to propose it for the annex of the spec would it be a better solution for you ?

antoinesd:
Solution to provide this feature is now totally different (in ProcessBeanAttibutes lifecycle event). A new PR should be proposed.

PR #272, opened 2016-01-23
CDI-578 clarify the behaviour in various error cases

struberg:
on any error due to rollback and illegalstate:
- AFTER_SUCCESS only gets invoked immediately if there is no transaction active
- all other observers get invoked immediately

SystemException doesn't get treated in the spec. This indicates that there is some fundamental setup problem. So I think it's better to let the Exception blow out.

### DIFF
--- a/api/src/main/java/javax/enterprise/event/TransactionPhase.java
+++ b/api/src/main/java/javax/enterprise/event/TransactionPhase.java
@@ -26,10 +26,14 @@ package javax.enterprise.event;
  * Transactional observer methods are observer methods which receive event notifications during the before or after completion
  * phase of the transaction in which the event was fired. If no transaction is in progress when the event is fired, they are
  * notified at the same time as other observers.
+ * If the transaction is in an illegal state or already rolled back when the event is fired, the {@link #BEFORE_COMPLETION},
+ * {@link #AFTER_COMPLETION} and {@link #AFTER_FAILURE} observer methods are notified at the same time as other observers,
+ * but {@link #AFTER_SUCCESS} observer methods get skipped.
  * </p>
  * 
  * @author Pete Muir
  * @author Gavin King
+ * @author Mark Struberg
  * 
  */
 public enum TransactionPhase {
@@ -45,12 +49,30 @@ public enum TransactionPhase {
      * <p>
      * Identifies a before completion observer method, called during the before completion phase of the transaction.
      * </p>
+     * <p>
+     * It will also get invoked if
+     * <ul>
+     *     <li>there is no transaction in progress, or</li>
+     *     <li>the transaction already got rolled back, or</li>
+     *     <li>the transaction is marked for rollback, or</li>
+     *     <li>the transaction is in an illegal state.</li>
+     * </ul>
+     * </p>
      */
     BEFORE_COMPLETION,
 
     /**
      * <p>
      * Identifies an after completion observer method, called during the after completion phase of the transaction.
+     * </p>
+     * <p>
+     * It will also get invoked if
+     * <ul>
+     *     <li>there is no transaction in progress, or</li>
+     *     <li>the transaction already got rolled back, or</li>
+     *     <li>the transaction is marked for rollback, or</li>
+     *     <li>the transaction is in an illegal state.</li>
+     * </ul>
      * </p>
      */
     AFTER_COMPLETION,
@@ -60,6 +82,15 @@ public enum TransactionPhase {
      * Identifies an after failure observer method, called during the after completion phase of the transaction, only when the
      * transaction fails.
      * </p>
+     * <p>
+     * It will also get invoked if
+     * <ul>
+     *     <li>there is no transaction in progress, or</li>
+     *     <li>the transaction already got rolled back, or</li>
+     *     <li>the transaction is marked for rollback, or</li>
+     *     <li>the transaction is in an illegal state.</li>
+     * </ul>
+     * </p>
      */
     AFTER_FAILURE,
 
@@ -67,6 +98,9 @@ public enum TransactionPhase {
      * <p>
      * Identifies an after success observer method, called during the after completion phase of the transaction, only when the
      * transaction completes successfully.
+     * </p>
+     * <p>
+     * It will also get invoked if there is no transaction in progress.
      * </p>
      */
     AFTER_SUCCESS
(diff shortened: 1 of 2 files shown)


api/src/main/java/javax/enterprise/event/TransactionPhase.java
```diff
@@ -26,10 +26,14 @@
  * Transactional observer methods are observer methods which receive event notifications during the before or after completion
  * phase of the transaction in which the event was fired. If no transaction is in progress when the event is fired, they are
  * notified at the same time as other observers.
+ * If the transaction is in an illegal state or already rolled back when the event is fired, the {@link #BEFORE_COMPLETION},
```
manovotn: @struberg Could you please specify what exactly is an 'illegal state' here? I checked the spec, but still can't pinpoint that.

mkouba: What about _"If a transaction is in progress but a JTA `Synchronization` callback cannot be registered (e.g. the transaction is marked for rollback), the BEFORE_COMPLETION..."_

struberg:
Hi Marek, Martin!

I was referring to the Transaction Status. 
https://docs.oracle.com/javaee/6/api/javax/transaction/Status.html

The language is loosely taken from the Synchronisation class
https://docs.oracle.com/javaee/6/api/javax/transaction/Transaction.html#registerSynchronization%28javax.transaction.Synchronization%29

It's perfectly fine to tweak the language to better reflect the outcome. 
My main point is that we are technically pretty much bound to what registerSynchronization allows us. And this blows up with Exceptions in certain cases. And we have to either find a good way to handle it or at least define how we fail in those cases

tremes:
In general I agree with this proposal. Came across this question. When a transaction is marked for rollback then it's in the state when we cannot synchronize. Does it make sense to order transactional observers (something like simulate rollback order) accordingly in this situation?

mkouba:
@tremes That's a good question. I think it makes sense to notify `TransactionPhase.BEFORE_COMPLETION` first and then `AFTER_COMPLETION` and `AFTER_FAILURE` in no defined order.

+1 for tweaking the wording (my proposals are in line comments)

api/src/main/java/javax/enterprise/event/TransactionPhase.java
```diff
@@ -45,6 +49,15 @@
      * <p>
      * Identifies a before completion observer method, called during the before completion phase of the transaction.
      * </p>
+     * <p>
+     * It will also get invoked if
+     * <ul>
+     *     <li>there is no transaction in progress, or</li>
+     *     <li>the transaction already got rolled back, or</li>
```
mkouba: I would remove those these bullets with:

* a transaction is in progress but a JTA Synchronization callback cannot be registered (e.g. the transaction is marked for rollback)

struberg: That wording would then include 'SystemException', right? 
I'd rather leave SystemException bombing out. This usually indicates a serious problem on the server. Not sure neither if my wording catches this well though.

mkouba: Good point. So what about: _a transaction is in progress but a JTA Synchronization callback cannot be registered due to the transaction is already marked for rollback or in a state where Synchronization callbacks cannot be registered_. It's a bit verbose but clear imho...

manovotn:
One thing that came to me after yesterday's meeting.  What about the ordering of event that @tremes suggested? The idea was sort of forgotten which is imho shame.

So the question is:
**Should we impose the ordering of events being fired in this case?**

What would/could the ordering look like:
1) _before completition_  -> first
2) _after completition_ -> second
?) _after failure_ -> this one we needn't define, it can be fired anytime

Why should we do this?
* Well, it feels kind of natural to have _before_ event fired actually prior to the _after_ one
* It simulates the real transaction (failure) flow
* Not doing this will leave another unspecified hole in the spec, while we are at it, we might as well fill this hole now

struberg:
Yes we missed that point. We need to think about it.

From a user pov it would be nice in certain cases.

Otoh a user could easily use @Priority in CDI-2.0 and be done.
If we define this then we would also need to define the cross effect of the 'implicit' ordering in error case vs @Priority. 

So I am fine with leaving this unspecified.

LieGrue,
strub

manovotn:
I think you might be wrong about the @Priority stuff.

You can use priority to order/sort several _after failure_ observers or several _before completion_. But you this won't affect  the ordering of observers which observe different phase. Meaning there is no cross effect here.

E.g. @Priority(1) observer for _after failure_ **will not be notified before** @Priority(2) observer for _after failure_

mkouba:
@struberg `@Priority` can only be used to order transactional observers during certain completion phase (i.e. `beforeCompletion` and `afterCompletion`). E.g. if you define an AFTER_COMPLETION **observer1** with `@Priority(1)` and BEFORE_COMPLETION **observer2** with `@Priority(2)` then **observer2** will be always notified before **observer1** even if observers with smaller priority values are called first.

struberg:
Yes, but in case of those errors they will become kind of 'synchronous'. Thus @Priority will effectively define the order of the observer method invocations.



LieGrue,
strub

mkouba:
@struberg Yes, but the ordering will be different, e.g. AFTER_COMPLETION **observer2** will be notified after BEFORE_COMPLETION **observer1**.

manovotn:
> Thus @Priority will effectively define the order of the observer method invocations.

Not true, this is what I was trying to explain in my comment above. @Priority would only affect ordering of observes within one transactional phase.
In other words, any observer for _before completion_, no matter what priority it has, would be notified sooner than any other _after completition_ observer.

struberg:
@Priority will be applied to the observer method not the whole class. Maybe even the parameter, not sure how the last proposal is worded.
I agree that under normal circumstances the order of invocation is depending on the TX event order.
So even if you have an observer1 during=BEFORE_COMPLETION with @Prio 12 an another observer2 with during=AFTER_SUCCESS with @Prio 1 then observer1 will be called before observer2.

IF all goes ok!

BUT: if there is a problem with registrating the Synchronisation then this will be done in the exact order of the @Prio and thus it impacts the order in which the methods will get invoked

mkouba:
> BUT: if there is a problem with registrating the Synchronisation then this will be done in the exact order...

The whole point of the original proposal from @tremes  and @manovotn is that we could **simulate** the behavior a user would expect. Mainly because the **transaction is still in progress** (although e.g. marked for rollback). The reasoning is very similar to why we skip AFTER_SUCCESS observes - a user would expect this.

antoinesd:
@struberg, @mkouba, I'm not sure of this PR status. It seems stuck, what do we need to move on?

mkouba:
@antoinesd @struberg The proposed PR is conceptually OK but:
1. I believe we should tweak the wording a little bit and rely on JTA terminology
2. there is a proposal from @tremes  and @manovotn to simulate "expected" ordering  - I think it's ok but I'm also ok if we leave this undefined

PR #273, opened 2016-01-29
CDI-515 Allow to obtain bean metadata from Instance

mkouba:


antoinesd:
Looks good for me.

struberg:
just for curiosity: what is it needed for? 
You could get the same via BeanManager. At the end all the Instance stuff is meant for end users. If we introduce Bean<T> then we might blow it up too much?
What about Set<Bean> getBeans() in case it's ambiguous? Should be there but then where to stop?
I think all the Bean<?> handling should be dealt with in the BeanManager and left out of all the end user stuff like Instance.

mkouba:
Yep, it's just a convenient method. And the original intent was to help people with not really well-defined `Instance.destroy()` contract, e.g.:
````
@Inject @Juicy
Instance<FooInterface> fooInstance;  
public void someMethod() {  
          FooInterface foo = fooInstance.get();  
          foo.ping();  
          if (Dependent.class.equals(fooInstance.getBean().getScope())) {
              fooInstance.destroy(foo);  
          }
}  
````
See also https://developer.jboss.org/message/920438

struberg:
I'd rather introduce another method. Something like 'free' for example. 
destroy() -> really destroys the instance. Even normalscoped ones
free() -> mark as 'unused'. Dependent scoped beans will get destroyed, normalscoped ones get ignored.

antoinesd:
At a second thought the only use for that would be to retrieve the scope of the bean (the only info you don't have to provide to resolve the bean in a non ambiguous was).
Wouldn't it be more useful to provide a  `Bean<T> getBean<T> (T instance)` method ? Retrieving the bean from one instance is a very interesting feature.

In the CDI-515 ticket you mention section 5.5.8. Injecting Bean Meta data in instance is a way to retrieve Bean from one of its instance so if we want to provide the same functionality level we should add this IMO.

mkouba:
@antoinesd  https://issues.jboss.org/browse/CDI-515?focusedCommentId=13158110&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13158110

Well, there is another advantage over the BeanManager approach - you don't have to use annotation literals.

Let me think about other use cases. I don't remember all the things I had in mind when I created the issue ;-)

struberg:
Txs Martin, would be interesting. My main point is:
Instance<> is an end user thingy. It should be easy to deal with
Bean<T> is a hacker thingy. This is for the power user. But then they can easily also use the BeanManager for it and we don't need any duplicate functionality.

mkouba:
I cannot remember any other use cases. To sum it up - right now the user can use `BeanManager` (snippet below).

We could:

1. Add `getBean()` convenient method 
2. Add a new method dedicated to destroy `@Dependent` bean instances, e.g. `free()`
3. Add a `destroy` method with boolean param `ignoreClientProxies`
4. Leave it as it is

````java
@Inject @Juicy
Instance<FooInterface> fooInstance;

@Inject
BeanManager beanManager;

public void someMethod() {  
          FooInterface foo = fooInstance.get();  
          foo.ping();
          Bean<?> bean = beanManager.resolve(beanManager.getBeans(new TypeLiteral<FooInterface>() {}, new AnnotationLiteral<Juicy>() {}));
          if (Dependent.class.equals(bean.getScope())) {
              fooInstance.destroy(foo);  
          }
}  
````

mkouba:
@struberg @antoinesd Hola, guys what do you think about adding the `release(T instance)` to destroy `@Dependent` bean instances? I think the name is a bit more appropriate than `free()`. And it 
refers to `CreationalContext.release()`.

antoinesd:
+1 for me

mkouba:
I've created https://issues.jboss.org/browse/CDI-589 and will send a separete pull request.

antoinesd:
This PR is rejected in favor of PR #278.

PR #274, opened 2016-02-02
Cdi 420 adding bean-discovery-mode 'scoped'

struberg:


manovotn:
Seeing PR this quickly surprises me, the relevant issue (CDI-420) hasn't yet been resolved. In fact the discussion barely started. Although people seem to be in favor of this idea, I think sending PR is still a somewhat hasty.

struberg:
This was exactly as proposed by Antoine in the last meeting: don’t discuss much, send pull requests. 
And this ticket was open and has been discussed since 2014… How long do you like to wait?

LieGrue,
strub

manovotn:
Since when does one comment sent on 06/Mar/15 8:12 AM makes a discussion go on since 2014?
Anyway, I haven't checked the last meeting log so you might as well be right on that. 

I barely meant to say that that the actual discussion started few days ago and seems to move in good direction. People agree with this idea and propose different names for it. And before any agreement is reached you sent a PR with your solution. No offense meant, just wanted to underline how weird this feels.

struberg:
Please check the JIRA again:

> Created: 05/Feb/14 5:51 PM

This ticket is now almost to the day TWO years old!

LieGrue,
strub

mkouba:
https://issues.apache.org/jira/browse/DELTASPIKE-1143

struberg:
yes, I gonna ship another PR based on the <trim/> tag we discussed during the F2F meeting.
LieGrue,strub
 

    On Monday, 26 September 2016, 9:15, Martin Kouba <notifications@github.com> wrote:
 
 

 @antoinesd @struberg @manovotn I believe we should close this PR as the EG agreement is to introduce a new tag to tune bean-discovery-mode=all instead of a new discovery mode. —

PR #276, opened 2016-02-15
CDI-485 Literals for built-in annotations

mkouba:
https://issues.jboss.org/browse/CDI-485


antoinesd:
Looks good for me

sknitelius:
Looks good.

mkouba:
Two things I'm not so sure about are:
1. The name of the default instance, e.g. `Any.Literal.DEFAULT` is ok, `Default.Literal.DEFAULT` is not so nice. On the other hand `DEFAULT_INSTANCE` is too long. Maybe `INSTANCE`...
2. Whether we should allow to instantiate literals which currently only supply default instances (`Default`, `Any`, etc.). Currently, it's not allowed.

antoinesd:
> 1. The name of the default instance, e.g. Any.Literal.DEFAULT is ok, Default.Literal.DEFAULT is not so nice. On the other hand DEFAULT_INSTANCE is too long. Maybe INSTANCE...

I prefer INSTANCE. It is the name used in DeltaSpike and late Seam solder.

> 2. Whether we should allow to instantiate literals which currently only supply default instances (Default, Any, etc.). Currently, it's not allowed.

The only interest would be for the sake of consistency with Literals having fields... But I think it's ok like that.

mkouba:
* renamed default instance to `INSTANCE`
* removed most private constructors - it's nothing against nothing

PR #277, opened 2016-03-07
CDI-494 Events - clarify a wildcard type is not an unresolvable type variable

mkouba:
* https://issues.jboss.org/browse/CDI-494
* http://lists.jboss.org/pipermail/cdi-dev/2016-February/007789.html

### DIFF
--- a/spec/src/main/asciidoc/core/events.asciidoc
+++ b/spec/src/main/asciidoc/core/events.asciidoc
@@ -26,7 +26,7 @@ An observer method will be notified of an event if the event object is assignabl
 An event object is an instance of a concrete Java class with no unresolvable type variables.
 The _event types_ of the event include all superclasses and interfaces of the runtime class of the event object.
 
-An event type may not contain an unresolvable type variable.
+An event type may not contain an unresolvable type variable. A wildcard type is not considered an unresolvable type variable.
 
 An event qualifier type is just an ordinary qualifier type as specified in <<defining_qualifier_types>>, typically defined as `@Target({METHOD, FIELD, PARAMETER, TYPE})` or `@Target({FIELD, PARAMETER})`.
 
@@ -88,7 +88,7 @@ public void login() {
 Event fired with the `fire()` method is fired synchronously.
 All the resolved synchronous observers (as defined in <<observer_resolution>>) are called in the thread in which `fire()` was called.
 A synchronous observer notification blocks the calling thread until it completes.
- 
+
 [[firing_events_asynchronously]]
 
 ==== Firing events asynchronously
@@ -119,15 +119,15 @@ The `Event` interface provides a method for firing events with a specified combi
 [source, java]
 ----
 public interface Event<T> {
-      
+
     public void fire(T event);
     public <U extends T> CompletionStage<U> fireAsync(U event);
     public <U extends T> CompletionStage<U> fireAsync(U event, Executor executor);
-      
+
     public Event<T> select(Annotation... qualifiers);
     public <U extends T> Event<U> select(Class<U> subtype, Annotation... qualifiers);
     public <U extends T> Event<U> select(TypeLiteral<U> subtype, Annotation... qualifiers);
-      
+
 }
 ----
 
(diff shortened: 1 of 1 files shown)


manovotn:
+1 from me

Though I was thinking a  tiny code snippet showing some of the possibilities it gives would not hurt, WDYT?

mkouba:
Sure, this would enable e.g. the following code to be portable: 
```java
@Inject Event<List<?>> event;

void process() {
  List<String> stringList = new ArrayList<>();
  event.fire(stringList);
  List<Integer> intList = new ArrayList<>();
  event.fire(intList);
}

void observeAllLists(@Observes List<?> anyList) {
  ...
}
```
Note that in this example the runtime type of events is `ArrayList<E>` and `E` could be resolved to `?`.

manovotn:
I meant the snippet for spec, as talking wildcards can get a bit wild at time and not every reader will grasp this without a simple sample. On the other hand it might be unnecessary, we can wait and see what others think :)

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -26,7 +26,7 @@ An observer method will be notified of an event if the event object is assignabl
 An event object is an instance of a concrete Java class with no unresolvable type variables.
 The _event types_ of the event include all superclasses and interfaces of the runtime class of the event object.
 
-An event type may not contain an unresolvable type variable.
+An event type may not contain an unresolvable type variable. A wildcard type is not considered an unresolvable type variable.
```
antoinesd: Be careful here. In JLS Wildcard type are unbound (<?>) or bound (<? extends Number>) are you sure you mean both here?

antoinesd:
+1 for the PR if we talk about unbounded wildcard. Bounded wildcard would be nice, but I don't see how we could define them as resolvable type when all type resolution in the spec don't accept it.

spec/src/main/asciidoc/core/events.asciidoc
mkouba: Yes, for both. `<?>` translates to `<? extends Object>`.

mkouba:
> ...when all type resolution in the spec don't accept it

What exactly do you mean?

antoinesd:
All type resolution section in the spec, only mention unbounded wildcard like here:
http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#assignable_parameters
here:
http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#bean_metadata

mkouba:
Hm, I think the spec always mentions the wildcard bounds... And as I wrote above - the wildcard `? extends Object` is equivalent to the unbounded wildcard `?` (JLS "4.5.1 Type Arguments of Parameterized Types").

spec/src/main/asciidoc/core/events.asciidoc
antoinesd: `<? extends Object>` means any type while `<? extends Number>` means any type extending number, I think the restriction makes a lot of difference here, hence the therm "bounded" in JLS.

Now, if there are no technical problem at the impl level to support bounded wildcard, why not...

antoinesd:
We're saying the same: the spec only accept (and mention) **unbounded wildcards**. So I only pointed that allowing **bounded wildcards** should be done carefully (even if events are not the same than typesafe resolution).

mkouba:
The same except the other way around ;-)  The spec only mentions bounded wildcards...

antoinesd:
ok, my mistake :-).
Does the mention below (in 5.2.3) could be an inspiration for this PR?

> Any legal bean type, as defined in Legal bean types may be the required type of an injection point. Furthermore, **the required type of an injection point may contain a wildcard type parameter**. However, a type variable is not a legal injection point type.

mkouba:
See also https://github.com/mkouba/event-wildcard-test

slaskawi:
+1 for this feature - we need it in Infinispan (oftentimes when we fire event we don't know their types - they are use driven).

antoinesd:
+1 for me as well

antoinesd:
@slaskawi very interesting to have feedback from users. thx.
Unless we missed a big issue with this approach I'm +1 for it

johnament:
+1 from me as well.

PR #282, opened 2016-03-11
CDI-508 clear up conditions of which Java classes are managed beans.

tremes:


tremes:
Maybe we can use term `complete` class instead of non-abstract.

manovotn:
IMHO ```non-abstract``` is a better choice. To me it seems more accurate.

spec/src/main/asciidoc/core/implementation.asciidoc
```diff
@@ -30,10 +30,10 @@ If the managed bean class is a generic type, it must have scope `@Dependent`. If
 
 ==== Which Java classes are managed beans?
 
-A top-level Java class is a managed bean if it meets all of the following conditions:
+A Java class is a managed bean if it meets all of the following conditions:
 
-* It is not a non-static inner class.
-* It is a concrete class, or is annotated `@Decorator`.
+* It is not a non-static nested class.
```
johnament: This doesn't sound quite right.  Maybe its current spec behavior, but I would expect that static, not 'non-static' classes are eligible. Or maybe the sentence should be rephrased as

`It is a static nested class`

manovotn: @johnament  The change you suggested does not have the same meaning.
```It is not a non-static nested class.``` - this statement explicitly states that one specific case is forbidden.
```It is a static nested class``` - this statement explicitly states that one specific case is permitted.

E.g. it would require more changes to preserve the meaning (such as linking the sentences with ```or```)
So even though the version @tremes created may sound clumsy from language point of view, it is imho precise.

tremes:
I updated the wording according to our email discusssion at cdi-dev mailing list (see http://lists.jboss.org/pipermail/cdi-dev/2016-May/008221.html)

johnament:
I agree with the change, anyone concerned with merging?

antoinesd:
Nope, merging after a rebase from @tremes

PR #284, opened 2016-03-11
CDI-571 producer and disposer methods can be only local bussines meth…

tremes:
…od of EJB.


mkouba:
I'm not sure `local business method` is the best name. I've searched the EJB spec and it's only used twice (9.4.2.3) and apparently in a different sense. I think we should follow the terminology from "3.2 Local, Remote, and Web Service Client Views", although it's a bit verbose. E.g. something like `business method exposed by a local business interface or by a no-interface view`...

tremes:
Yes that's true and sounds reasonable. I changed the all related wording.

antoinesd:
+1 for me

johnament:
+1 from me

PR #286, opened 2016-03-21
CDI-589 Add Instance.release() method

mkouba:


api/src/main/java/javax/enterprise/inject/Instance.java
```diff
@@ -203,5 +203,17 @@
      *         destroying bean instances
      */
     public void destroy(T instance);
+    
+    /**
+     * <p>
+     * Instructs the container to destroy the instance. The instance passed should be a dependent scoped bean instance
```
manovotn: ```should be``` sounds weird here.
I suppose that if it _isn't_ a dependent scoped bean, you just ignore it.
If that is the case, I would rather say ```... instance passed has to be a dependent ...```.

mkouba: It's intentional. If we say `has to be dependent` it should probably fail if you pass any other instance (e.g. throw IAE). And we would like to allow users to pass anything and only destroy dependent bean instances. But I'm not native speaker so my interpretation might be wrong ;-)

antoinesd:
I don't know how to put it, but I think we should find a way to make clear why `release` was  introduced when `destroy` existed.

mkouba:
Well, we could mention the difference and describe the use case in javadoc/spec text. Or we could  introduce `destroy(T instance, boolean ignoreClientProxies)` instead. But I like `release()` more.

api/src/main/java/javax/enterprise/inject/Instance.java
manovotn: Hmm, I see. In that case it makes sense.
Although we might want to name this method so that it indicates the attempt to destroy the instance, wdyt? Something like ```tryToRelease```? Just to draw some clear line between ```destroy``` method and this one from user point of view.

johnament:
I don't see why we need a release method when we already have a destroy method, when both seem to do the same thing.

mkouba:
@johnament See also https://issues.jboss.org/browse/CDI-515 and https://github.com/cdi-spec/cdi/pull/273

johnament:
I would be more in favor of fixing the spec to allow you to destroy dependent instances that this instance didn't create.

mkouba:
> to destroy dependent instances that this instance didn't create...

I believe this is not possible. If you don't have the `CreationalContext` used during dependent instance creation you can't do that.

mkouba:
This is what we're working on in Weld:
http://weld.cdi-spec.org/news/2016/05/18/enhanced-instance/

antoinesd:
@mkouba re-reading the whole use case and how it was addressed in Weld I wonder why you didn't propose to standardize WeldInstance ?
For me, it adds more sepratio between simple/advanced usage and we gain the access to bean metadata that was lost when passing from CDI-515 to CDI-589.

mkouba:
I had a feeling that adding anything with `Bean<?>` on `Instance` is not desired. Anyway, we can discuss this at F2F - we plan to release 2.4.CR1 in early October.

mkouba:
By the way, this the final proposal of Weld API:
https://github.com/weld/api/blob/2.4/weld/src/main/java/org/jboss/weld/inject/WeldInstance.java

antoinesd:
I add it to our Agenda

johnament:
I would go with the handler approach defined by Weld.  This PR looks bad IMHO.  Can we close it?

antoinesd:
I agree @johnament. It feels half mesure, where WeldInstance contains all the useful feature. Do you agree to close it @mkouba and  go for WeldInstance standardization?

mkouba:
Sure, close the PR.

PR #287, opened 2016-04-04
CDI-558 Standardize the Meta-Data builders API (BeanBuilder, etc...)

antoinesd:
New PR for Configurators and Builders API

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
@@ -17,6 +17,8 @@
 package javax.enterprise.inject.spi;
 
 import javax.enterprise.context.spi.Context;
+import javax.enterprise.inject.spi.builder.BeanConfigurator;
+import javax.enterprise.inject.spi.builder.ObserverMethodConfigurator;
 
 /**
  * <p>
@@ -41,6 +43,7 @@ import javax.enterprise.context.spi.Context;
  * </p>
  * 
  * @author David Allen
+ * @author Antoine Sabot-Durand
  */
 public interface AfterBeanDiscovery {
     /**
@@ -64,6 +67,21 @@ public interface AfterBeanDiscovery {
     public void addBean(Bean<?> bean);
 
     /**
+     *
+     * Obtains a new {@link BeanConfigurator} to configure  a new {@link Bean}  and add it at the end of the observer invocation.
+     * It will then fire an event of type {@link javax.enterprise.inject.spi.ProcessBean} containing the built
+     * {@link javax.enterprise.inject.spi.Bean} from this configuration and then register it with the
+     * container, thereby making it available for injection into other beans.
+     *
+     * Each call returns a new BeanConfigurator.
+     *
+     * @return a non reusable {@link BeanConfigurator} to configure the bean to add
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public <T> BeanConfigurator<T> addBean();
+
+    /**
      * Fires an event of type {@link javax.enterprise.inject.spi.ProcessObserverMethod} containing the given
      * {@link javax.enterprise.inject.spi.ObserverMethod} and then registers the
      * {@link javax.enterprise.inject.spi.ObserverMethod} with the container, thereby making it available for event
@@ -73,6 +91,21 @@ public interface AfterBeanDiscovery {
      * @throws IllegalStateException if called outside of the observer method invocation
      */
     public void addObserverMethod(ObserverMethod<?> observerMethod);
+
+    /**
+     * obtains a new {@link ObserverMethodConfigurator} to configure a new {@link ObserverMethod} and add it at the end of the observer invocation.
+     * It will then fire an event of type {@link javax.enterprise.inject.spi.ProcessObserverMethod} containing the built
+     * {@link javax.enterprise.inject.spi.ObserverMethod} from this configuration and then registers the
+     * {@link javax.enterprise.inject.spi.ObserverMethod} with the container, thereby making it available for event
+     * notifications.
+     *
+     * Each call returns a new ObserverMethodConfigurator.
+     *
+     * @return a non reusable {@link ObserverMethodConfigurator} instance
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public <T> ObserverMethodConfigurator<T> addObserverMethod();
 
     /**
      * Registers a custom {@link javax.enterprise.context.spi.Context} object with the container.
@@ -106,5 +139,4 @@ public interface AfterBeanDiscovery {
      * @since 1.1
      */
     public <T> Iterable<AnnotatedType<T>> getAnnotatedTypes(Class<T> type);
-
 }
--- a/api/src/main/java/javax/enterprise/inject/spi/AfterTypeDiscovery.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/AfterTypeDiscovery.java
@@ -17,6 +17,7 @@
 
 package javax.enterprise.inject.spi;
 
+import javax.enterprise.inject.spi.builder.AnnotatedTypeConfigurator;
 import java.util.List;
 
 /**
@@ -32,6 +33,7 @@ import java.util.List;
  * </p>
  *
  * @author Pete Muir
+ * @author Antoine Sabot-Durand
  * @since 1.1
  */
 public interface AfterTypeDiscovery {
@@ -75,5 +77,34 @@ public interface AfterTypeDiscovery {
      * @throws IllegalStateException if called outside of the observer method invocation
      */
     public void addAnnotatedType(AnnotatedType<?> type, String id);
+
+    /**
+     * <p>
+     * Obtains a new {@link AnnotatedTypeConfigurator} to configure a new {@link javax.enterprise.inject.spi.AnnotatedType} and
+     * add it to the set of types which will be scanned during bean discovery at the end of the observer invocation.
+     * Calling this method multiple times will return a new AnnotatedTypeConfigurator.
+     * </p>
+     *
+     * <p>
+     * This method allows multiple annotated types, based on the same underlying type, to be defined. {@link AnnotatedType}s
+     * discovered by the container use the fully qualified class name of {@link AnnotatedType#getJavaClass()} to identify the
+     * type.
+     * </p>
+     *
+     * <p>
+     * {@link AfterBeanDiscovery#getAnnotatedType(Class, String)} and {@link AfterBeanDiscovery#getAnnotatedTypes(Class)} allows
+     * annotated types to be obtained by identifier.
+     * </p>
+     *
+     * Each call returns a new AnnotatedTypeConfigurator.
+     *
+     *
+     * @param id The id of the annotated type
+     * @param type type of the class that the {@link AnnotatedType} will represent
+     * @return a non reusable {@link AnnotatedTypeConfigurator} to configure the new AnnotatedType
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public <T> AnnotatedTypeConfigurator<T> addAnnotatedType(String id, Class<T> type);
 
 }
--- a/api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedConstructorConfigurator.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedConstructorConfigurator.java
@@ -1,0 +1,95 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package javax.enterprise.inject.spi.builder;
+
+import java.lang.annotation.Annotation;
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+import javax.enterprise.inject.spi.AnnotatedConstructor;
+import javax.enterprise.inject.spi.AnnotatedParameter;
+import javax.enterprise.util.Nonbinding;
+
+/**
+ *
+ * This interface is part of the {@link AnnotatedTypeConfigurator} spi and helps defining an {@link AnnotatedConstructor}
+ * 
+ * @author Martin Kouba
+ * @author Antoine Sabot-Durand
+ * @since 2.0
+ * @param <T> the class declaring the constructor
+ */
+public interface AnnotatedConstructorConfigurator<T> {
+
+    /**
+     * 
+     * @return the original {@link AnnotatedConstructor}
+     */
+    AnnotatedConstructor<T> getAnnotated();
+
+    /**
+     * Add an annotation to the constructor.
+     * 
+     * @param annotation annotation to add
+     * @return self
+     */
+    AnnotatedConstructorConfigurator<T> add(Annotation annotation);
+
+    /**
+     * Remove all annotation with (a) the same type and (b) the same annotation member value for each member which is not
+     * annotated {@link Nonbinding}. The container calls the {@link Object#equals(Object)} method of the annotation member value
+     * to compare values.
+     * 
+     * @param annotation annotation to remove
+     * @return self
+     */
+    AnnotatedConstructorConfigurator<T> remove(Annotation annotation);
+
+    /**
+     * Removes all annotations with the same type. Annotation members are ignored.
+     * 
+     * @param annotationType annotation class to remove
+     * @return self
+     */
+    AnnotatedConstructorConfigurator<T> remove(Class<? extends Annotation> annotationType);
+
+    /**
+     * Remove all annotations from the constructor.
+     * 
+     * @return self
+     */
+    AnnotatedConstructorConfigurator<T> removeAll();
+
+    /**
+     * 
+     * @return an immutable list of {@link AnnotatedParameterConfigurator}s reflecting the
+     *         {@link AnnotatedConstructor#getParameters()}
+     */
+    List<AnnotatedParameterConfigurator<T>> params();
+
+    /**
+     * 
+     * @param predicate Testing the original {@link AnnotatedParameter}
+     * @return a sequence of {@link AnnotatedParameterConfigurator}s matching the given predicate
+     * @see AnnotatedParameterConfigurator#getAnnotated()
+     */
+    default Stream<AnnotatedParameterConfigurator<T>> filterParams(Predicate<AnnotatedParameter<T>> predicate) {
+        return params().stream().filter(p -> predicate.test(p.getAnnotated()));
+    }
+
+}
--- a/api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedMethodConfigurator.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedMethodConfigurator.java
@@ -1,0 +1,95 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package javax.enterprise.inject.spi.builder;
+
+import java.lang.annotation.Annotation;
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+import javax.enterprise.inject.spi.AnnotatedMethod;
+import javax.enterprise.inject.spi.AnnotatedParameter;
+import javax.enterprise.util.Nonbinding;
+
+/**
+ *
+ * This interface is part of the {@link AnnotatedTypeConfigurator} spi and helps defining an {@link AnnotatedMethod}
+ * 
+ * @author Martin Kouba
+ * @author Antoine Sabot-Durand
+ * @since 2.0
+ * @param <T> the class declaring the method
+ */
+public interface AnnotatedMethodConfigurator<T> {
+
+    /**
+     * 
+     * @return the original {@link AnnotatedMethod}
+     */
+    AnnotatedMethod<T> getAnnotated();
+
+    /**
+     * Add an annotation to the field.
+     *
+     * @param annotation to add
+     * @return self
+     */
+    AnnotatedMethodConfigurator<T> add(Annotation annotation);
+
+    /**
+     * Remove annotations with (a) the same type and (b) the same annotation member value for each member which is not
+     * annotated {@link Nonbinding}. The container calls the {@link Object#equals(Object)} method of the annotation member value
+     * to compare values.
+     *
+     * @param annotation to remove
+     * @return self
+     */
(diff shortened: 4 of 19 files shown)


api/src/main/java/javax/enterprise/inject/spi/AfterTypeDiscovery.java
```diff
@@ -76,4 +78,32 @@
...
+     *
+     * Each call returns a new AnnotatedTypeConfigurator.
+     *
+     *
+     * @param id The id of the annotated type
+     * @return a non reusable {@link AnnotatedTypeConfigurator} to configure the new AnnotatedType
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public <T> AnnotatedTypeConfigurator<T> addAnnotatedType(String id);
```
mkouba: `Class<T> type` argument is missing. Note that we've removed all the `read()` methods from `AnnotatedTypeConfigurator`. The same applies to `BeforeBeanDiscovery`.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnnotatedConfiguratorsTest.java
```diff
@@ -0,0 +1,88 @@
+package javax.enterprise.inject.spi.builder;
+
+import javax.enterprise.event.Observes;
+import javax.enterprise.inject.spi.ProcessAnnotatedType;
+import javax.enterprise.util.AnnotationLiteral;
+import javax.inject.Inject;
+
+public class AnnnotatedConfiguratorsTest {
```
mkouba: Remove this class. It was intended just to demonstrate the usage of the `AnnotatedTypeConfigurator` API.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfiguratorOld.java
```diff
@@ -0,0 +1,432 @@
...
+ * This configurator is not thread safe and shall not be used concurrently.
+ *
+ * @see javax.enterprise.inject.spi.BeforeBeanDiscovery#addAnnotatedType(String)
+ * @see javax.enterprise.inject.spi.AfterTypeDiscovery#addAnnotatedType(String)
+ * @see ProcessAnnotatedType#configureAnnotatedType()
+ * @param <T> the class represented by the configured AnnotatedType
+ * @author Antoine Sabot-Durand
+ * @since 2.0
+ */
+public interface AnnotatedTypeConfiguratorOld<T> {
```
mkouba: Remove this interface. It's just a "backup" - an old version of `AnnotatedTypeConfigurator` API.

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -882,6 +885,29 @@ If any observer method of the `BeforeBeanDiscovery` event throws an exception, t
...
+* Add or remove annotation on a field with `addToField` and `removeFromField` methods.
+* Add or remove annotation on a method with `addToMethod` and `removeFromMethod` methods.
+* Add or remove annotation on a method parameter with `addToMethodParameter` and `removeFromMethodParameter` methods.
+* Add or remove annotation on a constructor with `addToConstructor` and `removeFromConstructor` methods.
+* Add or remove annotation on a constructor parameter with `addToConstructorParameter` and `removeFromConstructorParameter` methods.
+* Remove an annotation from the type or any type member with `removeFromAll` method.
+* Override a field type with `overrideFieldType` methods.
+* Override a method parameter type with `overrideMethodParameterType` method.
+* Override a constructor parameter type with `overrideConstructorParameterType` method.
+* Change the class that the `AnnotatedType` will represent.
```
mkouba: The list is outdated - e.g. override operations are not available in the new API proposal, on the other hand it's easy to remove annotation from all methods of the class.

antoinesd: yes, that's why I said in my mail to only review API, not spec for the moment.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnnotatedConfiguratorsTest.java
antoinesd: Came back during the rebase :-(

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfiguratorOld.java
antoinesd: same here

spec/src/main/asciidoc/core/spi.asciidoc
mkouba: I'm sorry, I guess I missed that ;-)

antoinesd: NP. I have to find a better way to introduce those in the spec. Finally, I'm moving it in the Alternative metadata source section.

api/src/main/java/javax/enterprise/inject/spi/AfterBeanDiscovery.java
```diff
@@ -75,6 +93,22 @@
...
+     * notifications.
+     *
+     * Each call returns a new ObserverMethodConfigurator.
+     *
+     * @return a non reusable {@link ObserverMethodConfigurator} instance
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public <T> ObserverMethodConfigurator<T> addObserverMethod();
+
```
mp911de: superfluous blank line 🙃

api/src/main/java/javax/enterprise/inject/spi/builder/Builders.java
```diff
@@ -0,0 +1,113 @@
...
+ * @see BuildersProvider
+ * @author Antoine Sabot-Durand
+ * @since 2.0
+ */
+public final class Builders {
+
+    /**
+     * {@link BuildersProvider} instance loader from {@link ServiceLoader}
+     */
+    protected static volatile BuildersProvider buildersProvider = null;
```
mp911de: Why is `buildersProvider` protected?

api/src/main/java/javax/enterprise/inject/spi/builder/Builders.java
```diff
@@ -0,0 +1,113 @@
...
+     * Obtains a {@link ObserverMethodBuilder} for the given event type
+     *
+     * @param type the type of the event for the built ObserverMethod
+     * @param <T> generic for type
+     * @return a reusable ObserverMethodBuilder
+     */
+    public static <T> ObserverMethodBuilder<T> observerMethod(Class<T> type) {
+        return provider().getObserverMethodBuilder(type);
+    }
+
```
mp911de: superfluous blank line 🙃

api/src/main/java/javax/enterprise/inject/spi/builder/BuildersProvider.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Obtains an {@link AnnotatedTypeBuilder} for the given type
+     *
+     * @param type the type that the AnnotatedType to build will represent
+     * @param <T> generic for type
+     * @return a reusable AnnotatedTypeBuilder for the given type
+     */
+    <T> AnnotatedTypeBuilder<T> getAnnotatedTypeBuilder(Class<T> type);
+
```
mp911de: superfluous blank line 🙃

api/src/main/java/javax/enterprise/inject/spi/builder/BuildersProvider.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Interface implemented by the CDI provider to give access to builders implementations
+ * Implementation will be loaded with {@link java.util.ServiceLoader}
+ *
+ * @see Builders
+ * @author Antoine Sabot-Durand
+ * @since 2.0
+ */
+public interface BuildersProvider {
```
mp911de: `BuildersProvider` or `BuilderProvider`?

api/src/main/java/javax/enterprise/inject/spi/builder/BuildersProvider.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Obtains a {@link ObserverMethodBuilder} for the given event type
+     *
+     * @param type the type of the event for the built ObserverMethod
+     * @param <T> generic for type
+     * @return a reusable BeanBuilder
+     */
+    <T> ObserverMethodBuilder<T> getObserverMethodBuilder(Class<T> type);
+
```
mp911de: superfluous blank line 🙃

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointConfigurator.java
```diff
@@ -0,0 +1,187 @@
...
+ * CDI container must provides an implementation of this interface.
+ *
+ * This builder is not thread safe and shall not be used concurrently.
+ *
+ * @see ProcessInjectionPoint#configureInjectionPoint()
+ * @author Antoine Sabot-Durand
+ * @since 2.0
+ */
+public interface InjectionPointConfigurator {
+
```
mp911de: superfluous blank line 🙃

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodConfigurator.java
```diff
@@ -0,0 +1,181 @@
...
+import java.lang.reflect.Type;
+import java.util.Set;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+
+/**
+ * This API is an helper to build a new {@link ObserverMethod} instance.
+ * CDI container must provides an implementation of this interface accessible.
+ *
+ * This configurator is not thread safe and shall not be used concurrently.
```
mp911de: How about aligning this sentence with wording from the JDK? ("If multiple threads access the configurator it must be synchronized externally")

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodConfigurator.java
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * @param <T> type of the event the configured ObserverMethod will observe
+ * @author Antoine Sabot-Durand
+ * @see AfterBeanDiscovery#addObserverMethod()
+ * @since 2.0
+ */
+public interface ObserverMethodConfigurator<T> {
+
```
mp911de: superfluous blank line 🙃

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedConstructorConfigurator.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    List<AnnotatedParameterConfigurator<T>> params();
+
+    /**
+     * 
+     * @param predicate Testing the original {@link AnnotatedParameter}
+     * @return a sequence of {@link AnnotatedParameterConfigurator}s matching the given predicate
+     * @see AnnotatedParameterConfigurator#getAnnotated()
+     */
+    default Stream<AnnotatedParameterConfigurator<T>> filterParams(Predicate<AnnotatedParameter<T>> predicate) {
```
mp911de: What's the use case for this method?

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedMethodConfigurator.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    List<AnnotatedParameterConfigurator<T>> params();
+
+    /**
+     * 
+     * @param predicate Testing the original {@link AnnotatedParameter}
+     * @return a sequence of {@link AnnotatedParameterConfigurator}s matching the given predicate
+     * @see AnnotatedParameterConfigurator#getAnnotated()
+     */
+    default Stream<AnnotatedParameterConfigurator<T>> filterParams(Predicate<AnnotatedParameter<T>> predicate) {
```
mp911de: What's the use case for this method?

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
```diff
@@ -0,0 +1,136 @@
...
+     * 
+     * @return self
+     */
+    AnnotatedTypeConfigurator<T> removeAll();
+
+    /**
+     * 
+     * @return an immutable set of {@link AnnotatedMethodConfigurator}s reflecting the {@link AnnotatedType#getMethods()}
+     */
+    Set<AnnotatedMethodConfigurator<T>> methods();
```
mkouba: The return type should be `Set<AnnotatedMethodConfigurator<? super T>>`

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    default Stream<AnnotatedMethodConfigurator<T>> filterMethods(Predicate<AnnotatedMethod<T>> predicate) {
+        return methods().stream().filter(c -> predicate.test(c.getAnnotated()));
+    }
+
+    /**
+     * 
+     * @return an immutable set of {@link AnnotatedFieldConfigurator}s reflecting the {@link AnnotatedType#getFields()}
+     */
+    Set<AnnotatedFieldConfigurator<T>> fields();
```
mkouba: The return type should be `Set<AnnotatedFieldConfigurator<? super T>>`

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedConstructorConfigurator.java
mkouba: All these filter methods are just convenient shortcuts so that a user can use e.g.:
```
event.configureAnnotatedType().filterMethods((m) -> m.getParameters().size() > 0).forEach(m -> ...)
```
instead of:
```
event.configureAnnotatedType().params().stream().filter(c -> c.getAnnotated().getParameters().size() > 0).forEach(m -> ...)
```

manovotn:
I stumbled across a pretty weird chicken-egg style problem with *BeanConfigurator*.

Say you are turning an invalid bean (e.g. having a parametrized constructor) into a valid one via *AfterBeanDiscovery* event and BeanConfigurator. You also want to add some **new** injection point(s).
Now, calling *configurator.addInjectionPoint()* requires you to have an IP point prepared. But to prepare an IP, you need to give it an instance of *Bean<T>*. But you do not yet have such instance, cause you are just in the creation process. See the problem? There is no way to achieve that at the moment

One way to fix that would be to add a method such as *InjectionPointConfigurator<T> addInjectionPoint();* (note the **Injection**PointConfigurator there), where you could set anything you like (such as field, qualifier, type) and the *Bean<T>* would be set automatically once your bean has been created.

I came across this while playing with the TCK tests for this. The situation can be seen here -> https://github.com/manovotn/cdi-tck/blob/5586b6c804c026dbaff87d15d8ca1b48f00c1b6c/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/bean/AfterBeanDiscoveryObserver.java#L67

mkouba:
`BeanAttributesConfigurator.read()` methods probably don't make much sense as the configurator will only be used during `ProcessBeanAttributes` notification. The existence of `BeanAttributesBuilder` is also questionable.

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointConfigurator.java
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * Read the InjectionPoint information from the given {@link Method}.
+     * All relevant information is overwritten.
+     *
+     * @param method defining the InjectionPoint.
+     * @param param  the parameter's index in the Method defining the InjectionPoint
+     * @return self
+     */
+    InjectionPointConfigurator read(Method method, int param);
```
mkouba: This and `read(Constructor<?> constructor, int param)` could be replaced with single method:  `read(java.lang.reflect.Parameter param)`.

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodConfigurator.java
```diff
@@ -0,0 +1,181 @@
...
+    ObserverMethodConfigurator<T> read(ObserverMethod<T> method);
+
+    /**
+     * Set the class of the Bean containing this observer.
+     * If not set, the extension class is used.
+     *
+     * @param type the bean class containing this configurator.
+     * @return self
+     */
+    ObserverMethodConfigurator<T> beanClass(Class<?> type);
```
tremes: I am not sure we should allow this. This seems to be forbidden in Weld since ages. See https://github.com/cdi-spec/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/observerMethod/ProcessObserverMethodObserver.java#L38

mkouba: The point is we should allow this for `AfterBeanDiscovery.addObserverMethod()` or say that `ObserverMethod.getBeanClass()` will return the extension class for this case or something like that...

api/src/main/java/javax/enterprise/inject/spi/builder/Builders.java
antoinesd: To ease tests.
It's like the fields in CDI class...
What visibility did you expect?

mp911de: `private` How about using reflection for tests?

api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointConfigurator.java
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     *
+     * Add the qualifier to the InjectionPoint to build
+     * If the builder declares the {@link Default} qualifier, it's automatically removed.
+     *
+     * @param qualifier the qualifier to add
+     * @return self
+     */
+    InjectionPointConfigurator addQualifier(Annotation qualifier);
```
tremes: This is little bit problematic. When I add some additional qualifier during PIP the processed IP still have also `@Default` which is not desirable. Shouldn't we go only with "replaceable" methods in this case??

api/src/main/java/javax/enterprise/inject/spi/builder/BeanAttributesConfigurator.java
```diff
@@ -0,0 +1,230 @@
...
+     *
+     * @param scope new scope for the configured bean
+     * @return self
+     */
+    BeanAttributesConfigurator<T> scope(Class<? extends Annotation> scope);
+
+    /**
+     *
+     * Add a qualifier to the configured bean
+     * If the builder declares the {@link Default} qualifier, it's automatically removed.
```
tremes: It's not a builder. We should state (explicitly in the spec) that `@Default` is not removed when you add `@Named`.

api/src/main/java/javax/enterprise/inject/spi/builder/ObserverMethodConfigurator.java
```diff
@@ -0,0 +1,181 @@
...
+     * Set the type of the observed event
+     *
+     * @param type of the observed event
+     * @return self
+     */
+    ObserverMethodConfigurator<T> observedType(Type type);
+
+    /**
+     * Add the qualifier to the observed event
+     * If the builder declares the {@link Default} qualifier, it's automatically removed.
```
tremes: Not a builder again + mentioning `@Default` is not necessary here.

api/src/main/java/javax/enterprise/inject/spi/builder/AnnotatedTypeConfigurator.java
antoinesd: Yes but then `filterMethods` won't compile any more

mkouba: I see, then we should change `filterMethods` as well :-)
````
default Stream<AnnotatedMethodConfigurator<? super T>> filterMethods(Predicate<AnnotatedMethod<? super T>> predicate) {
        return methods().stream().filter(c -> predicate.test(c.getAnnotated()));
}
```

antoinesd: ok, thx

antoinesd:
> UPDATE: On the other hand, the builder and `read(BeanAttributes<?> beanAttributes)` method might be useful when using `BeanManager.createBean(BeanAttributes<T>, Class<T>, InjectionTargetFactory<T>)`.

@mkouba, I think `read()` methods should be removed and we should find a way to initialize a `BeanAttruibtesBuilder` with an existing `BeanAttributes`. That brings back the question of obtaining builders...

johnament:
I see a few things missing in this PR... maybe its lacking, or hasn't been thought through yet.  Here's a couple of ported examples:

Apache Camel CDI - registration of a default camel context.  Current code: https://github.com/apache/camel/blob/camel-2.17.x/components/camel-cdi/src/main/java/org/apache/camel/cdi/CdiCamelExtension.java#L223

I think all I have to do is replace it with this method call:

```java
// Add a @Default Camel context bean if any
abd.addBean()
        .addType(DefaultCamelContext.class)
        // .addScope(Dependent.class) why is there no add scope?
        .addQualifiers(AnyLiteral.INSTANCE, DefaultLiteral.INSTANCE)
```

Which this looks straight forward.  but how do i or CDI know that you're done building the object? I couldn't find a build() or close() type of method.  I get the `addQualifiers` since its additive by declaration, but I don't get the `addType` method.  A bean should only have one type? Or many types?  Wouldn't it be inferred from the registered type?

Second, I ported an annotated method method, original impl here: https://github.com/apache/camel/blob/camel-2.17.x/components/camel-cdi/src/main/java/org/apache/camel/cdi/CdiCamelExtension.java#L169

```java
    private void camelFactoryProducers(@Observes ProcessAnnotatedType<CdiCamelFactory> pat, BeanManager manager) {
        AnnotatedType<CdiCamelFactory> at = pat.getAnnotatedType();

        Set<AnnotatedMethod<? super CdiCamelFactory>> methods = pat
                .configureAnnotatedType()
                .filterMethods(CdiCamelExtension::isValidMethod)
                .map(amc -> {
                    AnnotatedMethod<? super CdiCamelFactory> am = amc.getAnnotated();
                    Set<Annotation> qualifiers = CdiSpiHelper.getQualifiers(am, manager);
                    producerQualifiers.put(am.getJavaMember(), qualifiers);
                    Set<Annotation> annotations = new HashSet<>(am.getAnnotations());
                    annotations.removeAll(qualifiers);
                    annotations.add(Excluded.INSTANCE);
                    return new AnnotatedMethodDelegate<>(am, annotations);
                })
                .collect(toSet());
        pat.setAnnotatedType(new AnnotatedTypeDelegate<>(at, methods));
    }

    private static boolean isValidMethod(AnnotatedMethod<?> am) {
        Class<?> type = CdiSpiHelper.getRawType(am.getBaseType());
        return     am.isAnnotationPresent(Produces.class)
                || Endpoint.class.isAssignableFrom(type)
                || ConsumerTemplate.class.equals(type)
                || ProducerTemplate.class.equals(type);
    }
```

This approach is more readable, but loses some flavor because of the switching between classes.  Probably the ugly part is that I need a bean manager to tell me if an annotation is a qualifier or not.  There may be some syntactic sugar we can add to remove that dependency.  But the bottom line is that I couldn't figure out a clean way to manipulate the annotated method to remove annotations, add annotations.  Maybe add a method `AnnotatedMethod.remove(Collection<?>)` or intersection?

manovotn:
Hi John,
here are some answers to the first part of your question:
> why is there no add scope?

There is a method named `scope`. This name was chosen because you are not adding it but rather replacing the current one.

>  A bean should only have one type? Or many types? Wouldn't it be inferred from the registered type?

If you use `read()` method you can read those types from BeanAttributes (or annotated type which will likely be removed). Otherwise there is `addType`, `addTypes`, `types` (replaces them) and finally `addTypeClosure`. So you can add multiple at once, replace current types or you can just use a closure type.

> but how do i or CDI know that you're done building the object?

You don't need to call any method to specifically invoke creation; using `AfterBeanDiscovery.addBean()` gives you a single instance of Configurator (non-reusable). So you can ofc build multiple beans within one observer - just call addBean several times. Such approach can be seen [here] (https://github.com/cdi-spec/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/bean/AfterBeanDiscoveryObserver.java#L37). I presume the bean creation is done once you exit the observer method.

I have to take a closer look at the second part of your question. I will add a comment once I do so :)

mkouba:
@johnament @manovotn  WRT `ProcessAnnotatedType` - actually this wouldn't work right now (see also https://issues.jboss.org/browse/CDI-596). Anyway, I did not study the `CdiCamelExtension` source yet, but something like this should work:
````
private void camelFactoryProducers(@Observes ProcessAnnotatedType<CdiCamelFactory> pat, BeanManager manager) {
    pat
        .configureAnnotatedType()
        .filterMethods(CdiCamelExtension::isValidMethod)
        .forEach(amc -> {
            for (Annotation qualifier : CdiSpiHelper.getQualifiers(am, manager)) {
                amc.remove(qualifier);
            }
            amc.add(Excluded.INSTANCE);
        });
}
````
Note that once you're configuring the AT, all the changes are propagated automatically.

PR #289, opened 2016-05-04
Cdi 457

johnament:


johnament:
I'm undecided if this should change or if section 5.6.1 should be updated.  

I'm also not sure which branch to target.

mkouba:
@johnament See also https://github.com/cdi-spec/cdi/pull/286 - it's related.

mkouba:
By the way, I think it is defined even now:

> 6.4.1. Dependent objects:
> An instance of a bean with scope @Dependent obtained by direct invocation of an Instance is a dependent object of the instance of Instance.

> 5.6.1. The Instance interface:
> The method destroy() instructs the container to destroy the instance. The bean instance passed to destroy() should be a dependent scoped bean instance, or a client proxy for a normal scoped bean. Applications are encouraged to always call destroy() when they no longer require an instance obtained from Instance.

The only questionable part is this {{javax.enterprise.inject.spi.CDI}} whose scope is not well-defined.

johnament:
While it may be defined, its not clear how or why you would use the methods currently which is really my aim here.  If you recall, this originally came up as a way to manage dependent objects that had no bounds.  

It seems like we may be looking at the same problem, so lets bring it up on the mailing list.  I'll close this out for now.

PR #291, opened 2016-05-15
CDI-30 Added a controller and interceptor to support starting and stopping contexts.

johnament:
I would consider this to be a first pass, and requires more discussion, but wanted to get the discussion points started before making spec changes.

### DIFF
--- a/api/src/main/java/javax/enterprise/context/spi/ContextController.java
+++ b/api/src/main/java/javax/enterprise/context/spi/ContextController.java
@@ -1,0 +1,75 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package javax.enterprise.context.spi;
+
+import javax.enterprise.context.ContextNotActiveException;
+import java.lang.annotation.Annotation;
+import java.util.Optional;
+
+/**
+ * A programmatic interface for retrieving, starting and stopping built in contexts
+ *
+ * The container provides a dependent scoped instance of this bean which may be injected to manage contexts
+ *
+ * <pre>
+ * &#064;Inject
+ * private ContextController contextController;
+ *
+ *
+ * ManagedContext context = contextController.getManagedContext(RequestScoped.class)
+ *                  .orElseGet(() -> contextController.activate(RequestScoped.class));
+ *
+ * // do some work.
+ *
+ * context.deactivate();
+ *
+ * </pre>
+ *
+ * @author John D. Ament
+ */
+public interface ContextController {
+
+    /**
+     * Obtains an active {@linkplain ManagedContext context object} for the given
+     * {@linkplain javax.enterprise.context scope} based on what is active within the current thread.
+     *
+     * @param scopeType the {@linkplain javax.enterprise.context scope}
+     * @return an Optional representation of the {@linkplain ManagedContext context object}
+     *         or else Optional.empty
+     * @throws IllegalArgumentException if there is more than one active context object for the given scope
+     */
+    Optional<ManagedContext> getManagedContext(Class<? extends Annotation> scopeType);
+
+    /**
+     * Obtains an active {@linkplain ManagedContext context object} for the given
+     * {@linkplain javax.enterprise.context scope} based on its id.
+     * @param id the id of the context to be looked up
+     * @param scopeType the {@linkplain javax.enterprise.context scope}
+     * @return the {@linkplain ManagedContext context object}
+     * @throws ContextNotActiveException if there is no active context object for the given scope with this ID
+     */
+    ManagedContext getManagedContext(String id, Class<? extends Annotation> scopeType);
+
+    /**
+     * Activates the provided scope type, creating a new Context object for the scope type.
+     *
+     * The activated context is immediately associated to the current thread.
+     */
+    ManagedContext activate(Class<? extends Annotation> scopeType) throws IllegalStateException;
+
+}
--- a/api/src/main/java/javax/enterprise/context/spi/ManagedContext.java
+++ b/api/src/main/java/javax/enterprise/context/spi/ManagedContext.java
@@ -1,0 +1,80 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2013, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package javax.enterprise.context.spi;
+
+import javax.enterprise.context.ContextNotActiveException;
+import javax.enterprise.inject.Instance;
+import java.lang.annotation.Annotation;
+
+/**
+ * Representation of a context which has been activated
+ *
+ * @author John D. Ament
+ */
+public interface ManagedContext {
+
+    /**
+     * Provides access to an internal ID used to uniquely identify this active context
+     *
+     * @return the internal ID of this context, if it is supported
+     */
+    String getId();
+
+    /**
+     * Associates the context to the given thread.
+     * @param thread the thread to associate with.  This does not remove associations with other threads, but acts additive
+     * @throws IllegalArgumentException if the given thread is already associated with this context.
+     */
+    void associate(Thread thread);
+
+    /**
+     * Disassociates
+     * @param thread
+     */
+    void disassociate(Thread thread);
+
+    /**
+     * Get the scope type of the context object.
+     *
+     * @return the scope
+     */
+    Class<? extends Annotation> getScope();
+
+    /**
+     * Determines if the context object is active.
+     *
+     * @return <tt>true</tt> if the context is active, or <tt>false</tt> otherwise.
+     */
+    boolean isActive();
+
+    /**
+     * Provides an Instance object that is bound to this context.  This instance may be used to retrieve or create
+     * beans within the scope of this context.
+     *
+     * @return an Instance object to retrieve beans from this context.
+     */
+    Instance<Object> instance();
+
+    /**
+     * Deactivates the given managed context.  Can only be used to deactivate contexts that were activated by a context controller
+     *
+     * @throws ContextNotActiveException if the given context has already been deactivated
+     */
+    void deactivate();
+
+}
--- a/spec/src/main/asciidoc/core/scopescontexts.asciidoc
+++ b/spec/src/main/asciidoc/core/scopescontexts.asciidoc
@@ -506,3 +506,66 @@ If `end()` is called, and the current conversation is marked transient, an `Ille
 If `begin()` is called, and the current conversation is already marked long-running, an `IllegalStateException` is thrown.
 
 If `begin()` is called with an explicit conversation identifier, and a long-running conversation with that identifier already exists, an `IllegalArgumentException` is thrown.
+
+=== Context control for built-in scopes
+
+Most of the time the container will handle the lifecycle of built-in contexts automatically. However, in some cases it may be vital to have the ability to start and stop a context manually.
+To support these cases, the container provides built in beans that can handle the operations for activation and deactivation.
+
+[source, java]
+----
+public interface ContextController {
+    Optional<ManagedContext> getManagedContext(Class<? extends Annotation> scopeType);
+    ManagedContext getManagedContext(String id, Class<? extends Annotation> scopeType);
+    ManagedContext activate(Class<? extends Annotation> scopeType) throws IllegalStateException;
+}
+----
+
+A built in bean exists of type `ContextController`  which may be used to manage contexts.  This built in bean has scope `Dependent` and the qualifier `Default`.
+
+* `getManagedContext()` allows you to look up a context.  When only passing in the scopeType, the current thread will be used to look up an `ManagedContext`.
+* `activate()` starts the provided context on the current thread, allowing beans of that scope to be used.  May throw `IllegalStateException` if the context is already active on this thread.  You should check if the context is active before calling.
+* `deactivate()` stops the given context if it was started by this bean.  Throws a `ContextNotActiveException` if the given `ManagedContext` is no longer active, or was not activated by a `ContextController`
+
+==== Contexts supporting activation
+
+The following built in contexts may be activated manually
+
+* `ConversationScoped`
+* `RequestScoped`
+* `SessionScoped`
+
+`ApplicationScoped` may not be activated as it is always guaranteed to be active while CDI is running.
+
+==== Active Contexts
+
+Once you have retrieved (or activated) a context, you may use the `ManagedContext` reference to interact with beans defined within that context
+
+[source, java]
+----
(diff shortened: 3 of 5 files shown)


mkouba:
I like the idea of interceptors, which is what Weld SE does for `@RequestScoped`: http://docs.jboss.org/weld/reference/latest/en-US/html/environments.html#_java_se

However, I would not use a "generic" interceptor binding but rather a single binding per scope, e.g. `@ActivateRequestScope`.

Also does it really make sense for all built-in scopes?

johnament:
@mkouba It doesn't if we clarify that app scoped is always on, then it will only make sense for session/conversation/request scopes.  I'm fine with using specific ones, and probably even the programmatic API should be more strongly typed, what about:

    Context c = contextController.activateRequestContext();

Also, why not `@ActivateRequestContext` ? My interpretation is that the scope is the boundaries of the bean, whereas context is what is active at some point.

johnament:
@mkouba I've made the changes to the interceptor... let me know what you think.

mkouba:
+1 for interceptor bindings. I'm not so sure about the `ContextController`. I understand the motivation but I think we should be careful here. I like the idea of "strongly typed API" you've mentioned in https://github.com/cdi-spec/cdi/pull/291#issuecomment-219405823. I would also use `activate/deactivate` instead of `start/stop` as the spec only speaks about built-in contexts being `active` and there's only `Context.isActive()`. `ContextActivator` might be also a better name for this new built-in bean.

johnament:
I think this new version is much more type safe.  Let me know.

mkouba:
Looks good, except:
* `deactivate()` should not do anything (or throw exception) if the scope was not previously activated with this specific "activator" instance
* I wouldn't call it `ActivatableContext` because it's not a context anyway, it's just a thing which activates/deactivates a context for a built-in scope (e.g. provides a request context object for the built-in request scope)
* also I would explicitly list the built-in scopes supported (for example `@Dependent` is also a built-in scope but it does not make sense to provide an "activator" bean for it)

johnament:
I'm expecting things like the built in scopes supported to be in the spec once those changes go in.  I agree with making deactivate only work if it was activated here, however I was also expecting that the weld impl may look like

```
public interface BoundRequestContext extends RequestContext, BoundContext<Map<String, Object>>, ActivatableContext<RequestScope>
```

But that's just my thoughts on how you may implement.

mkouba:
> I'm expecting things like the built in scopes supported to be in the spec once those changes go in.

Ok, I see. You only want to discuss API ri

> however I was also expecting...

Yes, but it's really an implementation detail, ie. it shouldn't be taken into account right now.

api/src/main/java/javax/enterprise/context/spi/ActiveContext.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    boolean isActive();
+
+    /**
+     * Provides an Instance object that is bound to this context.  This instance may be used to retrieve or create
+     * beans within the scope of this context.
+     *
+     * @return an Instance object to retrieve beans from this context.
+     */
+    Instance<Object> instance();
```
mkouba: `Instance` is never bound to any context. I don't undertand the purpose of this method...

api/src/main/java/javax/enterprise/context/spi/ActiveContext.java
```diff
@@ -0,0 +1,72 @@
...
+
+import javax.enterprise.inject.Instance;
+import java.lang.annotation.Annotation;
+
+/**
+ * Representation of a context which has been activated
+ *
+ * @author John D. Ament
+ */
+public interface ActiveContext {
```
mkouba: `ActiveContext` -> `ActivatedContext`?

api/src/main/java/javax/enterprise/context/spi/ContextController.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Obtains an active {@linkplain javax.enterprise.context.spi.ActiveContext context object} for the given
+     * {@linkplain javax.enterprise.context scope} based on its id.
+     * @param id the id of the context to be looked up
+     * @param scopeType the {@linkplain javax.enterprise.context scope}
+     * @return the {@linkplain javax.enterprise.context.spi.ActiveContext context object}
+     * @throws ContextNotActiveException if there is no active context object for the given scope with this ID
+     */
+    ActiveContext getActiveContext(String id, Class<? extends Annotation> scopeType);
```
mkouba: I don't understand the concept of this context `id`. A simple example might help.

api/src/main/java/javax/enterprise/context/spi/ActiveContext.java
johnament: Its to provide some way of giving a handle to allow a user to do programmatic lookup.  I agree that the injected instance isn't bound to a context, but the instance returned by this method would be.

I'm open to other ideas though, as long as we can allow programmatic lookup.

api/src/main/java/javax/enterprise/context/spi/ActiveContext.java
johnament: Nah, past tense doesn't make sense.

api/src/main/java/javax/enterprise/context/spi/ContextController.java
johnament: The two concrete ones we spoke about in IRC were for conversation and session.  Conversations have an ID, which is passed in as a URL parameter currently to look up the current conversation for the request.

Similarly, HttpSessions are bound to an ID, and that ID indicates what session scope to activate.

api/src/main/java/javax/enterprise/context/spi/ActiveContext.java
tremes: Yes but `ActiveContext.isActive` looks little bit weird IMO.

mkouba: Well, `ActiveContext.isActive()` does not make sense either.

api/src/main/java/javax/enterprise/context/spi/ActiveContext.java
mkouba: So it should be a replacement for `Context.get(Contextual<T>)`?

johnament: Functionally, yes.  Note that a context != an active context, active context is specifically not thread bound.  It also doesn't make sense for it to take a contextual.

api/src/main/java/javax/enterprise/context/spi/ActiveContext.java
johnament: I can agree with @tremes that it looks a little weird, but what doesn't make sense about it? Using past tense makes it look like it was already active.

api/src/main/java/javax/enterprise/context/spi/ActiveContext.java
```diff
@@ -0,0 +1,72 @@
...
+     * @return the internal ID of this context, if it is supported
+     */
+    String getId();
+
+    /**
+     * Associates the context to the given thread.
+     * @param thread the thread to associate with.  This does not remove associations with other threads, but acts additive
+     * @throws IllegalArgumentException if the given thread is already associated with this context.
+     */
+    void associate(Thread thread);
```
tremes: Ok so this basically allows user to spawn multiple requestScope, sessionScope etc. right? I am not sure I like it but could be helpful if you can introduce some basic example (code snippet) of usage e.g via gist. 

Or maybe this comment should be linked to `ActiveContext activate(Class<? extends Annotation> scopeType)`?? What happens if I activate a context but don't associate it with any thread? I am not sure I understand to this "lifecycle" of context.

spec/src/main/asciidoc/core/scopescontexts.asciidoc
```diff
@@ -506,3 +506,65 @@ If `end()` is called, and the current conversation is marked transient, an `Ille
 If `begin()` is called, and the current conversation is already marked long-running, an `IllegalStateException` is thrown.
 
 If `begin()` is called with an explicit conversation identifier, and a long-running conversation with that identifier already exists, an `IllegalArgumentException` is thrown.
+
+=== Context control for built-in scopes
+
+The container provides support for automatically starting contexts based on occurrences within the container.  There are cases where manually starting and stopping a context
```
mkouba: What does `contexts based on occurrences` actually mean?

api/src/main/java/javax/enterprise/context/spi/ContextController.java
```diff
@@ -0,0 +1,82 @@
...
+     * The activated context is immediately associated to the current thread.
+     */
+    ActiveContext activate(Class<? extends Annotation> scopeType) throws IllegalStateException;
+
+    /**
+     * Deactivates the given active context.  Can only be used to deactivate contexts that were activated by a context controller
+     *
+     * @throws ContextNotActiveException if the given context has already been deactivated
+     */
+    void deactivate(ActiveContext activeContext) throws ContextNotActiveException;
```
mkouba: Wouldn't be better to move this method to `ActiveContext`?

api/src/main/java/javax/enterprise/context/spi/ActiveContext.java
tremes: Ahh it's started in the current thread. Ok. So ignore my second paragraph in above comment.

tremes:
How can I stop/deactivate e.g sessionScope which is associated to some (not current) thread? By calling disassociate? It's not clear to me from this proposal.

spec/src/main/asciidoc/core/scopescontexts.asciidoc
```diff
@@ -506,3 +506,65 @@ If `end()` is called, and the current conversation is marked transient, an `Ille
...
+Once you have retrieved (or activated) a context, you may use the `ActiveContext` reference to interact with beans defined within that context
+
+[source, java]
+----
+public interface ActiveContext {
+    String getId();
+    void associate(Thread thread);
+    void disassociate(Thread thread);
+    Class<? extends Annotation> getScope();
+    boolean isActive();
```
Emily-Jiang: ActiveContext.isActive() is odd. It should return true all the time.

api/src/main/java/javax/enterprise/context/spi/ContextController.java
mkouba: Neither of these "activated" contexts will be backed by an HTTP session so I don't think the ID makes any sense. It would only make sense if you allow to associate the context with the different storage (e.g. user-provided Map). See also [BoundConversationContext example](http://docs.jboss.org/weld/reference/latest/en-US/html/contexts.html#_managing_the_built_in_contexts). But this would complicate the API and I believe it's not worth including. I think we should either support a kind of **transient** conversations and sessions or don't support activation of session and conversation context at all.

api/src/main/java/javax/enterprise/context/spi/ContextController.java
johnament: I'm ok with that move.

spec/src/main/asciidoc/core/scopescontexts.asciidoc
johnament: Do you have any recommendation on how to phrase this better?  I'm basically referring to the fact that an incoming HTTP request starts a Request Context.

api/src/main/java/javax/enterprise/context/spi/ActiveContext.java
johnament: @tremes No, its not started in the current thread.  One of the ideas here is to dis-associate the threading model and the context model.  We can still map back to the threading model via this method, but ideally the context that gets created is managed outside of any thread, and injection points are satisfied from the owning context.

api/src/main/java/javax/enterprise/context/spi/ActiveContext.java
johnament: Any more suggestions on better names?

spec/src/main/asciidoc/core/scopescontexts.asciidoc
mkouba: I'm not sure, maybe something like _"The container usually activates and destroys the built-in contexts automatically. However, there are cases where manually starting..."_

manovotn: How about this?
_Most of the time the container will handle the lifecycle of built-in contexts automatically. However, in some cases it may be vital to have the ability to start and stop a context manually. To support ..._

api/src/main/java/javax/enterprise/context/spi/ActiveContext.java
manovotn: I'll take a shot in the dark - _ManagedContext_ ? Since you can actually manage its lifecycle yourself (or the interceptor can manage it for you).
Another one could be _ActivableContext_. (granted, _activable_ is a bit a of a neologism :-) )
Since it does not really need to be "active" at all times I would try to refrain from using word "active" in its name. Which would instantly make a call to `isActive()` sound meaningful.

johnament: I like it.

api/src/main/java/javax/enterprise/context/spi/ContextController.java
```diff
@@ -0,0 +1,75 @@
...
+ * A programmatic interface for retrieving, starting and stopping built in contexts
+ *
+ * The container provides a dependent scoped instance of this bean which may be injected to manage contexts
+ *
+ * <pre>
+ * &#064;Inject
+ * private ContextController contextController;
+ *
+ *
+ * ActiveContext context = contextController.getActiveContext(RequestScoped.class)
```
manovotn: `ActiveContext` -> `ManagedContext`

spec/src/main/asciidoc/core/scopescontexts.asciidoc
```diff
@@ -506,3 +506,66 @@ If `end()` is called, and the current conversation is marked transient, an `Ille
...
+public interface ContextController {
+    Optional<ManagedContext> getManagedContext(Class<? extends Annotation> scopeType);
+    ManagedContext getManagedContext(String id, Class<? extends Annotation> scopeType);
+    ManagedContext activate(Class<? extends Annotation> scopeType) throws IllegalStateException;
+}
+----
+
+A built in bean exists of type `ContextController`  which may be used to manage contexts.  This built in bean has scope `Dependent` and the qualifier `Default`.
+
+* `getActiveContext()` allows you to look up a context.  When only passing in the scopeType, the current thread will be used to look up an `ActiveContext`.
```
manovotn: `getActiveContext` -> `getManagedContext`
`ActiveContext` -> `ManagedContext`

johnament: Should be all good now.

api/src/main/java/javax/enterprise/context/spi/ManagedContext.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Associates the context to the given thread.
+     * @param thread the thread to associate with.  This does not remove associations with other threads, but acts additive
+     * @throws IllegalArgumentException if the given thread is already associated with this context.
+     */
+    void associate(Thread thread);
+
+    /**
+     * Disassociates
```
Emily-Jiang: unfinished javadoc

Emily-Jiang: do we need to add a method getAssociatedThread()

api/src/main/java/javax/enterprise/context/spi/ManagedContext.java
```diff
@@ -0,0 +1,80 @@
...
+     * @return the internal ID of this context, if it is supported
+     */
+    String getId();
+
+    /**
+     * Associates the context to the given thread.
+     * @param thread the thread to associate with.  This does not remove associations with other threads, but acts additive
+     * @throws IllegalArgumentException if the given thread is already associated with this context.
+     */
+    void associate(Thread thread);
```
antoinesd: I'm not a big fan of explicitly add a dependency to `Thread` in our API.
When re-reading original JSR we wrote on point 3:

> The context model in CDI is based around the thread-bound request-response model. This model is more prevalent in web applications than elsewhere, and, also is less pervasive than it once was. We would like to overhaul the context model to allow the application or container to portably push the active context to CDI when it requires the use of CDI. This change would be largely transparent to applications, and we might want to consider an API facade over it

To be honest, I'm not totally sure of what Pete meant we that, but my feeling is that if we add something context related we should make sure that it could be used in a more modern reactive world.

I'm just wondering if we are not closing doors here.

api/src/main/java/javax/enterprise/context/spi/ManagedContext.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.enterprise.context.ContextNotActiveException;
+import javax.enterprise.inject.Instance;
+import java.lang.annotation.Annotation;
+
+/**
+ * Representation of a context which has been activated
+ *
+ * @author John D. Ament
+ */
+public interface ManagedContext {
```
antoinesd: This interface has multiple method that exist in `Context` (and thus `AlterableContext`). I there a reason to not extend `AlterableContext` here ?

johnament:
Closing this PR.  Will create a new one that's clearer.

PR #294, opened 2016-05-26
CDI-603 Create separate doc section for each Annotated*Configurator

antoinesd:
CDI-603 Create separate doc section for each Annotated*Configurator

### DIFF
--- a/spec/src/main/asciidoc/core/spi.asciidoc
+++ b/spec/src/main/asciidoc/core/spi.asciidoc
@@ -846,7 +846,7 @@ public interface AnnotatedTypeConfigurator<T> {
 
 ===== `AnnotatedMethodConfigurator`
 
-`AnnotatedMethodConfigurator` is obtained through `AnnotatedTypeConfigurator` as defined in <<AnnotatedTypeConfigurator>>
+`AnnotatedMethodConfigurator` is obtained through `AnnotatedTypeConfigurator` as defined in <<annotated_type_configurator>>
 
 [source, java]
 ----
@@ -873,7 +873,7 @@ public interface AnnotatedMethodConfigurator<T> {
 
 ===== `AnnotatedConstructorConfigurator`
 
-`AnnotatedConstructorConfigurator` is obtained through `AnnotatedTypeConfigurator` as defined in <<AnnotatedTypeConfigurator>>
+`AnnotatedConstructorConfigurator` is obtained through `AnnotatedTypeConfigurator` as defined in <<annotated_type_configurator>>
 
 [source, java]
 ----
@@ -921,7 +921,7 @@ public interface AnnotatedParameterConfigurator<T> {
 
 ===== `AnnotatedFieldConfigurator`
 
-`AnnotatedFieldConfigurator` is obtained through `AnnotatedTypeConfigurator` as defined in <<AnnotatedTypeConfigurator>>:
+`AnnotatedFieldConfigurator` is obtained through `AnnotatedTypeConfigurator` as defined in <<annotated_type_configurator>>:
 
 [source, java]
 ----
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -798,13 +798,21 @@ CDI 2.0 introduced a new SPI to help defining and creating instance for type met
...
-* `javax.enterprise.inject.spi.builder.AnnotatedMethodConfigurator` to configure an `AnnotatedMethod`
-* `javax.enterprise.inject.spi.builder.AnnotatedParameterConfigurator` to configure an `AnnotatedParameter`
+* `javax.enterprise.inject.spi.builder.AnnotatedFieldConfigurator` (defined in <<annotated_field_configurator>>) to configure an `AnnotatedField`
+* `javax.enterprise.inject.spi.builder.AnnotatedConstructorConfigurator` (defined in <<annotated_constructor_configurator>>) to configure an `AnnotatedConstructor`
+* `javax.enterprise.inject.spi.builder.AnnotatedMethodConfigurator` (defined in <<annotated_method_configurator>>) to configure an `AnnotatedMethod`
+* `javax.enterprise.inject.spi.builder.AnnotatedParameterConfigurator` (defined in <<annotated_parameter_configurator>>) to configure an `AnnotatedParameter`
 
+The container must provide an implementation for each of these interfaces.
+
+* `AfterTypeDiscovery` as defined in <<after_type_discovery>>
```
tremes: This list item should be moved below right? Or I don't get it.

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -835,7 +843,11 @@ public interface AnnotatedTypeConfigurator<T> {
 * `constructors()` returns a set of `AnnotatedConstructorConfigurator` to configure annotations on constructors.
 * `filterConstructors()` returns a `Stream<AnnotatedConstructorConfigurator>` filtered by applying the provided Predicate on `Constructors()`
 
-`AnnotatedMethodConfigurator` and `AnnotatedConstructorConfigurator` are accessed through `AnnotatedTypeConfigurator`:
+[[annotated_method_configurator]]
+
+==== `AnnotatedMethodConfigurator`
```
manovotn: Why is the first section named `AnnotatedTypeConfigurator SPI`, while the sections below (method, constructor,...)  lack the **SPI** keyword.
It really seems weird since they are all on the same level (all are equivalent subsection of `Alternative metadata sources`), I suppose you wanted to make it obvious that `AnnotatedTypeConfigurator` is the entry point for the whole thing? But that is already said in the text. I suggest we make the Annotated*Configurator chapters a subsection of `AnnotatedTypeConfigurator SPI`. That would IMO underline what is the entry point and also shed some light on how does the hierarchy work.

E.g. using current section numbers it would look like this:
11.4.1 `AnnotatedTypeConfigurator SPI`
 11.4.1.1 `AnnotatedMethodConfigurator`
 11.4.1.2 `AnnotatedConstructorConfigurator`
 11.4.1.3 `AnnotatedParameterConfigurator`
 11.4.1.4 `AnnotatedFieldConfigurator`
11.5  `Container lifecycle events`
...

WDYT?

antoinesd: I hesitated to go that way (to avoid to deep level title). But as we already have a 4 deep level title in  2.8.1, it makes sense. I modify the PR accordingly.

spec/src/main/asciidoc/core/spi.asciidoc
antoinesd: Right, bad cut & paste.

tremes:
Links to `AnnotatedTypeConfigurator` don't work for me in HTML and also look different than other links (but this is the problem of entire document). Otherwise looks good.

PR #295, opened 2016-06-01
CDI-611 Enhance 10.3.3. Multiple event qualifiers section.

manovotn:


spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -296,15 +296,18 @@ An event parameter may have multiple qualifiers.
 public void afterDocumentUpdatedByAdmin(@Observes @Updated @ByAdmin Document doc) { ... }
 ----
 
-Then this observer method will only be notified if all the observed event qualifiers are specified when the event is fired:
+Then this observer method will be notified if the set of observer qualifiers is a subset of the fired event's qualifiers:
```
antoinesd: Including empty set.

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -296,15 +296,18 @@ An event parameter may have multiple qualifiers.
...
 [source, java]
 ----
 @Inject Event<Document> documentEvent;
 ...
-documentEvent.select(new UpdatedQualifier(), new ByAdminQualifier()).fire(document);
+documentEvent.select(new UpdatedQualifier(), new ByAdminQualifier(), new ClarificationQualifier()).fire(document);
 ----
 
+In the above example the event is fired with `@ByAdmin`, `@Updated` and `@Clarification` qualifiers while the observer qualifiers are `@Updated` and `@ByAdmin`.
+Observer qualifiers therefore form a subset of event qualifiers and the observer will be notified.
```
antoinesd: same here, we should explicitly said that an observer without qualifier will also be notified

manovotn: @antoinesd well ... empty set is a subset of all other sets and itself . The question is, whether Average Joe knows this :)
Anyway, the example just below this text shows that
`public void afterDocumentEvent(@Observes Document doc) { ... }` will also be notified, which is exactly the case you are reffering to.

johnament: Is it worth pointing out the `@Any` qualifier caveats here?

manovotn: @johnament could you elaborate? I don't get you.

antoinesd: I think @johnament mention that `@Any` is useless regarding event. You don't need to use it for injection if you want to select an event of a given qualifier later.
Regarding empty set I agree, the problem is that it's working differently than typesafe resolution but look very similar to programmatic lookup. We should find a way to clarify this ambiguity.

manovotn:
@antoinesd I slightly changed the PR and included a notion of `@Any`. CDI spec already clearly states that `@Any` is by default added to every event but it won't hurt to mention it probably. 
What I am not so sure about is that every observer also has `@Any` by default. It makes sense (e.g. explains why observer with no qualifier gets notified for all events of given type) but I couldn't find it anywhere in the spec, so I mentioned it here.

@johnament does that somewhat clears up the caveat?

Feel free to object :)

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -296,22 +296,27 @@ An event parameter may have multiple qualifiers.
...
 [source, java]
 ----
 @Inject Event<Document> documentEvent;
 ...
-documentEvent.select(new UpdatedQualifier(), new ByAdminQualifier()).fire(document);
+documentEvent.select(new UpdatedQualifier(), new ByAdminQualifier(), new ClarificationQualifier()).fire(document);
 ----
 
-Other, less specific, observers will also be notified of this event:
+In the above example the event is fired with `@ByAdmin`, `@Updated`, `@Clarification` and `@Any` qualifiers. `@Any` is added automatically as described in <<event_types_and_qualifier_types>>. The observer qualifiers are `@Updated`, `@ByAdmin` and again `@Any`.
```
manovotn: Just to clarify - I am aware that in practice, the event has `@Default` qualifier as well (derived from section 2.3.1 Built-in qualifier types in spec), but I would rather avoid going into that here as it would probably introduce more confusion.

antoinesd: In fact the @Any mention brings more confusion IMO than @Default mention.
From user perspective @Any is useless regarding observer resolution since firing a qualified event will triggers an unqualified observer (if it has an eligible event type). Adding @Any on one or both side doesn't change this behavior.
OTHO @Default has an effect. If I add @Default to an observer it will be resolved only for events having no qualifier (this use case is not given as an example in the spec).

To make short. Even if @Any is added on Event and Observer (BTW the spec doesn't say anything about this last one) it's totally neutral from user POV, so IMO we should avoid example with @Any to limit user confusion.
Adding an example with @Default on an observer would useful because a lot of user (including myself) wonder how to declare an observer resolved only if the fired event has no qualifier.

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -296,22 +296,27 @@ An event parameter may have multiple qualifiers.
...
+Observer qualifiers therefore form a subset of event qualifiers and the observer will be notified.
+
+Other, less specific, observer will also be notified of this event:
 
 [source, java]
 ----
 public void afterDocumentUpdated(@Observes @Updated Document doc) { ... }
 ----
 
+Last but not least, the following observer is the most general one and since it does not declare any qualifier, it has `@Any` by default.
```
antoinesd: @manovotn There is no @Any added on observer side. IMO, the fact this observer gets notified come from section [10.3 Observer resolution](http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#observer_resolution)

> An event is delivered to an observer method if:
> 
> * The observer method belongs to an enabled bean.
> * An event type is assignable to the observed event type, taking type parameters into consideration.
> * **The observer method has no event qualifiers** or **has a subset of the event qualifiers.**
> An observer method has an event qualifier if it has an observed event qualifier with (a) the same type and (b) the same annotation member value for each member which is not annotated `@javax.enterprise.util.Nonbinding`.
>...

I agree that empty set is a subset of all sets but "Average Joe" needed this Captain Obvious clarification to make the difference with [Typesafe Resolution](http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#performing_typesafe_resolution) for bean injection where there is never an empty set of qualifier (empty set is replaced by `@Default`):

> A bean is _assignable_ to a given injection point if:
> 
> * ...
> * The bean has all the required qualifiers. **If no required qualifiers were explicitly specified, the container assumes the required qualifier `@Default`**. A bean has a required qualifier if it has a qualifier with (a) the same type and (b) the same annotation member value for each member which is not annotated `@javax.enterprise.util.Nonbinding`.

manovotn:
@antoinesd I reflected the changed you required:
1) explicitly mentioned empty set
2) Removed `@Any` from samples
3) Added sample with `@Default`

WDYT about it now?

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -317,6 +320,24 @@ public void afterDocumentUpdated(@Observes @Updated Document doc) { ... }
...
 ----
 
+On the other hand, following observer will not be notified as slightly different behaviour applies to observers with `@Default` qualifier:
+
+[source, java]
+----
+public void afterDocumentDefaultEvent(@Observes @Default Document doc) { ... }
+----
+
+Such observer will only be notified for events having either no qualifiers or only `@Default` qualifier:
```
johnament: Is this backwards compatible?

manovotn: This is nothing new. It is how it works ATM.
From specification, [section 2.3.1](http://docs.jboss.org/cdi/spec/2.0.EDR1/cdi-spec.html#builtin_qualifiers): _The default qualifier is also assumed for any injection point that does not explicitly declare a qualifier, as defined in The default qualifier at injection points._

This then points to [section 3.8](http://docs.jboss.org/cdi/spec/2.0.EDR1/cdi-spec.html#injection_point_default_qualifier) where there is an example:

> Finally, the following are equivalent:
> `@Inject Order order;`
> `@Inject @Default Order order;`

antoinesd:
+1 for me @manovotn

tremes:
+1

PR #296, opened 2016-06-29
Cdi 30 beanmanager

johnament:


api/src/main/java/javax/enterprise/context/spi/BaseContext.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * <p>
+     * Destroy the existing contextual instance. If there is no existing instance, no action is taken.
+     * </p>
+     *
+     * @param contextual the contextual type
+     * @throws ContextNotActiveException if the context is not active
+     */
+    default void destroy(Contextual<?> contextual) {
```
johnament: Please note - I moved this to avoid duplicating the method in the next unamanaged context class.

mkouba: I understand the reason but this breaks the backward compatibility, i.e. all contexts which implement  `javax.enterprise.context.spi.Context` but not `AlterableContext`. The truth is extensions-provided contexts **_should_** implement `AlterableContext` instead of `Context`... still, they don't have to.

manovotn: Unfortunately there is an issue with this -> backward compatibility problem.
I ran Java compliance api checker against this solution and current CDI 1.2 and this causes a source incompatibility (binary is fine). If users have their own context implementation, this will cause failures. I think we might need to look for another way?

johnament: Can you give me an example to try? I might be missing something.  What compliance checker did you use and is it aware of default methods?

mkouba:
I think the concept is fine. But the fact that `UnmanagedContext` does not extend `Context` is imho confusing. If the `AfterBeanDiscovery.addContext()` is the only reason to refactor the hierarchy then **-1** from me. Extension developers should be more advanced users who know what they're doing.

api/src/main/java/javax/enterprise/context/spi/BaseContext.java
mkouba: Hm, I've missed the method is declared as default. In this case, the compatibility is ok. But on the other hand, alterable context author will not be forced to implement this method. Not good.

manovotn: Already found out the checker was indeed not aware of default methods and I created an issue there :-/
For the record, here is what I used -> https://github.com/lvc/japi-compliance-checker/

mkouba:
By the way, it's a pity we cannot simply extend `javax.enterprise.inject.spi.Unmanaged` as it's too "non-contextual instance centric".

api/src/main/java/javax/enterprise/context/spi/BaseContext.java
johnament: Ah ok, thanks for double checking.  I agree making it not required is an issue.  I'm not wholely against duplicating the method in alterable context and unmanagedcontext if it makes more sense.  This was just an idea.

johnament:
@mkouba I don't believe that the extension issue is the only problem.  We can fix the extension issue by saying an exception is thrown if you try to register an unmanaged context.  It does create some syntactic ugliness even for app developers.

Where I see an issue is that the two objects mean two different things.  A Context is something that's already running.  An UnmaangedContext is a representation of one of those Contexts, which can be started manually.  BaseContext reflects that better, indicating a more composite based approach, rather than direct inheritance from unrelated object types.

If you think about it from a polymorphism point of view, multiple implementations are meant to be substitutable.  That isn't the case here.  UnmanagedContext shouldn't be treated as a drop in replacement for a Context object.

mkouba:
@johnament If you think these are two different things then the name `UnmanagedContext` is just wrong. Also I believe we should stick with the spec, where a context is **NOT** _"something that's already running"_. It may be **active** with respect to the current thread. It has a **scope**. And it allows to **obtain contextual instances**. That's all. And there are some built-in contexts which are activated/deactivated automatically. So the only difference is that this "unmanaged buil-in context" is activated manually. I don't see a problem here.

api/src/main/java/javax/enterprise/context/spi/BaseContext.java
antoinesd: Less elegant but probably safer to duplicate the code here as we said during last meeting...

johnament:
@antoinesd all methods duplicated now.

antoinesd:
@tremes or @manovotn, since @mkouba is not available, do have detail on his comment?

tremes:
I don't know the exact Martin's opinion on this (didn't talk to him for last two weeks). Maybe he tried to imply that those two things (unmanaged context vs. common ones) are not really different. I would agree with him. Currently I cannot see any reason why `UnmanagedContext` cannot extend `Context`. Can somebody elaborate?
I also think it would be helpful to enrich javadoc for these new methods `activate` and `deactivate`. 
Matej is on PTO this week AFAIK.

johnament:
These were the decisions taken during the last meeting: http://transcripts.jboss.org/meeting/irc.freenode.org/cdi-dev/2016/cdi-dev.2016-06-28-16.04.txt

This PR was to reflect those decisions.

I agree that the JavaDoc needs some beefing up, but didn't see that mentioned in the PR.  We talked about why UnmanagedContext wasn't a Context as defined today as well, with two key points:

- You can register a Context at runtime.  You shouldn't be able to register an UnmanagedContext
- It is confusing to users, even advance users, why this Context is also an UnmanagedContext.  Other than avoiding copying methods, it didn't save anything clarifying.  If you look at most specifications, extension of a base is meant to give it more capabilities.  UnmanagedContext is for a completely different use-case.

Therefore it made sense to introduce the base class.

tremes:
Ok. I don't see these two points as a problem. I think it's ok to simply throw an exception when you try to register `UnmanagedContext` at runtime (as you said). When you are reigstering your custom context impl then I think you are aware of what you are doing. 
But it's true that if we introduce `UnmanagedContext` extending `Context` then this `BeanManager.getUnmanagedContext` doesn't fit well. 
Just another rough idea - how about to move this activate/deactivate methods to `AlterableContext`?

mkouba:
My opinion is summarized in https://github.com/cdi-spec/cdi/pull/296#issuecomment-229942823

I'm sorry I just don't find this API clear and easy to understand. Unfortunately, I will not be able to help until Jul 18.

tremes:
I was thinking about this proposal again and I gues we could do the following. I think this "new class" (currently `UnmanagedContext`) can extend `Context` (or `AlterableContext`) and we can simply adjust the `BeanManager#getContext` method javadoc to omit the word "active" and remove this new method `BeanManager#getUnmanagedContext`.
It doesn't seem to me as a significant backward compatibility break. `BeanManager#getContext` method previously returned only active context or threw an exception. Now it will return active or inactive context. WDYT?

manovotn:
One more idea...
Could we keep this "new class" (`UnmanagedContext` for the time beiing) in the context hierarchy (`extends [Alterable]Context`) and in the same time keep both methods on `BeanManager`?

`BeanManager#getContext` would still behave as it does - throw an exception if `UnmanagedContext` is not running. Return the instance otherwise.
`BeanManager#getUnmanagedContext` would allow you to obtain the inactive `UnmanagedContext` to actually activate it.

Or am I missing something obvious here?

api/src/main/java/javax/enterprise/context/spi/Context.java
```diff
@@ -58,7 +58,7 @@
      * 
      * @return the scope
      */
-    public Class<? extends Annotation> getScope();
```
Emily-Jiang: All methods on interfaces are public by default. What is the point to change them?

manovotn: I believe this is just a clean up. As you said they are public "by default" so there is no need to explicitly state it.
My guess is that John's IDE was going nuts seeing all those modifiers :)

manovotn:
@johnament WDYT about what Tomas and I wrote earlier in the comments?
It seems we still search a way to align the `UnmanagedContext` with the current Context hierarchy we have...

api/src/main/java/javax/enterprise/context/spi/Context.java
johnament: Yep, I may revert it before we're done.  We should clean all of that up IMHO.

johnament:
@manovotn I'm still concerned about the hierarchy.  However, I still feel like extending Context is a bad idea.  What concerns do you have over the new base interface, assuming that deactivate doesn't get moved?

manovotn:
I am on the opposite side of the barrier you might say. I think it better to extend current context hierarchy and preserve current structure rather than copying it anew with most methods identical.

If the only reason not to extend current structure is that you could register it with ABD, then I think this can be handled in implementations (check `instanceof ManageableContext` and you have it) plus we can adjust spec accordingly. OTOH having it as we do now exposes the problem with `BeanManager.getContext` which may lead to something not really backward compatible if I get it correctly.

As for not percieving Manageable**Context** as Context, well... why do we even talk about it in terms of Contexts then? As Martin pointed out, Context is not only 'something always active' and I agree with that.

api/src/main/java/javax/enterprise/context/spi/ManageableContext.java
```diff
@@ -0,0 +1,116 @@
...
+     * @throws ContextNotActiveException if the context is not active
+     */
+    void destroy(Contextual<?> contextual);
+
+    /**
+     * Starts the given context, allowing beans to be looked up within the bounds of that context.
+     *
+     * @throws IllegalStateException if the context has already been started.
+     */
+    void activate();
```
mkouba: This method could return the context instance so that it's easier to use with `try-with-resource` block:
````
try (ManageableContext ctx = beanManager.createManagedContext(RequestScoped.class).activate()) {
  ...
}
````

File: spec/src/main/asciidoc/core/scopescontexts.asciidoc [hunk 4]
```diff
@@ -126,6 +126,63 @@ The context object is responsible for destroying any contextual instance it crea
...
+    public void run() {
+        BeanManager beanManager = CDI.current().getBeanManager();
+
+        try(ManageableContext context = beanManager.createManagedContext(RequestScoped.class)) {
+            context.activate();
+            Set<Bean<?>> beans = beanManager.getBeans(MyBean.class);
+            Bean<?> bean = beanManager.resolve(beans);
+            CreationalContext<?> creationalContext = beanManager.createCreationalContext(bean);
+            MyBean myBean = (MyBean)beanManager.getReference(bean, MyBean.class, creationalContext);
+            // do some work against that bean
```
mkouba: I would rather recommend to use:
````
CDI cdi = CDI.current();
try (ManageableContext ctx = cdi.getBeanManager().createManagedContext(RequestScoped.class)) {
  ctx.activate();
  MyBean myBean = cdi.select(MyBean.class).get();
  // do some work against that bean
}
````

File: api/src/main/java/javax/enterprise/inject/spi/BeanManager.java [hunk 5]
```diff
@@ -406,14 +408,25 @@
...
+    /**
+     * Creates a new {@link ManageableContext} that represents the given built-in scope.  This ManageableContext is not
+     * yet active, it must be activated when ready to use.  Upon activation, you may use it to retrieve beans from
+     * the context.
+     *
+     * @param scopeType the  {@linkplain javax.enterprise.context scope}
+     * @return the {@linkplain ManageableContext context object}
+     * @throws IllegalArgumentException if the given scopeType is not a valid built in scope.
+     */
+    ManageableContext createManageableContext(Class<? extends Annotation> scopeType);
```
mkouba: I think that `create` is not appropriate here. In the end, a context is/is not active with respect to the current thread. So we could return the same instance of MC.

antoinesd: Sorry @mkouba, I don't understand your suggestion. Could you detail it.
thx

mkouba: The point is that a context is active with respect to the current thread and so it's not necessary to create a new managable context instance per invocation. It's the same as with registering regular contexts - an extension developer registers a single instance of context, see also `javax.enterprise.inject.spi.AfterBeanDiscovery.addContext(Context)`. Of course, it would be a different case if the EG decides to specify `ManageableContext` outside the context hierarchy (and possibly rename it).

api/src/main/java/javax/enterprise/context/spi/ManageableContext.java
johnament: as long as its clear that its returning `this` and not a new `ManageableContext`

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
johnament: If that's still your opinion, then its not clear at all what this is trying to do and may be the root of why you and I are so far separated on this subject.  This method specifically creates a new `ManageableContext`, it's not bound to anything per se.  Repeated calls may end up activating the same context over and over again.  To put it another way...

```
ManageableContext c1 = beanManager.createManageableContext(RequestScoped.class);
ManageableContext c2 = beanManager.createManageableContext(RequestScoped.class);
assert !c1.equals(c2);
```

mkouba: I kind of understand your point, but it's not legal to have more than one active context object for the given scope type (see also 6.5.1. The active context object for a scope).

johnament: Thats why activate throws exceptions.

antoinesd: Guys. I think most of us are a bit lost here. Each of us try to figure how this feature will be used and assume different use case.

For instance I wonder if this feature will help 3rd party framework devs to reuse our built-in context by making them active/inactive when needed? I'm feeling that the answer is no..
What other use cases are covered here?

Shouldn't we step back and put this in perspective and describe the user story behind it?

johnament: @antoinesd could you explain why you think it won't help 3rd party fw devs?

antoinesd: I don't say it won't help them. I say that one of my use case is the 3rd party fw dev that want to change activation of @RequestScoped beans without recreating his own request context and use the built-in one (i.e WebSocket spec example). I'm not sure this feature will allow that but I may be wrong...

mkouba: @johnament 
> This method specifically creates a new ManageableContext, it's not bound to anything per se.

Ok, it's not bound to anything until activated - after activation it must be "bound" to the current thread. Otherwise, it would contradict the spec and a user would not be able to inject normal scoped beans without the reference to the context object... 

@antoinesd Yep, it's quite complicated topic. And an ideal topic for F2F BTW ;-)

johnament: Ok, I think that's clarified.

johnament:
Pinging for comments.

antoinesd:
Clarification looks good for me. Usage is now clearer.
Do we reached an agreement or should we continue discussing this PR (i.e the class hierarchy for `ManageableContext `) @mkouba ?

mkouba:
I still don't understand why `BeanManager.createManageableContext()` returns a new instance. All my arguments are explained above. 

Actually, why do we add a new method to `BeanManager`? The first proposal had a built-in bean and I like this approach more. The truth is we would have to introduce some qualifier to distinguish the builtin contexts: `CDI.current().select(BuiltinContext.RequestLiteral, ManageableContext.class).get()`.

Also it's not clear which built-in contexts are supported. I think only request context and "transient" conversation context make sense.

The class hierarchy is also something which deserves discussion. Right now, I don't see any strong arguments "why ManageableContext is not in the Context hierarchy".

But maybe I'm missing something. Feedback from OWB guys would be great.

johnament:
@mkouba we introduced a method on bean manager rather than a built in bean because of the target audience.  If they're doing CDI integration, they're already working with BeanManager, know how to get a reference, etc.  Using a built in bean means they need to look up yet another component.  It being dependent scoped (by definition) means they also need to worry about clean up.

mkouba:
Well, I don't think any cleanup is needed here. So the only difference would be lookup:
```
CDI.current().getBeanManager().getManageableContext(RequestScoped.class);
// VS
CDI.current().select(BuiltinContext.RequestLiteral, ManageableContext.class).get();
```
WRT target audience - fair enough but on the other hand, `BeanManager` is also source of confusion, esp. if obtained with `CDI.current().getBeanManager()` (note that for example `BeanManager.getBeans()` does not cover `CDI.current()` use case at all).

johnament:
Assuming they're using CDI.current(), sure.  There's still libraries doing the many hooped approach to looking up bean manager (e.g. another redhat project https://github.com/resteasy/Resteasy/blob/master/resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/CdiInjectorFactory.java#L132 )

antoinesd:
@mkouba this 2 audience story was discussed more than one month ago. The part with Interceptor and injection should be provided in a second time when the basic for low level integration were available
Reading all the discussion on this PR I really wonder if we share the same goal for this feature...

PR #297, opened 2016-07-28
CDI-602 Set UnmanagedInstance.disposed to true during dispose()

mkouba:


antoinesd:
+1 for me

tremes:
I think this `IllegalStateException` could be mentioned also in spec text. WDYT?

mkouba:
@tremes Assertions may be implied, right?

antoinesd:
@tremes what Martin did was to make working a use case described in javadoc.
The problem with `Unmanaged` is that the spec doesn't describe the class, just show an example in 11.3.4.

PR #298, opened 2016-07-28
CDI-596 ProcessAnnotatedType - clarify what happens if both setAnnotatedType() and configurator are used within observer notification

antoinesd:
An IllegalStateException is thrown

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/ProcessAnnotatedType.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/ProcessAnnotatedType.java
@@ -24,9 +24,9 @@ import javax.enterprise.inject.spi.builder.AnnotatedTypeConfigurator;
  * the declared annotations.
  * </p>
  * <p>
- * Any observer of this event is permitted to wrap and/or replace the {@link javax.enterprise.inject.spi.AnnotatedType}. The
- * container must use the final value of this property, after all observers have been called, to discover the types and read the
- * annotations of the program elements.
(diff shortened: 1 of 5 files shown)


api/src/main/java/javax/enterprise/inject/spi/ProcessAnnotatedType.java
```diff
@@ -24,9 +24,9 @@
  * the declared annotations.
  * </p>
  * <p>
- * Any observer of this event is permitted to wrap and/or replace the {@link javax.enterprise.inject.spi.AnnotatedType}. The
- * container must use the final value of this property, after all observers have been called, to discover the types and read the
- * annotations of the program elements.
+ * Any observer of this event is permitted to wrap and/or replace the {@link javax.enterprise.inject.spi.AnnotatedType} by calling either {@link #setAnnotatedType(AnnotatedType)} or {@link #configureAnnotatedType()}.
+ * If both methods are called in the observer method, an {@link IllegalStateException} is thrown.
```
mkouba: If both methods are called within an observer notification...

mkouba:
This should apply to `ProcessObserverMethod`, `ProcessBeanAttributes` and `ProcessInjectionPoint` as well.

antoinesd:
Yes but I wanted to have the correct wording before going for the other 3...
If we are good, I go for it now...

api/src/main/java/javax/enterprise/inject/spi/ProcessAnnotatedType.java
antoinesd: ok, making the same change in spec wording.

antoinesd:
IMO Should be ready to merge now. WDYT @mkouba?

mkouba:
Yep, looks good.

PR #300, opened 2016-09-07
CDI-626 How should CDI.current() and CDI.getBeanManager() behave for non-CDI apps?

antoinesd:


spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -235,7 +235,8 @@ public abstract class CDI<T> implements Instance<T> {
 }
 ----
 
-A portable extension or other object may obtain a reference to the current container by calling `CDI.current()`. `CDI.getBeanManager()` may be called at any time after the container fires the `BeforeBeanDiscovery` container lifecycle event until the container fires the `BeforeShutdown` container lifecycle event.
+A portable extension or other object may obtain a reference to the current container by calling `CDI.current()`.
+`CDI.getBeanManager()` may be called at any time after the container fires the `BeforeBeanDiscovery` container lifecycle event until the container fires the `BeforeShutdown` container lifecycle event.
```
johnament: This usage is misleading, as this is not a static method.  If we're saying that `CDI.current()` can blow up, then the case Mark described should never happen.

api/src/main/java/javax/enterprise/inject/spi/CDI.java
```diff
@@ -54,6 +54,7 @@
     /**
      *
      * @return the {@link CDIProvider} retrieved by serviceloader or setted by user
+     * @throws IllegalStateException if no {@link CDIProvider} is available
      */
     public static CDIProvider getCDIProvider() {
```
mkouba: This method should not be public or what's the point?

api/src/main/java/javax/enterprise/inject/spi/CDI.java
```diff
@@ -119,6 +120,7 @@ private static void findAllProviders() {
      * Get the CDI BeanManager for the current context
      *
      * @return the BeanManager
+     * @throws IllegalStateException if no {@link CDIProvider} is available
```
mkouba: This is unnecessary - `CDI.current()` blows up if no CDIProvider is available.

mkouba:
I'm missing the main point - `javax.enterprise.inject.spi.CDIProvider.getCDI()` should throw an `IllegalStateException` if no CDI container is available.

spec/src/main/asciidoc/core/spi.asciidoc
mkouba: Isn't it only formatting which changed here?

tremes: Yes it seems so but it's still a little bit misleading because `CDI.getBeanManager()` is not static as John pointed out.

antoinesd: Yes I only added a line feed (one sentence by line).
If it's miselading you should probably fill a new ticket @johnament.

api/src/main/java/javax/enterprise/inject/spi/CDI.java
antoinesd: You're right

api/src/main/java/javax/enterprise/inject/spi/CDI.java
antoinesd: become useful if previous method is nor more public ;)

antoinesd:
I amended PR in that direction @mkouba

api/src/main/java/javax/enterprise/inject/spi/CDI.java
```diff
@@ -120,7 +119,7 @@ private static void findAllProviders() {
      * Get the CDI BeanManager for the current context
      *
      * @return the BeanManager
-     * @throws IllegalStateException if no {@link CDIProvider} is available
+     * @throws IllegalStateException if no CDI container is available
```
mkouba: Please remove this line completely. You cannot call this method if it's not available because `CDI.current()` must call `CDIProvider.getCDI()` and so would get ISE before...

mkouba:
Looks good now.

spec/src/main/asciidoc/core/spi.asciidoc
johnament: I'll create a separate ticket.

johnament:
LGTM

PR #301, opened 2016-09-08
CDI-620 Not clear what the id for an annotated type is.

antoinesd:
add clarification for id
correct parameter switching
correct the spec that was wrong


johnament:
LGTM

PR #303, opened 2016-09-12
CDI-625 Make it clear when exactly are context init/destroy events fired

mkouba:
- add @BeforeDestroyed and @AfterDestroyed
- deprecate @Destroyed

### DIFF
--- a/api/src/main/java/javax/enterprise/context/BeforeDestroyed.java
+++ b/api/src/main/java/javax/enterprise/context/BeforeDestroyed.java
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source
- * Copyright 2013, Red Hat, Inc., and individual contributors
+ * Copyright 2016, Red Hat, Inc., and individual contributors
  * by the @authors tag. See the copyright.txt in the distribution for a
  * full listing of individual contributors.
  *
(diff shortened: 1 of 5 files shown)


api/src/main/java/javax/enterprise/context/BeforeDestroyed.java
```diff
@@ -0,0 +1,83 @@
...
+     * The scope for which to observe destruction
+     */
+    Class<? extends Annotation> value();
+
+    /**
+     * Supports inline instantiation of the {@link BeforeDestroyed} qualifier.
+     */
+    public final static class Literal extends AnnotationLiteral<BeforeDestroyed> implements BeforeDestroyed {
+
+        public static final Literal REQUEST = of(RequestScoped.class);
```
johnament: do all of these need to be present? Seem duplicated.

mkouba: Nope. This one is for `@BeforeDestroyed(RequestScoped.class)`. The one in `Initialized` is for `@Initialized(RequestScoped.class)` and so on...

tremes:
Looks good.

api/src/main/java/javax/enterprise/context/Destroyed.java
```diff
@@ -33,13 +33,12 @@
 
 /**
  * <p>
- * The <code>@Destroyed</code> qualifier.
+ * This qualifier is deprecated from CDI 2.0 and {@link BeforeDestroyed} or {@link AfterDestroyed} should be used instead.
```
manovotn: Shouldn't we add a note that the new `@AfterDestroyed` behaves equally to this deprecated one?

mkouba: Yep, makes sense.

manovotn:
I think it's good now, +1

antoinesd:
Ok for me but I just have a backward compatibility question. Shouldn't we keep mentioning @Destroyed as being deprecated in the spec like we did with @New ?

mkouba:
@antoinesd Do you mean the spec text?

antoinesd:
@mkouba. Yes, we did that for [@New](http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#new).
I'm not a big fan of letting deprecated features in the spec, but I'm afraid we don't have choice... wdyt ?

mkouba:
If it's not required by some JCP rule I would remove @Destroyed from the spec text.

manovotn:
I think we shouldn't mention it. That would just bloat the spec.
We still remain backward compatible and `@Destroyed` will work for those who have it in code now. Plus the new annotation is nothing far-fetched from the old one - it is easy to figure it out.

antoinesd:
Guys I'm not saying I **want** it to stay in the doc, I'm asking if it's allowed to remove it. I just sent a mail to Linda to be sure. Hope she'll tell we can do it...

antoinesd:
Found: https://java.net/projects/javaee-spec/pages/CompatibilityRequirements
point 6 deals with deprecated methods (not exactly our use case, but very close):

> 6. An expert group may decide to designate a method or interface as being "deprecated" when its use is no longer recommended or when the expert group considers that a different interface or method should be used instead. This can be stated in the specification and in the relevant javadoc. However the @Deprecated annotation must not be used as this is noisy and annoying for users. The only exception to this is if there is a serious security or functional defect in the method or interface.

wdyt ?

manovotn:
Pretty much close to our case. 

> This _**can**_ be stated in the specification

The way I understand it, it's up to us.

mkouba:
@cdi-spec/eg @johnament PR changed - follows EG decision.

antoinesd:
LGTM

PR #304, opened 2016-10-03
CDI-634 clarify contextual ref. validity.

manovotn:
A proposed clarification for CDI-634.

I am not sure if we should place this in section `6.5.4. Contextual reference validity` (as proposed here) since it is generally valid for both, EE and SE. Or if it should go into SE part of the doc.

Feel free to suggest && comment.

### DIFF
--- a/spec/src/main/asciidoc/core/scopescontexts.asciidoc
+++ b/spec/src/main/asciidoc/core/scopescontexts.asciidoc
@@ -283,6 +283,7 @@ The validity of a contextual reference for a bean depends upon whether the scope
 
 * Any reference to a bean with a normal scope is valid as long as the application maintains a hard reference to it.
 However, it may only be invoked when the context associated with the normal scope is active. If it is invoked when the context is inactive, a `ContextNotActiveException` is thrown by the container.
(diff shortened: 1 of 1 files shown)


johnament:
@manovotn the only thing I don't get is why this applies to only normal scoped beans.

mkouba:
@johnament Hm, for pseudo-scopes you don't get a client proxy, i.e. you always have a hard reference to a contextual instance. So there is a bigger chance it will work (if you don't inject any normal scoped beans ;-). I think it's ok to leave it as non-portable for pseudo-scopes.

johnament:
@mkouba not always, if you have interceptors enabled for pseudo scopes (`@Dependent`) you'll get a proxy now.

mkouba:
@johnament Yes, but it's not a _contextual reference (client proxy)_ per definition. It's a _contextual instance_ - an internal container construct. In Weld this is a subclass proxy. Interceptors are not the main purpose of client proxies - see also http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#client_proxies.

johnament:
ok then its fine @mkouba.  Should we mention what happens for a contextual instance?

antoinesd:
This looks to me. Do we have any blocker here?

PR #305, opened 2016-10-04
CDI-30 Added a context controller specifically for requests and an in…

johnament:
…terceptor binding.


api/src/main/java/javax/enterprise/context/control/RequestContextController.java
```diff
@@ -0,0 +1,68 @@
...
+   /**
+    * Activates a RequestContext for the current thread if one is not already active.
+    * @return the underlying {@see Context} object associated with the request scope that was just activated, otherwise null
+    * if no context was activated (due to it already being active)
+    */
+   Context activate();
+
+   /**
+    * Deactivates the current Request Context if it was started by this context controller.  If the context is active
+    * but was not started by this controller, then it may not be stopped by this controller, this method will do nothing.
```
tremes: What's the difference between "start" and "activate" the context? It seems to me confusing. I would try to use one single term.

johnament: For the sake of the PR, assume they're the same.  I'll clean up any wording later on.

mkouba:
John, the API looks good. I have few ideas though ;-)

* do we really need `javax.enterprise.context.control.ContextController`?, the `getScope()` method seems rather useless to me...
* we should probably rename `ActivateRequestScope` to `ActivateRequestContext` as it's context which gets activated (I rember we discussed this on a Weld issue and yes, we've changed our mind and renamed the binding in  [Weld 2.4 API](https://github.com/weld/api/blob/2.4/weld/src/main/java/org/jboss/weld/context/activator/ActivateRequestContext.java) ;-)
* `RequestContextController.activate()` - I would prefer to return a boolean, do you expect users to call `Context` methods after activation?

johnament:
@mkouba agreed on #2 point.

RE #1 - I generally prefer to be more fine grained in APIs than all out.  While we only need `RequestContextController` having a base interface for common methods (I suspect we'll need something for session before the end of the release).

RE #3 - I'd prefer to return the `Context` object, but at the same time think its odd that it returns null.  I'll change it to boolean.

mkouba:
> having a base interface for common methods

OK

johnament:
Anything blocking merging this then @mkouba ?  @struberg  any comments?

struberg:
will review in the next few hours.

spec/src/main/asciidoc/core/scopescontexts.asciidoc
```diff
@@ -228,6 +228,34 @@ From time to time, the container must obtain an _active context object_ for a ce
 
 If there is exactly one active instance of `Context` associated with the scope type, we say that the scope is _active_.
 
+==== Activating Built In Contexts
```
johnament: I just want to check with everyone.  Is this in the right place? Is it enough detail?

manovotn: Hmm.. I was thinking maybe putting in as the last chapter under `6. Scopes and contexts` would fit better (e.g. create `6.8`)? It just my feeling though, anywhere within contexts chapter is probably a good shot.

spec/src/main/asciidoc/core/scopescontexts.asciidoc
```diff
@@ -228,6 +228,34 @@ From time to time, the container must obtain an _active context object_ for a ce
...
+
+Certain built in contexts support the ability to be activated and deactivated.  This allows developers to control built-in contexts in ways that they could also manage custom built contexts.
```
manovotn: I would start with something like this in order to emphasize that you normally don't do this, but you can should you need it:
`Built-in context activation and deactivation is managed by CDI container but sometimes it might be necessary to handle this manually. Certain contexts have this capability and therefore allows developers to control built-in contexts in ways that they could also manage custom built contexts.`
WDYT? Just wanted to point it out so that we avoid people (especially in SE I guess) who would overuse this.

manovotn:
Just a wild thought...
What happens if there are multiple same active contexts? E.g. multiple Request scope contexts. In Weld we throw IllegalStateException ATM. We might want to clarify it within bounds of this issue because you can have following situtation:

1. I manually activate req scope with controller
2. While I am running some operations container attempts to activate the scope as well
3. What is the outcome? Exception? Currently undefined.
4. Who has the right to deactivate the context now? Container, or can I do it manually via controller?

johnament:
@manovotn This API is specifically activating once, and only if not already active.

manovotn:
@johnament

manovotn:
Oops, sorry, clicked "close and comment" instead of "comment", sorry. Reopening.

@johnament it still doesn't answer the question. I think the container can try to active the context itself after you did so manually, and that is not covered.

mkouba:
@manovotn @johnament It's actually defined in "6.5.1. The active context object for a scope":

> If more than one active context object exists for the given scope type, the container must throw an IllegalStateException.

manovotn:
IC, all clear then. Thanks Martin.

johnament:
Ok, then I misunderstood your question.  Do we want to clarify this?

antoinesd:
Looks Good to Me, is there any objection to merge this PR (at last;) )

PR #306, opened 2016-10-07
CDI-614 remove InjectionPointConfigurator read methods.

tremes:
I consider them completely useless. Please prove if I am wrong and demonstrate some nice use case when it could be useful.

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/configurator/InjectionPointConfigurator.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/configurator/InjectionPointConfigurator.java
@@ -17,19 +17,15 @@
 
 package javax.enterprise.inject.spi.configurator;
 
-import javax.enterprise.inject.spi.AnnotatedField;
-import javax.enterprise.inject.spi.AnnotatedParameter;
 import javax.enterprise.inject.spi.Bean;
 import javax.enterprise.inject.spi.InjectionPoint;
 import javax.enterprise.inject.spi.ProcessInjectionPoint;
 import java.lang.annotation.Annotation;
-import java.lang.reflect.Field;
-import java.lang.reflect.Parameter;
 import java.lang.reflect.Type;
 import java.util.Set;
 
 /**
- * This API is an helper to configure a new {@link InjectionPoint} instance.
+ * This API is an helper to configure an existing {@link InjectionPoint} instance.
  * CDI container must provides an implementation of this interface.
  *
  * This builder is not thread safe and shall not be used concurrently.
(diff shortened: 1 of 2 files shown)


api/src/main/java/javax/enterprise/inject/spi/builder/InjectionPointConfigurator.java
```diff
@@ -41,51 +37,6 @@
 public interface InjectionPointConfigurator {
 
     /**
-     * Read the InjectionPoint information from the given {@link Field}.
-     * All relevant information is overwritten.
-     *
-     * @param field defining the InjectionPoint.
-     * @return self
-     */
-    InjectionPointConfigurator read(Field field);
```
johnament: I'm pretty sure that by removing this methods, you're removing the simplified way of loading from an arbitrary field/method parameter.

tremes: Why would you need that? If you have some example I am all ears.I think you can simply catch it in PIP and you can't load it anywhere else AFAIK.

johnament: Suppose I have a field

```
@ConfigSetting("boop")
private String boop;
```

I could read the field, and add it as an injectable field.  It wouldn't have a PIP since it doesn't have `@Inject`

tremes: Ok but how would you add it as injectable field then? Note that `InjectionPointConfigurator` has only `addQualifier` methods.

manovotn: Hi John, I think that the point of `IPConfigurator` if **not** to give it a field and create an IP point from it (a new one that is). The goal is rather to be able to **temper with currently existing IPs** - meaning changing their qualifiers as there isn't much else you can temper with.

The truth is, we need to also correct the javadoc. It currently says `This API is an helper to configure a new {@link InjectionPoint} instance.`. Which should be changed into something like `This API is a helper to configure an existing {@link InjectionPoint} instance.` Same problem goes for spec - [11.5.7.1](http://docs.jboss.org/cdi/spec/2.0.EDR2/cdi-spec.html#injection_point_configurator). It would be good to include doc/spec text changes i nthis PR as well.

tremes: @manovotn yes I agree. I wanted to go through the javadoc and spec text as well. I'll update. 

Additionaly I wanted to emphasize that it doesn't make sense to me to introduce such "implicit rule" for IPs. It would be just confusing and  brings next questions about its functionality (Does PIP get fired for this and etc.)

manovotn:
Looks good to me now.

antoinesd:
LGTM

tremes:
Updated.

PR #307, opened 2016-10-10
CDI-636 Introduce the ability to stream bean instances, as well as parallel stream them.

johnament:
Introduces `stream()` and `parallelStream()` methods for bean instances.

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/Instance.java
+++ b/api/src/main/java/javax/enterprise/inject/Instance.java
@@ -17,10 +17,11 @@
 
 package javax.enterprise.inject;
 
-import java.lang.annotation.Annotation;
-
 import javax.enterprise.util.TypeLiteral;
 import javax.inject.Provider;
+import java.lang.annotation.Annotation;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 /**
  * <p>
@@ -132,7 +133,7 @@ public interface Instance<T> extends Iterable<T>, Provider<T> {
      * @throws IllegalArgumentException if passed two instances of the same qualifier type, or an instance of an annotation that
      *         is not a qualifier type
      */
-    public Instance<T> select(Annotation... qualifiers);
+    Instance<T> select(Annotation... qualifiers);
 
     /**
      * <p>
@@ -146,7 +147,7 @@ public interface Instance<T> extends Iterable<T>, Provider<T> {
      * @throws IllegalArgumentException if passed two instances of the same qualifier type, or an instance of an annotation that
      *         is not a qualifier type
      */
-    public <U extends T> Instance<U> select(Class<U> subtype, Annotation... qualifiers);
+    <U extends T> Instance<U> select(Class<U> subtype, Annotation... qualifiers);
 
     /**
      * <p>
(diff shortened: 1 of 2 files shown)


api/src/main/java/javax/enterprise/inject/Instance.java
```diff
@@ -132,7 +133,7 @@
      * @throws IllegalArgumentException if passed two instances of the same qualifier type, or an instance of an annotation that
      *         is not a qualifier type
      */
-    public Instance<T> select(Annotation... qualifiers);
+    Instance<T> select(Annotation... qualifiers);
```
johnament: Hope no one minds the clean up.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -668,14 +668,18 @@ The `Instance` interface provides a method for obtaining instances of beans with
...
-    public Instance<T> select(Annotation... qualifiers);
-    public <U extends T> Instance<U> select(Class<U> subtype, Annotation... qualifiers);
-    public <U extends T> Instance<U> select(TypeLiteral<U> subtype, Annotation... qualifiers);
+    Instance<T> select(Annotation... qualifiers);
+    <U extends T> Instance<U> select(Class<U> subtype, Annotation... qualifiers);
+    <U extends T> Instance<U> select(TypeLiteral<U> subtype, Annotation... qualifiers);
 
-    public boolean isUnsatisfied();
-    public boolean isAmbiguous();
+    Stream<T> stream();
```
johnament: Please note the order.  I figure reading should be before the booleans and destroy.

api/src/main/java/javax/enterprise/inject/Instance.java
```diff
@@ -160,7 +161,31 @@
...
+
+    /**
+     * <p>
+     * When called, provides back a Stream of the beans available in this Instance.  If no beans are found, it
+     * returns an empty stream.
+     * </p>
+     *
+     * @return a <tt>Stream</tt> representing the beans associated with this {@link Instance} object
+     */
+    default Stream<T> parallelStream(){
```
mkouba: I'm not so sure it's a good idea to promote parallel streams in CDI. AFAIK all parallel streams use common fork-join thread pool.

johnament: I don't have a strong use case for it.  I can see some usefulness with CDI SE, for start up purposes.

manovotn: I think it's fair enough for SE. We could keep it and add a javadoc notion about the executor used and the fact that you might not want to use it in EE.

johnament: At the same time, since a regular stream can be made parallel, I'm not sure we change much of anything.  The impl's can be container aware, and if needed can look up an executor.

manovotn: Not sure I am following here, how would you make it "container aware". 

Btw @tremes raised a good point - can we allow parallel processing at all? It spawns new threads in which we **do not** propagate context. Therefore if there are some, say, session scoped beans in the Instance, their context would not work in other thread. Or am I missing something?

mkouba: I wouldn't add this to the API. The use case is disputable (even in SE). And if a user really wants to do it, she/he can use `StreamSupport.stream(instance.spliterator(), true)`.

johnament: Or just call `.paralllel()` on the stream.  I'll drop it.  If I do, good to go?

johnament:
Dropped parallel.

api/src/main/java/javax/enterprise/inject/Instance.java
mkouba: Yep, looks good now.

johnament:
Safe to merge? @mkouba @antoinesd ?

mkouba:
Looks good!

antoinesd:
LGTM

PR #308, opened 2016-10-10
CDI-523 Use a proper entity manager that can be closed.

johnament:
Fixed the example to use proper producer/disposer pattern for un-managed entity manager.

### DIFF
--- a/spec/src/main/asciidoc/core/implementation.asciidoc
+++ b/spec/src/main/asciidoc/core/implementation.asciidoc
@@ -398,11 +398,19 @@ public class UserDatabaseEntityManager {
 [source, java]
 ----
 public class Resources {
-    
-    @PersistenceContext
+
+    private EntityManagerFactory emf;
+
+    @PostConstruct
+    public void setupEntityManagerFactory() {
+        emf = Persistence.createEntityManagerFactory("userDatabase");
+    }
+
     @Produces @UserDatabase
-    private EntityManager em;
-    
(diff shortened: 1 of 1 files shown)


antoinesd:
LGFM

MartinanderssonDotcom:
I don't know why there would ever be a reason to use default `@Dependent` scope on an application-managed entity manager. Shouldn't it be `@RequestScoped?`

With `@Dependent`, the code path during a request execution/processing risk using different persistence contexts and greatly increase the probability for an `OptimisticLockException `even though the request was never handled by more than one thread serially (persistence context 1 see version 1, persistence context 2 update version to 2, when persistence context 1 wants to commit version checking fail). Again, this I don't know why anyone would want.

johnament:
@MartinanderssonDotcom You're presuming a lot about what the `@UserDatabase` annotation is.  Maybe its a qualifier.  Maybe its a stereotype that combines `@RequestScoped`, `@Transactional` and another qualifier.  The purpose of the example is to focus on how to declare a disposer method.

MartinanderssonDotcom:
Oh that's right. I forgot all about the `@UserDatabase` annotation. I still think we should throw `@RequestScoped` in there. Our examples should follow best practice and not add to confusion - which I think this one does. Just imagine how much human time and headache one extra word/annotation on the producer would save all over this globe when so many people read this spec. I don't think most people will think "oh @UserDatabase must be a stereotype annotated @RequestScoped".

PR #309, opened 2016-10-11
CDI-627 also re-enable the rules of CDI-1.0

struberg:
This is for fixing a backward-incompatible change we introduced in CDI-1.1.

The whole reason for verifying the <class> entries is to ensure that a user
doesn't have typos or list classes which do not exists or are no
'candidates' for being an Alternative at least.
So it's all about preventing user errors. Disallowing to veto() @Alternative
classes (which was perfectly allowed in CDI-1.0) was overreaching.

### DIFF
--- a/spec/src/main/asciidoc/core/injectionandresolution.asciidoc
+++ b/spec/src/main/asciidoc/core/injectionandresolution.asciidoc
@@ -74,7 +74,12 @@ An alternative is selected for the bean archive if either:
 ----
 
 Each child `<class>` element must specify the name of a bean class of an alternative bean.
-If there is no bean whose bean class has the specified name, or if no bean whose bean class has the specified name is an alternative, the container automatically detects the problem and treats it as a deployment problem.
+For each of them the container verifies that either:
+
+* a class T exists which is annotated with `@Alternative` or contains a producer field or producer method which is annotated with `@Alternative`, or
+* an alternative `Bean<T>` exists for the bean class.
+
+Otherwise the container automatically detects the problem and treats it as a deployment problem.
 
 Each child `<stereotype>` element must specify the name of an `@Alternative` stereotype annotation.
 If there is no annotation with the specified name, or the annotation is not an `@Alternative` stereotype, the container automatically detects the problem and treats it as a deployment problem.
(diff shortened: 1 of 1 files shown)


manovotn:
LGTM

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -74,7 +74,12 @@ An alternative is selected for the bean archive if either:
 ----
 
 Each child `<class>` element must specify the name of a bean class of an alternative bean.
```
tremes: If we go for it (I think it's basically ok but on the other hand we are going to confuse the users again little bit) then this sentence needs some change as well IMHO. Because when I veto it then it's not a bean anymore. So just something like?:
` Each child `<class>` element must specify the name of a class of an alternative bean candidate.` or 
`Each child `<class>` element must specify the name of a class which is annotated with @Alternative.`
WDYT?

struberg: I removed the first sentence and merged it with the 2nd. Should make things clearer. The first sentence was redundant with the paragraph above anyway.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -73,8 +73,12 @@ An alternative is selected for the bean archive if either:
 </beans>
 ----
 
-Each child `<class>` element must specify the name of a bean class of an alternative bean.
-If there is no bean whose bean class has the specified name, or if no bean whose bean class has the specified name is an alternative, the container automatically detects the problem and treats it as a deployment problem.
+For each child `<class>` element the container verifies that either:
+
+* a class T exists which is annotated with `@Alternative` or contains a producer field or producer method which is annotated with `@Alternative`, or
```
johnament: Please make T `T` instead since its code

struberg: sure, txs!

antoinesd:
LGTM

tremes:
I am looking again at this because of TCK assertion update and I am affraid that this wording excludes alternative stereotypes.

struberg:
good catch @tremes there should be another bullet, similar to the first one - just for alternative stereotypes

Emily-Jiang:
Can this be fixed in Weld?

PR #310, opened 2016-10-11
CDI-420 Introduce a way to discover all types but make only scope annotated classes eligible for bean

struberg:
<trim/> as discussed during the face2face meeting.

### DIFF
--- a/spec/src/main/asciidoc/core/definition.asciidoc
+++ b/spec/src/main/asciidoc/core/definition.asciidoc
@@ -493,7 +493,7 @@ Alternatively, a scope type may be specified using a stereotype annotation, as d
 
 ==== Default scope
 
-When no scope is explicitly declared by annotating the bean class or producer method or field the scope of a bean is defaulted.
+When no scope is explicitly declared by annotating the bean class or producer method or field and the bean is not deployed in a <<trimmed_bean_archive>> the scope of a bean is defaulted.
 
 The _default scope_ for a bean which does not explicitly declare a scope depends upon its declared stereotypes:
 
--- a/spec/src/main/asciidoc/core/definition.asciidoc
+++ b/spec/src/main/asciidoc/core/definition.asciidoc
@@ -499,7 +499,8 @@ The _default scope_ for a bean which does not explicitly declare a scope depends
 
 * If the bean does not declare any stereotype with a declared default scope, the default scope for the bean is `@Dependent`.
 * If all stereotypes declared by the bean that have some declared default scope have the same default scope, then that scope is the default scope for the bean.
-* If there are two different stereotypes declared by the bean that declare different default scopes, then there is no default scope and the bean must explicitly declare a scope.
+
+If there are two different stereotypes declared by the bean that declare different default scopes, then there is no default scope and the bean must explicitly declare a scope.
 If it does not explicitly declare a scope, the container automatically detects the problem and treats it as a definition error.
 
 
--- a/spec/src/main/asciidoc/core/packagingdeployment.asciidoc
+++ b/spec/src/main/asciidoc/core/packagingdeployment.asciidoc
@@ -163,6 +163,25 @@ For example, consider the follow `beans.xml` file:
 The first exclude filter will exclude all classes in `com.acme.rest` package. The second exclude filter will exclude all classes in the `com.acme.faces` package, and any subpackages, but only if JSF is not available. The third exclude filter will exclude all classes in the `com.acme.verbose` package if the system property `verbosity` has the value `low`. The fourth exclude filter will exclude all classes in the `com.acme.ejb` package, and any subpackages if the system property `exclude-ejbs` is set (with any value) and at the same time, the `javax.enterprise.inject.Model` class is available to the classloader.
 
 
+[[trimmed_bean_archive]]
+
+==== Trimmed bean archive
+
+A bean archive might be marked as to be 'trimmed' by adding the following tag to beans.xml:
+[source, xml]
+<beans xmlns="http://xmlns.jcp.org/xml/ns/javaee">
+    <trim/>
+</beans>
+
+If a bean archive is marked with the `<trim/>` element in it's beans.xml, only AnnotatedTypes will get processed as Beans if they
+
+* have an explicit Scope annotation, or
+* are an Interceptor or a Decorator
+
+after the `ProcessAnnotatedType` event got fired.
+`AnnotatedTypes` with no scope will not get a <<default_scope>>.
+
+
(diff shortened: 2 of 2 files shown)


spec/src/main/asciidoc/core/packagingdeployment.asciidoc
```diff
@@ -163,6 +163,25 @@ For example, consider the follow `beans.xml` file:
...
+A bean archive might be marked as to be 'trimmed' by adding the following tag to beans.xml:
+[source, xml]
+<beans xmlns="http://xmlns.jcp.org/xml/ns/javaee">
+    <trim/>
+</beans>
+
+If a bean archive is marked with the `<trim/>` element in it's beans.xml, only AnnotatedTypes will get processed as Beans if they
+
+* have an explicit Scope annotation, or
+* are an Interceptor or a Decorator
```
tremes: I would simply say `if they have bean defining annotation` (possibly as link to the chapter).

struberg: Maybe "if they have bean defining annotations OR a valid scope". 
By only referring to the bean defining annotations we will loose JSR-330 @Scope scopes.

antoinesd: Sorry to go back that @Scope story. But since the spec is not very supportive for pseudo-scopes what is the interest of supporting except for the love of complexity ?

See https://issues.jboss.org/browse/CDI-226 or this passage of the spec (6.3)

> Any scope that is not a normal scope is called a pseudo-scope. The concept of a current instance is not well-defined in the case of a pseudo-scope.

manovotn:
> as discussed during the face2face meeting.

@struberg as you yourself stated here, this PR should be based on what we agreed on during the f2f. And I am pretty sure there was **no notion of @Scope**. As far as I recall it as was Antoine said above; that it should "behave like `annotated` mode except all AnnotatedTypes are discovered"

I think you should adjust the PR accordingly.

struberg:
It was. And of course we support @Scope scopes. See 6.3. Why would we define how JSR-330 scopes do behave if we do not support them? If we would not support them then we would not have mentioned them at all or even explicitly defined that we do not support them.

Again: @Scope are not part of BeanDefiningAnnotations because there have been problems with it's use in implicit bean archives (jars without any beans.xml). Iirc it was about spring and guice jars. 
But IF there is a beans.xml then we perfectly support them since CDI-1.0. And not by accident but very intentionally!

struberg:
An example of code which explicitly uses @Scope (and not @NormalScoped)
https://github.com/vaadin/cdi/blob/master/vaadin-cdi/src/main/java/com/vaadin/cdi/UIScoped.java

The reason according to Vaadin guys is that they had problems with equals and hashCode (due to our proxies) and thus they opted for non-normalscoped.

spec/src/main/asciidoc/core/packagingdeployment.asciidoc
mkouba: -1 for scope annotation
+1 for `if they have a bean defining annotation`

struberg: @mkouba that trashes a large amount of use cases for no reason!
@antoinesd the spec passage you quoted only means that the instance selection is up to the context implementation. Obviously the Context.get() gets invoked for every injection point. And it's totally up to the respective Context which instance it returns. That's all.

Also read the paragraph below the one you quoted. It makes it clear that @Scope IS supported!

> All pseudo-scopes must be explicitly declared @Scope, to indicate to the container that no client proxy is required.

Why would we have such a paragraph if we would not support JSR-330 scopes?

mkouba:
Mark, I think we should rather update the `12.4.1. Type discovery` section and say something like: If an `AnnotatedType` fired for a type from a trimmed bean archive is not annotated with `...` it is automatically vetoed.

spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -493,13 +493,14 @@ Alternatively, a scope type may be specified using a stereotype annotation, as d
 
 ==== Default scope
 
-When no scope is explicitly declared by annotating the bean class or producer method or field the scope of a bean is defaulted.
+When no scope is explicitly declared by annotating the bean class or producer method or field and the bean is not deployed in a <<trimmed_bean_archive>> the scope of a bean is defaulted.
```
struberg: cannot follow you. 'bean class' in that case most likely doesn't refer to Bean<T>#getBeanClass() but 'managed bean'.

manovotn: I think what's meant here is that if you `trim` the archive, the classes without annotations will not actually become bean classes (you will veto them) and therefore you don't need to explicitly state that the scope will not be defaulted. Hence this change seems unnecessary.

struberg: Yes there are 2 ways to word what we aim for. Either declare that it doesn't get a default scope and thus is not a Bean. Or to veto it. Both ways lead to the same behaviour and are internally pretty much the same

manovotn: I am trying to that either way, you needn't change **this spec section**. The other part of spec changes you made are sufficient.

johnament:
@struberg in this PR, can you also include a `beans_2_0.xsd` with your addition to it?  Or should someone else introduce the xsd so that it can be modified in this PR?

antoinesd:
you're right @johnament. I added https://issues.jboss.org/browse/CDI-638 to track this new xsd

spec/src/main/asciidoc/core/packagingdeployment.asciidoc
```diff
@@ -163,6 +163,25 @@ For example, consider the follow `beans.xml` file:
 The first exclude filter will exclude all classes in `com.acme.rest` package. The second exclude filter will exclude all classes in the `com.acme.faces` package, and any subpackages, but only if JSF is not available. The third exclude filter will exclude all classes in the `com.acme.verbose` package if the system property `verbosity` has the value `low`. The fourth exclude filter will exclude all classes in the `com.acme.ejb` package, and any subpackages if the system property `exclude-ejbs` is set (with any value) and at the same time, the `javax.enterprise.inject.Model` class is available to the classloader.
 
 
+[[trimmed_bean_archive]]
+
+==== Trimmed bean archive
+
+A bean archive might be marked as to be 'trimmed' by adding the following tag to beans.xml:
```
mkouba: "An **explicit** bean archive might be..." (it does not make sense for implicit bean archives)

PR #311, opened 2016-10-11
CDI-449 fix beans.xml files in the spec.

tremes:


spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -64,7 +64,7 @@ An alternative is selected for the bean archive if either:
 ----
 <beans xmlns="http://xmlns.jcp.org/xml/ns/javaee"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-       xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd">
+       xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd" bean-discovery-mode="all" version="1.1">
```
johnament: should we use `version=2.0` since this is for the 2.0 spec?

tremes: Hm so do you suggest to keep `beans_1_1.xsd` and just use `version=2.0`? I am not sure if we can go for this option but I think we already discussed it but can't remember when and where.

johnament: It would make sense for the spec to reference the current version of its assets, e.g. the 2.0 xsd and use version 2.0.

tremes: @antoinesd so are we going to introduce new `beans_2_0.xsd`? Should I update the PR?

johnament: I think we have to have a new `beans_2_0.xsd` for the new `<trim/>` element.  Made a comment on #310 but I don't have a strong opinion on if it should be introduced there or not.

antoinesd: yes the `<trim />` addition will force us to update to `bean_2_0.xsd`. I just created https://issues.jboss.org/browse/CDI-638 for this.

PR #313, opened 2016-10-12
CDI-623 if the configurator is not initialized with any information, …

tremes:
…treat it as deployment problem.


spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1109,9 +1109,11 @@ public interface AfterBeanDiscovery {
 The given `Bean` may implement `Interceptor` or `Decorator`.
 +
 The second version of the method, returns a new `BeanConfigurator` as defined in <<bean_configurator>> to easily configure the `Bean` which will be added at the end of observer invocation.
+If the container is unable to process configurator it automatically detects the problem and treats it as a deployment problem.
```
johnament: Should be "process the configurator" I think, for both of these lines.

tremes: @johnament thanks!! will update.

antoinesd:
LGTM

PR #315, opened 2016-10-17
CDI-580 Allow interceptors and decorators to be applied to the return value of a producer

antoinesd:
CDI-580 Allow interceptors and decorators to be applied to the return value of a producer method
Also works for custom beans create()

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
@@ -17,11 +17,6 @@
 
 package javax.enterprise.inject.spi;
 
-import java.lang.annotation.Annotation;
-import java.lang.reflect.Type;
-import java.util.List;
-import java.util.Set;
-
 import javax.el.ELResolver;
 import javax.el.ExpressionFactory;
 import javax.enterprise.context.ContextNotActiveException;
@@ -33,6 +28,10 @@ import javax.enterprise.inject.AmbiguousResolutionException;
 import javax.enterprise.inject.InjectionException;
 import javax.enterprise.inject.UnsatisfiedResolutionException;
 import javax.enterprise.util.Nonbinding;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.List;
+import java.util.Set;
 
 /**
  * <p>
@@ -88,6 +87,7 @@ import javax.enterprise.util.Nonbinding;
  * @author Antoine Sabot-Durand
  */
 public interface BeanManager {
+
 
     /**
      * <p>
@@ -605,5 +605,16 @@ public interface BeanManager {
      * @since 1.1
      */
     public <T extends Extension> T getExtension(Class<T> extensionClass);
+
+    /**
+     *
+     * Create an {@link InterceptionProxyFactory} for the given {@link CreationalContext} and type.
+     *
+     * @param ctx {@link CreationalContext} for the {@link InterceptionProxyFactory} to create
+     * @param clazz class of the instance to apply proxies on
+     * @param <T> type of the instance to apply proxies on
+     * @return a new {@link InterceptionProxyFactory} to create proxies for an instance of T
+     */
+    <T> InterceptionProxyFactory<T> createInterceptionFactory(CreationalContext<T> ctx, Class<T> clazz);
 
 }
--- a/api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
@@ -614,6 +614,7 @@ public interface BeanManager {
      * @param clazz class of the instance to apply proxies on
      * @param <T> type of the instance to apply proxies on
      * @return a new {@link InterceptionProxyFactory} to create proxies for an instance of T
+     * @since 2.0
      */
     <T> InterceptionProxyFactory<T> createInterceptionFactory(CreationalContext<T> ctx, Class<T> clazz);
 
--- a/api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
@@ -608,14 +608,14 @@ public interface BeanManager {
 
     /**
      *
-     * Create an {@link InterceptionProxyFactory} for the given {@link CreationalContext} and type.
+     * Create an {@link InterceptionFactory} for the given {@link CreationalContext} and type.
      *
-     * @param ctx {@link CreationalContext} for the {@link InterceptionProxyFactory} to create
-     * @param clazz class of the instance to apply proxies on
-     * @param <T> type of the instance to apply proxies on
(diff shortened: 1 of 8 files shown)


api/src/main/java/javax/enterprise/inject/spi/InterceptionProxyFactory.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+public interface InterceptionProxyFactory<T> {
+
+    /**
+     *
+     * Forces the creation of the proxy even if the targeted class has public final methods.
+     *
+     * @return self
+     */
+    InterceptionProxyFactory<T> ignoreFinalMethods();
```
manovotn: Using this will mean that you cannot call the final methods on the proxy (correct me if I am wrong). I think it is worth mentioning this at least in javadoc.

antoinesd: Done

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
```diff
@@ -606,4 +606,16 @@
...
+     *
+     * Create an {@link InterceptionProxyFactory} for the given {@link CreationalContext} and type.
+     *
+     * @param ctx {@link CreationalContext} for the {@link InterceptionProxyFactory} to create
+     * @param clazz class of the instance to apply proxies on
+     * @param <T> type of the instance to apply proxies on
+     * @return a new {@link InterceptionProxyFactory} to create proxies for an instance of T
+     * @since 2.0
+     */
+    <T> InterceptionProxyFactory<T> createInterceptionFactory(CreationalContext<T> ctx, Class<T> clazz);
```
manovotn: If anyone can refresh my mind, I know we talked about built-in beans, but what was the reason behind a BM method? I would guess Extensions, is that right?

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1580,3 +1594,35 @@ CDI 2.0 introduced the following Configurators interface:
 * <<observer_method_configurator>> for `ObserverMethod` configuration
 
 The container must provide implementation for all these configurators and make them available in matching container lifecycle events as defined in <<init_events>>.
+
+[[interception_proxy_factory]]
+
+=== Apply interceptor and decorators programmatically
+
+Contextual instances of bean having interceptors binding or defined decorators are provided by the container as a proxy.
```
mkouba: The term `proxy` is undefined from the spec point of view. The spec defines `client proxy` but it's not related to interceptors and decorators. I believe the wording should follow section `7.2. Container invocations and interception` and use terms like `contextual reference` and  `business method invocation`.

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1580,3 +1594,35 @@ CDI 2.0 introduced the following Configurators interface:
...
+This method will bypass rules defined in <<unproxyable>>
+Final methods on T won't be available on the produced proxy.
+* `configure()` returns an `AnnotatedTypeConfigurator` (as defined in <<annotated_type_configurator>>) initialized with the instance type to easily apply specific interceptor binding to apply when creating the proxy.
+The method always return the same `AnnotatedTypeConfigurator`
+* `createInterceptionProxy()` returns the proxy adding interceptors and decorators on the provided instance.
+The method can be only called once, subsequent calls will throw an exception.
+
+An `InterceptionProxyFactory` can be obtain be calling `BeanManager.createInterceptionFactory` as defined in <<bm_obtain_interception_proxy_factory>>
+
+The container must also provide an instance of `InterceptionProxyFactory` for every producer methods injecting this interface.
```
manovotn: If this is supposed to be about the built in beans, it sounds rather confusing. I suggest this instead:
`The container must also provide an InterceptionProxyFactory built-in bean. Therefore, another way to use this interface is to inject the built-in bean directly as a parameter of a producer method.`
Or something similar which will specifically mention the built in bean.

mkouba: Must provide a built-in bean with bean type `javax.enterprise.inject.spi.InterceptionProxyFactory`, qualifier `@Default` and _"injectable only in a producer method parameter"_ (needs rewording).

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
antoinesd: To get it for custom bean create method

spec/src/main/asciidoc/core/spi.asciidoc
antoinesd: ok.

spec/src/main/asciidoc/core/spi.asciidoc
antoinesd: On a second thought if we go in that direction (no mention of proxy) we should perhaps find another name for `InterceptionProxyFactory` interface, something like `ContextualReferenceFactory` or `InterceptedInstanceFactory`?
wdyt?

antoinesd:
This new version remove the term proxy from Javadoc and doc to mention enhancement of instance by making method invocation business method invocation as defined in section 7.2.
This new interface definitely need a new name without proxy in it. My suggestions are:

* `InstanceEnhancer` (short but  not very clear)
* `BusinessMethodInvocationFactory` (more exact from spec pov, but is it clear from user pov?)
* `InterceptionFactory` (cleared from user pov and near our initial name)
* `InterceptionEnhancer`

Other suggestion are welcome.

manovotn:
I think we should specifically mention in the spec that this feature enables the interceptor with `@AroundInvoke` binding. 
It might not be even possible to implement this for other bindings like `@AroundConstruct`.

antoinesd:
Good point @manovotn. Adding it with consistent rephrasing on ignoreFinals() with #328

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1580,3 +1593,37 @@ CDI 2.0 introduced the following Configurators interface:
...
+}
+----
+
+* `ignoreFinalMethods()` `ignoreFinalMethods()` Instructs the container to ignore all non-static, final methods with public, protected or default visibility declared on any bean type of the specific bean during validation of injection points that require proxyable bean type.
+These method should never be invoked upon bean instances. Otherwise, unpredictable behavior results.
+ It will bypass standard rules defined in <<unproxyable>>.
+* `configure()` returns an `AnnotatedTypeConfigurator` (as defined in <<annotated_type_configurator>>) initialized with the instance type to easily apply specific interceptor binding to apply when enhancing the instance.
+The method always return the same `AnnotatedTypeConfigurator`
+* `createInterceptedInstance()` returns an enhanced version of the instance for which each method invocations will be a business method invocation.
+The method can be only called once, subsequent calls will throw an exception.
```
tremes: Sorry I didn't have enough time to go through this PR. This sentence is too general.

PR #317, opened 2016-10-19
CDI-592 Add overloader ObserverMethod.notify()

mkouba:
https://issues.jboss.org/browse/CDI-592

I'm not 100% sure this is the best solution but I haven't found anything better.

This PR also introduces a new interface: `EventContext` which represents a context of a fired event (payload and metadata). I like this concept more than adding just another method parameter because it's easily extensible.

NOTE: If we go this way (introduce `EventContext`) we may also simplify `ObserverMethodConfigurator` which currently contains two versions of `notifyWith()`.

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/EventContext.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/EventContext.java
@@ -1,0 +1,42 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016 Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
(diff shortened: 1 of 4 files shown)


johnament:
This is concerning, since now you don't have to implement either method.

mkouba:
Yep, that's why I added _"If the given `ObserverMethod` does not override either `ObserverMethod.notify(T)` or `ObserverMethod.notify(EventContext<T>)`, the container automatically detects the problem and treats it as a definition error."_. I agree it's not nice but as I said I haven't found anything better.

johnament:
@mkouba and how would the container detect that?

mkouba:
@johnament `ObserverMethod.getClass().getMethods()` and `java.lang.reflect.Method.isDefault()` or something like that should work.

antoinesd:
LGTM.

PR #318, opened 2016-10-19
CDI-500 Clarify @Intercepted bean metadata injection for EE components

mkouba:


antoinesd:
LGTM

spec/src/main/asciidoc/javaee/injectionandresolution_ee.asciidoc
```diff
@@ -94,3 +94,7 @@ The container is also required to ensure that:
 ==== Destruction of dependent objects in Java EE
 
 When running in Java EE, the container must extend the rules defined for managed beans in <<dependent_objects_destruction>> to any other Java EE component class supporting injection and perform destruction after the servlet `destroy()` method is called.
+
+==== Bean metadata in Java EE
+
+Interceptor and decorator instances associated with Java EE components that are not considered beans (as defined by this specification) cannot obtain information about the beans they intercept and decorate (as defined in <<bean_metadata>>) and thus `null` is injected into relevant injection points.
```
tremes: I think we should remove `decorator` here because it would anticipate that decoration of EE components works and is supported which is not true AFAIK. WDYT @antoinesd ?

PR #320, opened 2016-10-21
CDI-627 also include alternative stereotypes in the check

struberg:
This version also includes @Alternative @Stereotypes. Txs to @tremes for the catch

### DIFF
--- a/spec/src/main/asciidoc/core/injectionandresolution.asciidoc
+++ b/spec/src/main/asciidoc/core/injectionandresolution.asciidoc
@@ -75,7 +75,8 @@ An alternative is selected for the bean archive if either:
 
 For each child `<class>` element the container verifies that either:
 
-* a class `T` exists which is annotated with `@Alternative` or contains a producer field or producer method which is annotated with `@Alternative`, or
+* a class `T` exists which is annotated with `@Alternative` or an <<alternative_stereotype,@Alternative @Stereotype>>, or
+* a class `T` exists which contains a producer field or producer method which is annotated with `@Alternative` or an <<alternative_stereotype,@Alternative @Stereotype>> , or
 * an alternative `Bean<T>` exists for the bean class.
 
 Otherwise the container automatically detects the problem and treats it as a deployment problem.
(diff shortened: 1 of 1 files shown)


mkouba:
Looks good but I would change the wording a little bit.

1. a class `T` =>  a bean class; it shouldn't be any class and `T` is imho confusing here
2. class which contains => class which declares
3. `@Alternative @Stereotype` => `@Alternative` stereotype; this wording is already used above and makes sense
4. an alternative `Bean<T>` => an alternative bean; `BeanAttributes.isAlternative()` defines what it means
5.  for the bean class => whose bean class has the specified name

In other words:

* a **bean class exists** which is annotated with `@Alternative` or an `@Alternative` stereotype, or
* a **bean class exists** which declares a producer field or a producer method which is annotated with `@Alternative` or an `@Alternative` stereotype , or
* an **alternative bean exists** whose bean class has the specified name

tremes:
+1 for @mkouba wording:-)

manovotn:
I don't think @mkouba solution is correct. As far as I recall, we intentionally avoided using **`bean class`** and instead went for just **`class`**. The reason is that (correct me if I am wrong) spec doesn't state what *bean class* actually is (even tough it extensively uses the term). Plus. if you actually veto the alternative `bean class`, you are left with plain class.

mkouba:
I see. Well, the phrase *bean class* is also used in [CDI 1.0 - Declaring selected alternatives for a bean archive](http://docs.jboss.org/cdi/spec/1.0/html/injectionelresolution.html#d0e3098). So the original 1.0 wording was not correct either? IMHO the bean class is defined as the result of `javax.enterprise.inject.spi.Bean.getBeanClass()` aka the _bean implementation class_.

manovotn:
Fair enough, though I don't claim it was incorrect. I think we only meant to stick to mere `class` for the sake of clarity.

struberg:
hi martin! 
ad 1: I fear we cannot use 'bean class' as it would again be confusing because sometimes they never become Beans.
ad 2: fine for me, will change
ad 3: also fine
ad 4: no, would stick with Bean. BeanAttributes is only an intermediate result
ad 5: puh, need to dig deeper

mkouba:
> ad 4: no, would stick with Bean. BeanAttributes is only an intermediate result

I don't think we should use `Bean<T>` as we don't use it anywhere in the spec, _alternative bean_ should be enough IMHO. But let's wait what others think.

Emily-Jiang:
LGTM

antoinesd:
Closing the PR. It's included in #342

antoinesd:
Closing this PR in favour of #342

PR #321, opened 2016-10-26
CDI-640 mention IllegalStateException in SeContainer methods javadoc.

tremes:


antoinesd:
LGTM

api/src/main/java/javax/enterprise/inject/se/SeContainer.java
```diff
@@ -43,13 +43,15 @@
      * Check if the container is running or was shut down
      *
      * @return true if called before container shutdown
+     * @throws IllegalStateException if the container is already shutdown
```
mkouba: @tremes @antoinesd It should be specified on `close()` method...

tremes: Yes this is my mistake. I overlooked this. Will fix.

PR #328, opened 2016-11-09
CDI-527 allow proxying of classes with final methds

antoinesd:


api/src/main/java/javax/enterprise/inject/spi/ProcessBeanAttributes.java
```diff
@@ -92,4 +92,16 @@
      * @throws IllegalStateException if called outside of the observer method invocation
      */
     public void veto();
+
+
+    /**
+     * Forces bean creation even if the bean class has non static and non private final methods.
```
mkouba: This wording is imho inaccurate. First of all, the container must only validate that a bean type is proxyable if an injection point resolves to a bean that requires a client proxy or has interceptor or has decorator. See also 3.15. Unproxyable bean types.

Also `non-static, final methods with public, protected or default visibility` would be probably better than ` non static and non private`.

api/src/main/java/javax/enterprise/inject/spi/ProcessBeanAttributes.java
```diff
@@ -92,4 +92,16 @@
...
+     *
+     * Calling this method will bypass standard rules for unproxyable bean types (section 3.11 of the spec)
+     *
+     * Non static and non private final methods on Bean class won't be available on bean instances.
+     *
+     * @since 2.0
+     */
```
mkouba: Also throws IllegalStateException if called outside of the observer method invocation.

api/src/main/java/javax/enterprise/inject/spi/ProcessBeanAttributes.java
mkouba: Something like: _"Instructs the container to ignore all non-static, final methods with public, protected or default visibility declared on any bean type of the specific bean during validation of injection points that require proxyable bean type."_ I know it's verbose but it's a spec anyway...

api/src/main/java/javax/enterprise/inject/spi/ProcessBeanAttributes.java
```diff
@@ -95,12 +95,13 @@
 
 
     /**
-     * Forces bean creation even if the bean class has non static and non private final methods.
+     * Instructs the container to ignore all non-static, final methods with public, protected or default visibility declared on any bean type of the specific bean during validation of injection points that require proxyable bean type.
+     * These methods shouldn't be invoked in the bean instances.
```
manovotn: `in` -> `on`

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1396,8 +1396,9 @@ public interface ProcessBeanAttributes<T> {
 The method always returns the same `BeanAttributesConfigurator`.
 * `addDefinitionError()` registers a definition error with the container, causing the container to abort deployment after bean discovery is complete.
 * `veto()` forces the container to ignore the bean.
-* `ignoreFinalMethods()` forces bean creation even if the bean class has non static and non private final methods.
-It will bypass standard rules defined in <<unproxyable>>
+* `ignoreFinalMethods()` Instructs the container to ignore all non-static, final methods with public, protected or default visibility declared on any bean type of the specific bean during validation of injection points that require proxyable bean type.
+These methods shouldn't be invoked in the bean instances
```
manovotn: Again, `in` -> `on`

manovotn:
Looks good now, +1

api/src/main/java/javax/enterprise/inject/spi/ProcessBeanAttributes.java
```diff
@@ -92,4 +92,18 @@
...
      */
     public void veto();
+
+
+    /**
+     * Instructs the container to ignore all non-static, final methods with public, protected or default visibility declared on
+     * any bean type of the specific bean during validation of injection points that require proxyable bean type.
+     * These methods shouldn't be invoked on the bean instances.
+     *
+     * Calling this method will bypass standard rules for unproxyable bean types (section 3.11 of the spec)
```
mkouba: I would leave out this sentence. Also if referencing a spec section it might be better to use `<a>` hyperlink so that it's easier to find/refactor links if needed.

manovotn: I disagree, I would keep it as a kind of a warning sign.

mkouba: For me this sentence attempts to express the same as the first sentence above. I think it would be better to add the reference to the spec at the end of the first sentence, e.g. `... of injection points that require proxyable bean type, as defined in section 3.11 Unproxyable bean types`.

A better "warning sign" would be to move `These methods shouldn't be invoked on the bean instances.` to a separate paragraph and reword to something like:
`These method should never be invoked upon bean instances. Otherwise, unpredictable behavior results.`

manovotn: I am fine with any wording as long as we keep the "warning sign" up so we avoid people misusing/overusing this.

PR #330, opened 2016-11-11
CDI-471 Introduce SPI for retrieving multiple annotations of the same…

johnament:
… type, deprecating the old one.  Updated spec docs for qualifiers to reflect repeatability and spec docs for annotated for new method.

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/Annotated.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/Annotated.java
@@ -31,6 +31,7 @@ import java.util.Set;
  * @author Gavin King
  * @author Pete Muir
  * @author Clint Popetz
+ * @author John D. Ament
  * 
  */
 public interface Annotated {
@@ -42,7 +43,7 @@ public interface Annotated {
      * 
      * @return the type of the annotated program element
      */
-    public Type getBaseType();
+    Type getBaseType();
 
     /**
      * <p>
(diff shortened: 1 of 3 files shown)


johnament:
One thing strikes me about this - is it possible people implemented their own annotated?  I can't think of a reason right now.

spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -382,6 +382,67 @@ PaymentProcessor getPaymentProcessor(@Synchronous PaymentProcessor sync,
 }
 ----
 
+[[repeating_qualifiers]]
```
tremes: Great.:-) Appreciate.

api/src/main/java/javax/enterprise/inject/spi/Annotated.java
```diff
@@ -51,18 +52,34 @@
...
      * @return a set of all types to which the base type should be considered assignable
      */
-    public Set<Type> getTypeClosure();
+    Set<Type> getTypeClosure();
 
     /**
      * <p>
      * Get program element annotation of a certain annotation type.
+     *
+     * Since 2.0, calling this method is considered deprecated.  To support {@see Repeatable} annotations, call {@see getAnnotation}
```
mkouba: I would use the "settled" wording (see for example `BeanManager.createInjectionTarget()`):
> This method is deprecated from CDI 2.0 and {@link #getAnnotations(Class)} should be used instead.

Also I don't think it's possible to use `{@see}` inline.

api/src/main/java/javax/enterprise/inject/spi/Annotated.java
```diff
@@ -51,18 +52,34 @@
...
      * </p>
      * 
      * @param <T> the type of the annotation
      * @param annotationType the class of the annotation type
      * @return the program element annotation of the given annotation type, or a null value
```
mkouba: _the **first**  program element annotation found, or a null value_?

api/src/main/java/javax/enterprise/inject/spi/Annotated.java
```diff
@@ -51,18 +52,34 @@
...
      */
-    public <T extends Annotation> T getAnnotation(Class<T> annotationType);
+    @Deprecated
```
mkouba: AFAIK we never use `@Deprecated` in CDI API.

api/src/main/java/javax/enterprise/inject/spi/Annotated.java
johnament: Ok, I'll replace the `@Deprecated` with this as well.  Link was what I was looking for, but couldn't remember.

johnament:
@mkouba all fixed, I hope!

mkouba:
@johnament Looks good!

PR #331, opened 2016-11-14
CDI-645 update Weld parent version which includes latest javadoc plug…

tremes:
…in version.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>org.jboss.weld</groupId>
 		<artifactId>weld-parent</artifactId>
-		<version>31</version>
+		<version>36</version>
 		<relativePath />
 	</parent>
 


api/pom.xml
```diff
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>org.jboss.weld</groupId>
 		<artifactId>weld-parent</artifactId>
-		<version>31</version>
+		<version>36</version>
```
johnament: Unrelated to this PR, it strikes me that the CDI spec's API is dependent on the impl.  I feel like it should be the other way around.

tremes: Well yeah. We should discuss it somewhere else then, but I don't consider it so problematic. `weld-parent` is just pom.xml and no impl in fact.

antoinesd: I agree with @johnament, I'd like to fix that before the release of 2.0 if possible

manovotn: Weld itself (as a CDI impl) depends on `weld-parent`. Because `weld-parent` is basically just a declaration of plugin versions and nothing more. So I don't really see it as a problem.

PR #332, opened 2016-11-15
CDI-481 Introduce AnnotatedParameter.getJavaParameter()

mkouba:
@cdi-spec/eg please review.


antoinesd:
Looks good but why putting the code in the API and not in the implementation ?

PR #333, opened 2016-11-15
Apache License is abbreviated as ALv2 not ASL2

struberg:


johnament:
I'd rather the full name, like https://github.com/cdi-spec/cdi/blob/master/api/pom.xml#L22

PR #334, opened 2016-11-21
CDI-633 introduce BeanManager.getEvent()

antoinesd:


api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
```diff
@@ -618,4 +623,15 @@
...
+
+    /**
+     *
+     * Returns an instance of {@link Event} whose event type is Object and without qualifier.
+     * It allows typesafe synchronous or asynchronous event firing without injection of {@link Event} built-in bean requirement.
+     *
+     * @return a new {@link Event} object whose event type is Object and without qualifier
+     * @since 2.0
+     */
+    Event<Object> getEvent();
```
mkouba: I think we might consider adding something like `selectEvent(Class<T> eventType, Annotation... qualifiers)` instead, because users will mostly call something like `getEvent().select(Foo.class).fire(new Foo())` anyway...

mkouba: Or even `getEvent(Class<T> eventType, Annotation... qualifiers)`...

antoinesd: but then we need to also add `selectEvent(TypeLiteral<U> subtype, Annotation... qualifiers)`

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -401,6 +401,20 @@ public void validate(InjectionPoint injectionPoint);
 
 ==== Firing an event
 
+The method `BeanManager.getEvent()` returns an an instance of `Event<Object>` and without qualifier.
```
mkouba: Given that this method should simulate `@Inject Event<Object>` I think we should use something like:

The method `BeanManager.getEvent()` returns an instance of `Event` with _specified type_ `java.lang.Object` and _specified qualifier_ `@Default`.

See also 3.8. "The default qualifier at injection points". Terms specified type and specified qualifiers are used in "10.2.3. The Event interface".

Hm, but it might be better to use the  _specified type_ `java.lang.Object` and _specified qualifier_ `@Any`. WDYT?

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
mkouba: Good point... -1 for adding multiple methods :-(

mkouba: It's a pity we don't have an `EventBuilder` ;-)

E.g. `BeanManager.event().type(Foo.class).qualifiers(Foo.INSTANCE).fire(new Foo())`

spec/src/main/asciidoc/core/spi.asciidoc
antoinesd: '@Inject Event<Object>' doesn't have `@Default` qualifier since firing it it won't trigger `@Observes @Default Object ` unless you explicit `select()` the `@Default` qualifier before `fire()`.

So "specified qualifier `@Any`" is better and "with no qualifier" is nearer to the truth.

mkouba: > '@Inject Event' doesn't have @Default qualifier since firing it it won't trigger @Observes @Default Object unless you explicit select() the @Default qualifier before fire().

Then it's a bug. See also 3.12. The default qualifier at injection points.

antoinesd: Sorry, it's not exactly that: section 3.12 is applied to Event IP but  qualifiers at `Event<>` injection point are "erased" and not added if you specify new qualifiers with the 'select()' method.

So the wording:

> The method `BeanManager.getEvent()` returns an instance of `Event` with _specified type_ `java.lang.Object` and _specified qualifier_ `@Default`. 

is probably the best one.

manovotn:
+1

mkouba:
Looks good.

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -401,6 +401,20 @@ public void validate(InjectionPoint injectionPoint);
...
+The method `BeanManager.getEvent()` returns an instance of `Event` with specified type `java.lang.Object` and specified qualifier `@Default`.
+
+[source, java]
+----
+Event<Object> getEvent();
+----
+
+The returned instance can be used like a standard `Event` as described in <<events>>.
+
+===== `fireEvent()` method
```
tremes: is this really intended as the new chapter? If so then it's missing id again:-)

antoinesd: Merged too fast. In fact I propose a more straightforward deprecation text in PR #351

PR #336, opened 2016-11-22
CDI-471 Clarified the behavios of Annotated and repeatable annotation…

johnament:
…s.  Provide a default implementation in case there are custom implementations out there.

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/Annotated.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/Annotated.java
@@ -18,8 +18,13 @@
 package javax.enterprise.inject.spi;
 
 import java.lang.annotation.Annotation;
+import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Type;
+import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.Set;
+
+import static java.util.Arrays.asList;
 
 /**
  * <p>
@@ -57,11 +62,10 @@ public interface Annotated {
     /**
      * <p>
      * Get program element annotation of a certain annotation type.
-     *</p>
      *
-     * <p>
-     * This method is deprecated from CDI 2.0 and {@link #getAnnotations(Class)} should be used instead.
-     * </p>
+     * The behavior of this method is intended to be the same behavior as {@link AnnotatedElement#getAnnotation(Class)},
+     * where repeatable annotations are not supported.
+     *</p>
      *
      * @param <T> the type of the annotation
      * @param annotationType the class of the annotation type
@@ -72,11 +76,14 @@ public interface Annotated {
     /**
      * <p>
      * Get program element annotations of a certain annotation type.
+     * <p>
+     * This method returns back all annotations, including repeatable annotations of this type.
      *
-     * This value returns back all annotations, including repeatable annotations of this type.
+     * The behavior of this method is intended to be the same behavior as {@link AnnotatedElement#getAnnotationsByType(Class)},
+     * where repeatable annotations are supported.
      * </p>
      *
-     * @param <T> the type of the annotation
+     * @param <T>            the type of the annotation
      * @param annotationType the class of the annotation type
      * @return the program element annotations of the given annotation type, or an empty collection
      */
@@ -94,8 +101,10 @@ public interface Annotated {
     /**
      * <p>
      * Determine if the program element has an annotation of a certain annotation type.
-     * <p>
-     * 
+     *
+     * The behavior of this method is similar to {@link AnnotatedElement#isAnnotationPresent(Class)} for the underlying
+     * program element.
+     *
      * @param annotationType the annotation type to check for
      * @return <tt>true</tt> if the program element has an annotation of the given annotation type, or <tt>false</tt> otherwise
      */
--- a/api/src/test/java/org/jboss/cdi/api/test/annotated/AbstractAnnotatedTest.java
+++ b/api/src/test/java/org/jboss/cdi/api/test/annotated/AbstractAnnotatedTest.java
@@ -1,0 +1,52 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.cdi.api.test.annotated;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import javax.enterprise.inject.spi.Annotated;
+import java.lang.annotation.Annotation;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import static java.util.Arrays.asList;
+
+public abstract class AbstractAnnotatedTest {
+
+    @Test public void shouldFindAnnotationsOnAmmptated() {
+        Annotated annotated = getAnnotated();
+
+        Set<Repeater> repeaters = annotated.getAnnotations(Repeater.class);
+
+        Assert.assertEquals(repeaters, new LinkedHashSet<>(asList(getAnnotationsByType(Repeater.class))));
+    }
+
+    @Test public void shouldReturnSingletonForNonRepeatableAnnotation() {
+        Annotated annotated = getAnnotated();
+
+        Set<Repeaters> repeaterses = annotated.getAnnotations(Repeaters.class);
+
+        Assert.assertEquals(repeaterses, Collections.singleton(getAnnotation(Repeaters.class)));
+    }
+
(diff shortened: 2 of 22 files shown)


api/src/main/java/javax/enterprise/inject/spi/Annotated.java
```diff
@@ -94,6 +113,8 @@
     /**
      * <p>
      * Determine if the program element has an annotation of a certain annotation type.
+     *
+     * If the <tt>annotationType</tt> is repeatable, then returns true if at least one instance is available
```
mkouba: I think we should follow the Reflection API here, i.e. ignore repeatable annotations and return `getAnnotation(annotationClass) != null`.

mkouba:
http://lists.jboss.org/pipermail/cdi-dev/2016-November/009294.html

api/src/main/java/javax/enterprise/inject/spi/Annotated.java
johnament: I think you and Martin are contradicting each other.  I've been testing, `getAnnotation(SomeRepeatableAnnotation) == null` is true in these cases.  That's why I had added this to the javadoc, its not necessarily consistent with the JVM's impl.

mkouba: Sorry, I don't understand. My point is that `Annotated.isAnnotationPresent()` should behave as `java.lang.reflect.AnnotatedElement.isAnnotationPresent()`, i.e. return `false` for repeatable annotation.

api/src/test/java/org/jboss/cdi/api/test/annotated/AnnotatedTest.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import static java.util.Arrays.asList;
+
+/**
+ * @author John D. Ament
+ */
+public class AnnotatedTest {
```
mkouba: This only tests `AnnotatedType`. Tests for method, param, constructor and field are missing. Pls read also this message: http://lists.jboss.org/pipermail/cdi-dev/2016-November/009294.html

api/src/main/java/javax/enterprise/inject/spi/Annotated.java
johnament: Right, and in my opinion we shouldn't necessarily follow what the JDK does.

api/src/test/java/org/jboss/cdi/api/test/annotated/AnnotatedTest.java
johnament: Actually, it tests `Annotated`.  I didn't understand that's what you meant in that email.  I'll add tests for each of the `Annotated*` interfaces.

mkouba: > Actually, it tests Annotated.

There's no Java program element for `Annotated`, only for extending interfaces like `AnnotatedType` and `AnnotatedParameter`.

> Though that kind of defeats the purpose of this being a default implementation.

Why?? The default impl must work for all extending interfaces in the API. Or extending interfaces (such as `AnnotatedMethod`) must provide their own default impls.

johnament: I'm inclined to believe that the correct solution is to:

- Leave `Annotated` with just the method declaration
- Add default implementations to `AnnotatedMember`, `AnnotatedParameter` and `AnnotatedType`.  The reason being is that the target of what is being scanned for annotations differs in each of those.  So the default impl in the root `Annotated` doesn't make much sense since there's no real target.

And then add tests for those three.

There's still an underlying problem, and this is where we're getting some discrepancies.  The current API is focused on working in a compatible way - however we have other areas that are not compatible.  So either `getAnnotations` and all related methods (`isAnnotationPresent` for instance) either should behave like the JDK or not like the JDK and that should be consistent in this class.  The current impl of `getAnnotations` is not consistent with the JDK.

mkouba: > Leave Annotated with just the method declaration...

Ok. If it's not easily doable for all the subclasses. 

> And then add tests for those three.

+10

> There's still an underlying problem...

I think it's ok to use the name `getAnnotations(Class)` instead of `getAnnotationsByType(Class)` (JDK). But the existing methods (`getAnnotations()`, `isAnnotationPresent()` and `getAnnotation(Class))` should not change the behavior (repeatable annotations are not supported now!) and follow JDK.

api/src/main/java/javax/enterprise/inject/spi/Annotated.java
antoinesd: Usually we follow JLS and JDK rule but there are exception. My POV here is there's more added value in @johnament approach than following how the JDK.

mkouba: I think this would NOT be backward compatible...

antoinesd: Could you give an example of incompatibility problem that could raise @mkouba ?

mkouba: Up to CDI 1.2 this method should return false if you pass a repeatable annotation...

api/src/test/java/org/jboss/cdi/api/test/annotated/AnnotatedTest.java
antoinesd: Tests are ok for me.
As said above, I agree with Martin, we should align behaviour on JDK.

johnament: We would need to add a method to check if a repeatable annotation is found as well.

antoinesd: Perhaps we could introduce a method with the same name  `isAnnotationPresent(boolean checkRepeatable)`

mkouba: What's wrong with `!Annotated.getAnnotations(FooRepeatable.class).isEmpty()`?

mkouba:
@johnament @antoinesd Guys, we really need this to be fixed ASAP so that we can release a new CDI API Alpha and align Weld afterwards.

api/src/test/java/org/jboss/cdi/api/test/annotated/AnnotatedTest.java
antoinesd: Only for consistency sake: what was wrong with `Annotated.getAnnotation(Foo.class) != null` in the first place.
This said, it's not critical.

antoinesd:
@johnament if you don't have time, feel free to give us right on the PR by checking "allow edits from maintainers" checkbox.

mkouba:
@johnament @antoinesd I've re-read the JLS and to be honest I don't understand why the [default Weld implementation](https://github.com/weld/api/blob/master/weld/src/main/java/org/jboss/weld/experimental/ExperimentalAnnotated.java#L50) should not be portable accross JVMs.

See also **9.7.5 Multiple Annotations Of The Same Type**:

> If a declaration context or type context has multiple annotations of a repeatable
annotation type T , then it is as if the context has no explicitly declared annotations
of type T and one implicitly declared annotation of the containing annotation type
of T .
The implicitly declared annotation is called the container annotation, and the
multiple annotations of type T which appeared in the context are called the base
annotations. The elements of the (array-typed) value element of the container
annotation are all the base annotations in the left-to-right order in which they
appeared in the context.

John which part of the Weld default impl is problematic?

johnament:
@mkouba You may want to read through section 9.6.3 to see where the full list of bugs are with your method.  http://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.6.3

The ones I see right off:

- Its not handling Container-Containers
- You're missing the check that the annotations of the container are actually the right annotations

IMHO the new proposed solution is better even if it doesn't handle all cases.  I can't think of a scenario where someone has implemented `AnnotatedCallable` or `AnnotatedMember` directly and its done something useful.

mkouba:
> * Its not handling Container-Containers

Good point! I wonder if there is a use-case for repeatable containing annotation though ;-)

Ok, the current defailt implementation looks good.

api/src/test/java/org/jboss/cdi/api/test/annotated/AbstractAnnotatedTest.java
```diff
@@ -0,0 +1,52 @@
...
+import java.lang.annotation.Annotation;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import static java.util.Arrays.asList;
+
+public abstract class AbstractAnnotatedTest {
+
+    @Test public void shouldFindAnnotationsOnAmmptated() {
```
manovotn: Typo(s) in method name

api/src/test/java/org/jboss/cdi/api/test/annotated/AbstractAnnotatedTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+        Set<Repeater> repeaters = annotated.getAnnotations(Repeater.class);
+
+        Assert.assertEquals(repeaters, new LinkedHashSet<>(asList(getAnnotationsByType(Repeater.class))));
+    }
+
+    @Test public void shouldReturnSingletonForNonRepeatableAnnotation() {
+        Annotated annotated = getAnnotated();
+
+        Set<Repeaters> repeaterses = annotated.getAnnotations(Repeaters.class);
```
manovotn: `repeaterses` - > `repeaters`

manovotn:
Looks good, +1

PR #339, opened 2016-11-23
CDI-651 Add convenient Instance.isResolvable()

mkouba:


manovotn:
+1 from me

tremes:
Looks OK

mkouba:
Actually, I think we should provide a default impl like:

    boolean isResolvable() {
        return !isUnsatisfied() && !isAmbiguous();
    }

I will add one more commit to this PR.

PR #341, opened 2016-11-24
CDI-642 Allow defining interceptor binding or qualifier with an AnnotatedTypeConfigurator

antoinesd:
CDI-642 Allow defining interceptor binding or qualifier with an AnnotatedTypeConfigurator


antoinesd:
Spec part not written yet: API should be discussed first.

mkouba:
Hm, shouldn't the methods accept `Class<T>` arguments? Similarly as `javax.enterprise.inject.spi.BeforeBeanDiscovery.addAnnotatedType(Class<T>, String)`?

I also don't like the names. Maybe we could name the methods `configureQualifier(Class)` and `configureInterceptorBinding(Class)` since we're basically configuring the original definitions?

antoinesd:
Agree on both points. changing the PR in that direction

mkouba:
Looks good. We should merge it ASAP so that we can identify possible implementation problems in time ;-)

PR #343, opened 2016-11-24
CDI-643 Provide a way to easily configure injection point of an InjectionTarget

antoinesd:
CDI-643 Provide a way to easily configure injection point of an InjectionTarget

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/InjectionTargetFactory.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/InjectionTargetFactory.java
@@ -17,6 +17,8 @@
 
 package javax.enterprise.inject.spi;
 
+import javax.enterprise.inject.spi.configurator.AnnotatedTypeConfigurator;
+
 /**
  * <p>
  * An {@link InjectionTargetFactory} can create an {@link InjectionTarget} for a given bean.
@@ -43,6 +45,7 @@ package javax.enterprise.inject.spi;
  * </pre>
  * 
  * @author Pete Muir
+ * @author Antoine Sabot-Durand
  * @since 1.1
  * 
  */
@@ -55,5 +58,20 @@ public interface InjectionTargetFactory<T> {
      * @return the injection target
      */
     public InjectionTarget<T> createInjectionTarget(Bean<T> bean);
+
+    /**
+     *
+     * Returns an {@link AnnotatedTypeConfigurator} to add or remove injection point to the {@link InjectionTarget} to create.
+     * {@link #createInjectionTarget(Bean)} will use the {@link AnnotatedTypeConfigurator} to define injection points in the returned
+     * {@link InjectionTarget}
+     *
+     * Each call returns the same AnnotatedTypeConfigurator.
+     *
+     * @return an {@link AnnotatedTypeConfigurator} to configure injection points
+     * @since 2.0
+     */
+    default AnnotatedTypeConfigurator<T> configure() {
+        throw new UnsupportedOperationException("Configuration not supported here");
+    }
 
 }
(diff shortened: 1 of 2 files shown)


antoinesd:
Spec part not written yet. API should be discussed first.

api/src/main/java/javax/enterprise/inject/spi/InjectionTargetFactory.java
```diff
@@ -56,4 +59,19 @@
...
+     * {@link #createInjectionTarget(Bean)} will use the {@link AnnotatedTypeConfigurator} to define injection points in the returned
+     * {@link InjectionTarget}
+     *
+     * Each call returns the same AnnotatedTypeConfigurator.
+     *
+     * @return an {@link AnnotatedTypeConfigurator} to configure injection points
+     * @since 2.0
+     */
+    default AnnotatedTypeConfigurator<T> configure() {
+        return null;
```
manovotn: I don't think this is good. In fact, returning `null` in default impl seems rather nasty.
I can understand that you meant to provide a default method, however this class is not implemented by end user but rather by the CDI impl itself, therefore *not having* it `default` wouldn't cause so much harm IMO.

mkouba: I also think that the default impl is not necessary here.

antoinesd: I agree that it's not nice, but look at the javadoc of the interface. It shows creation of an anonymous `InjectionTargetFactory` to wrap the one returned by the `BeanManager`. This code will break if user did that and update to 2.0...

mkouba: Ok, then throwing `UnsupportedOperationException` might be more appropriate...

manovotn:
LGTM, +1

mkouba:
Hm, previously `getInjectionTargetFactory()` returned a fully configured (possibly immutable) instance. If we allow to configure the annotated type, then we must specify when exactly is the factory configuration "finished" (effectively immutable). I.e. when `createInjectionTarget()` is called for the first time. And possibly throw `IllegalStateException` for subsequent ` configure()` calls?

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -579,11 +579,17 @@ public <T> InjectionTargetFactory<T> getInjectionTargetFactory(AnnotatedType<T>
 public interface InjectionTargetFactory<T> {
 
     public InjectionTarget<T> createInjectionTarget(Bean<T> bean);
+    AnnotatedTypeConfigurator<T> configure();
```
mkouba: **public** AnnotatedTypeConfigurator<T> configure();

mkouba:
Well, it might make sense to have the same text in javadoc and spec. And to be honest, I had to re-read the javadoc several times to understand it. But otherwise looks good ;-)

antoinesd:
@mkouba, I aligned first sentence of javadoc on spec text..

johnament:
This change doesn't make much sense to me.  The use case is that I can implement my own `InjectionTarget` and use it here.  This code has me potentially implementing `configure` as well here.  Are you expecting end users to implement `AnnotatedTypeConfigurator` here?

antoinesd:
@johnament, nobody implements `InjectionTargetFactory` from scratch. You do that to decorate the one retuned by the `BeanManager` as shown in the Javadoc of the class.
As `BeanManager.createInjectionTarget()` is deprecated a common use is to use the factory to make a non contextual object an injection target.
Adding this configuration allows easy addition of injection points in the `InjectionTarget` to create.

manovotn:
I am fine with this PR, do we have any objections or can we merge it?

antoinesd:
I've created a very basic example to use this feature: https://gist.github.com/antoinesd/c51cdbfa4348abaeacc2024551f86e8d

PR #347, opened 2016-11-28
CDI-646 Annotated configurators

mkouba:


manovotn:
After convincing Martin to keep `removeAll` methods, this has my +1 :)

api/src/main/java/javax/enterprise/inject/spi/configurator/AnnotatedTypeConfigurator.java
```diff
@@ -53,33 +61,32 @@
...
      *
-     * @param annotationType annotation class to remove
+     * @param annotationType the annotation type
      * @return self
      */
     AnnotatedTypeConfigurator<T> remove(Class<? extends Annotation> annotationType);
 
     /**
-     * Remove all annotations from the type.
+     * Remove all the annotations from the type.
```
Emily-Jiang: all the - not grammatically correct. Revert change is better.

tremes:
Yes using predicates is nice idea. +1

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -847,13 +845,12 @@ public interface AnnotatedTypeConfigurator<T> {
 
 * `getAnnotated()` returns the original `AnnotatedType` with which this configurator was initialized
 * `add()` adds an annotation to the configured `AnnotatedType`
-* `remove()` methods remove all annotation from the configured `AnnotatedType`
-* `removeAll()` remove all existing annotation on the configured `AnnotatedType`
```
manovotn: Just noticed this - since we have kept the `removeAll` method, this line should stay, right?

mkouba: Yep, thanks for noticing.

PR #348, opened 2016-11-30
CDI-420 define trim also in beans xsd.

tremes:
Attempt to define `trim` in beans_2_0.xsd. This should be just an empty element if I am not wrong.

### DIFF
--- a/api/src/main/resources/beans_2_0.xsd
+++ b/api/src/main/resources/beans_2_0.xsd
@@ -77,13 +77,13 @@
       </xs:documentation>
     </xs:annotation>
     <xs:complexType>
-      <xs:choice minOccurs="0" maxOccurs="unbounded">
-        <xs:element ref="javaee:interceptors"/>
-        <xs:element ref="javaee:decorators"/>
-        <xs:element ref="javaee:alternatives"/>
-        <xs:element ref="javaee:scan"/>
-        <xs:any namespace="##other" processContents="lax"/>
-      </xs:choice>
+      <xs:all >
+        <xs:element ref="javaee:interceptors" minOccurs="0"/>
+        <xs:element ref="javaee:decorators" minOccurs="0" />
+        <xs:element ref="javaee:alternatives" minOccurs="0" />
+        <xs:element ref="javaee:scan" minOccurs="0" />
+        <xs:element ref="javaee:trim" minOccurs="0"/>
+      </xs:all>
       <xs:attribute name="version" default="2.0">
         <xs:annotation>
           <xs:documentation>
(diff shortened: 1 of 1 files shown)


api/src/main/resources/beans_2_0.xsd
```diff
@@ -353,5 +354,14 @@
       </xs:choice>
     </xs:complexType>
   </xs:element>
+  <xs:element name="trim" type="xs:string" fixed="">
```
manovotn: I was just thinking if we could/should add some restrictive attributes here. Such as limiting min/max occurrences?

antoinesd: +1 to force occurrence to 1. Otherwise looks good.

tremes: Yes I was thinking about this as well. So then it makes sense to change it as above. I don't think it makes sense to allow multiple occurences of e.g `<scan>`.  

I don't know xsd much, but now each of these elements could be specified only once or do somebody have better idea how to define this???

antoinesd: My science in xsd is not very high, but my friend Google told me that: 

> The all element specifies that the child elements can appear in any order and that each child element can occur zero or one time.

(See ref [here](http://www.w3schools.com/xml/el_all.asp))

So `minOccurs` attributes are useless here IMO.

antoinesd: You're right @tremes. Pr is good for me then

tremes:
My concern is that with this approach we could potentially break beans.xml files e.g with multiple `<alternatives>` elements, because I guess this was not validated before but need to check.

antoinesd:
@tremes it looks like a corner case and can't happen without user modification. 
It's not a backward compatibility issue IMO: if the user has this (wrong but valid in 1.1) pattern and doesn't touch the beans.xml file, it will continue to be validated against 1.1 xsd and will continue to work (if Weld 2.X accept it today)

tremes:
@antoinesd ok you are probably right. Thanks

manovotn:
+1 from me

Emily-Jiang:
I cannot see my comments were taken into consideration: 


Bean defining annotations include the any scope annotations. Therefore, it makes more sense to delete "or any scope annoation".
Any comments?

mkouba:
@Emily-Jiang Bean defining annotations only contain normal scopes and `@Dependent`. We'd like to support `@Singleton` and other pseudo-scopes as well.

Emily-Jiang:
@mkouba It is confusing to add the support for @Singleton and other pseudo-scopes as they are not beans. What user case do you have in mind?

mkouba:
Actually, `@Singleton` is supported by both OWB and Weld (although it's not clearly defined in the spec).

Emily-Jiang:
Then we should push to get @Singleton added to CDI-2.0 spec as bean defining annotations. -1 on adding other pseudo-scopes though.

tremes:
Problem is that @Singleton annotation is not from CDI API but from javax inject API. Pseudo scopes are necessary IMO. 
Btw I don't understand why we are discussing here when this PR doesn't change or update this wording and is closed.

antoinesd:
@Emily-Jiang take a look at https://issues.jboss.org/browse/CDI-377 resolved in CDI 1.2

Emily-Jiang:
@antoinesd thanks. I am not comfortable that we support these annotations as beans without being stated in the spec. This is another discussion.
@tremes We still need to discuss the remaining issues even if it is merged.

PR #349, opened 2016-11-30
CDI-653 introduce BeanManager#getInstance()

antoinesd:


api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
```diff
@@ -618,4 +620,18 @@
...
+     * Obtains an {@link Instance} giving access to beans instances
+     *
+     * Instances of dependent scoped beans obtained with this Instance should be explicitly release with {@link Instance#destroy(Object)}
+     *
+     * @return an {@link Instance} object to request beans instances
+     * @throws IllegalStateException if called during application initialization, before the {@link AfterDeploymentValidation}
+     *         event is fired.
+     * @since 2.0
+     */
+    Instance<Object> getInstance();
```
manovotn: This seems way too generic to me, why don't we go a step further and do something like this?
`Instance<T> getInstances(Class<T> type);`

In CDI-653 you yourself noted that was the intended use case anyway.

manovotn: Apart from that, originally proposed method name used plural form, 
e.g.`getInstance` -> `getInstances`

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
```diff
@@ -618,4 +620,18 @@
      */
     <T> InterceptionFactory<T> createInterceptionFactory(CreationalContext<T> ctx, Class<T> clazz);
 
+
+    /**
+     *
+     * Obtains an {@link Instance} giving access to beans instances
+     *
+     * Instances of dependent scoped beans obtained with this Instance should be explicitly release with {@link Instance#destroy(Object)}
+     *
```
mkouba: We should also mention that only beans which are eligible for injection into the class into which the `BeanManager` was injected are accessible from the returned `Instance`. Something similar to `javax.enterprise.inject.spi.BeanManager.getBeans(Type, Annotation...)` javadoc.

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -674,6 +674,20 @@ The method `BeanManager.getInterceptionFactory()` returns an `InterceptionFactor
...
+==== Obtain an `Instance`
+
+The method `BeanManager.getInstance()` returns an `Instance<Object>` to request bean instances programmatically as described in <<dynamic_lookup>>.
+
+[source, java]
+----
+Instance<Object> getInstance();
+----
+
+Instances of dependent scoped beans obtained with this `Instance` object should be explicitly released by calling `Instance.release()` method.
```
mkouba: There is no `Instance.release()` method yet. And the fix version for CDI-589 is 2.1.

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
antoinesd: ok, I wasn't sure of the mention. Thx

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
antoinesd: I'm not sure to understand. You mean the whole approach?

manovotn: Just saying I don't really see the value of this issue/PR. It's only a convenient method and even as such requires you to take several steps to actually achieve the goal (get the contextual instance). And for this little (if any) gain we add yet another method to already bloated BM. But that's probably just my skepticism.

astefanutti: IMHO, conveniency adds value from the developer experience standpoint. I'd rather prefer writing:

    manager.getInstance().select(type).get();

instead of:

    Bean<?> bean = manager.resolve(manager.getBeans(type, Any.Literal.INSTANCE));
    manager.getReference(bean, type, manager.createCreationalContext(bean));

or to have to create my util class in each extension that I develop, or to depends on DeltaSpike just for that purpose.

I feel a bit bad to be honest each time I present CDI extensions to people and have to find some convoluted way to retrieve contextual instances 😉.

antoinesd: @manovotn I understand your pov, but most of our work in CDI 2.0 has been to add convenient way to do things that were already doable (i.e. configurators api). I think these added value for users are very important because they will bring down the ticket price to write extension and advance CDI code. It will pay in the end.
Regarding the size of BeanManager I agree that it's not the best, but either we have multiple entry point to CDI SPI, either we have something like BM... Remember that JDK8 developers dared to introduce classes like `CompletableFuture` with nearly 80 methods ;)

manovotn:
Just to dispel confusion (given my previous comments), from my perspective, this PR still has +1

tremes:
I think it looks ok.

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
```diff
@@ -634,4 +636,25 @@
...
+     * this method will only return instances of beans discovered by the container before the {@link AfterBeanDiscovery} event is fired.
+     *
+     * Instances of dependent scoped beans obtained with this Instance should be explicitly destroyed with {@link Instance#destroy(Object)}
+     *
+     * @return an {@link Instance} object to request beans instances
+     * @throws IllegalStateException if called during application initialization, before the {@link AfterBeanDiscovery}
+     *         event is fired.
+     * @since 2.0
+     */
+    Instance<Object> getInstance();
```
johnament: I'd prefer to see `instance()` as a method.  the use of `getInstance()` makes it look like singletons.

antoinesd: np with that.
If we go that way we should also rename `BeanManager.getEvent()` in `BeanManager.event()` 

WDYT @cdi-spec/eg ?

manovotn: I would stay with `get` prefix.

> getInstance() makes it look like singletons.

Not to me. If you work with `Instace<T>`, you have to be aware that it can contain 0-n beans.

mkouba: John is right that [singleton pattern](https://en.wikipedia.org/wiki/Singleton_pattern) is usually using static `getInstance()` method. For me both names are ok if the javadoc/spec text is clear.

manovotn: I can see your point. But still, to me the association of singleton pattern with `BM.getInstance` seems a bit far fetched.

johnament: You have to think about it from a typical user's POV and not your own POV as a developer of the framework.

I don't have a strong opinion over not using `getInstance` but it does read incorrectly to me.

Emily-Jiang: I'm ok with instance() and event().

mkouba: -1 for two methods. The "root" `Instance` might be reused in theory so there is no real gain in terms of resources. Also `BeanManager` has even now too many methods.

+0 for `createInstance()`.

Hm, just an idea - what if we make `BeanManager implements Instance<Object>` similarly as we do for `CDI` and `SeContainer`?

antoinesd: > -1 for two methods. The "root" `Instance` might be reused in theory so there is no real gain in terms of resources. Also `BeanManager` has even now too many methods.

Same -1 for me, we should also add version with `TypeLeral` and `AnnotationLiteral`. The less method we add to `BeanManager`, the better we are.
 
> +0 for `createInstance()`.

I don't have strong opinion on method name, but `createInstance()` doesn't feel right: we are not creating a SPI element just asking for an easy way to look for bean instances.

> Hm, just an idea - what if we make `BeanManager implements Instance<Object>` similarly as we do for `CDI` and `SeContainer`?

I love this idea, it'll bring some consistency and avoid bloating the `BeanManager`.

struberg:
You shall not reuse any Instance! 
This internally has a Map with all the CreationalContext etc. Would create mem leaks!



--------------------------------------------
On Tue, 6/12/16, Martin Kouba <notifications@github.com> wrote:

 Subject: Re: [cdi-spec/cdi] CDI-653 introduce BeanManager#getInstance() (#349)
 To: "cdi-spec/cdi" <cdi@noreply.github.com>
 Cc: "Mark Struberg" <struberg@yahoo.de>, "Team mention" <team_mention@noreply.github.com>
 Date: Tuesday, 6 December, 2016, 9:37
 
 mkouba commented on this pull
 request.
 
 
 
 > +     *
 +     * The returned Instance object can
 only access instances of  beans that are available for
 injection in the module
 +     * or library containing the class
 into which the <tt>BeanManager</tt> was injected
 or the Java EE component from whose JNDI
 +     * environment namespace the
 <tt>BeanManager</tt> was obtained, according to
 the rules of typesafe resolution.
 +     *
 +     * Note that when called during
 invocation of an {@link AfterBeanDiscovery} event observer,
 +     * this method will only return
 instances of beans discovered by the container before the
 {@link AfterBeanDiscovery} event is fired.
 +     *
 +     * Instances of dependent scoped
 beans obtained with this Instance should be explicitly
 destroyed with {@link Instance#destroy(Object)}
 +     *
 +     * @return an {@link Instance}
 object to request beans instances
 +     * @throws IllegalStateException if
 called during application initialization, before the {@link
 AfterBeanDiscovery}
 +     *     
    event is fired.
 +     * @since 2.0
 +     */
 +    Instance<Object> getInstance();
 
 -1 for two methods. The "root" `Instance` might be reused in
 theory so there is no real gain in terms of resources. Also
 `BeanManager` has even now too many methods.
 
 +0 for `createInstance()`.
 
 Hm, just an idea - what if we make `BeanManager implements
 Instance<Object>` similarly as we do for `CDI` and
 `SeContainer`?
 
 -- 
 You are receiving this because you are on a team that was
 mentioned.
 Reply to this email directly or view it on GitHub:
 https://github.com/cdi-spec/cdi/pull/349

mkouba:
@struberg You shall not reuse any injected `Instance` but for objects like `CDI` and `SeContainer` the "root" instance has the "CreationalContext of application" because there is no real bean instance!

api/src/main/java/javax/enterprise/inject/spi/BeanManager.java
mkouba: Hm, forget about this idea, methods like `isUnsatisfied()`, `stream()`, etc. might bring even more confusion...

manovotn: > Hm, just an idea - what if we make BeanManager implements Instance<Object> similarly as we do for CDI and SeContainer?


I don't really like this. I mean it would be great to allow to do `BM.select(...)`, however in the same time one could do `BM.stream()`, or `BM.isResolvable()` etc. which does not really make sense.

antoinesd:
Ok, to rename the method to `createInstance()`, I'll add extra warnings on user responsibility to destroy dependent instances.

johnament:
I'm OK with `createInstance()`

keilw:
Sounds good to me as well.

PR #351, opened 2016-12-02
CDI-633 Deprecate fireEvent() method more explictly in spec doc

antoinesd:
Deprecate fireEvent() method more explictly in spec doc

### DIFF
--- a/spec/src/main/asciidoc/core/spi.asciidoc
+++ b/spec/src/main/asciidoc/core/spi.asciidoc
@@ -410,28 +410,7 @@ Event<Object> getEvent();
 
 The returned instance can be used like a standard `Event` as described in <<events>>.
 
-===== `fireEvent()` method
-
-_The following method was deprecated in CDI 2.0.
-CDI applications are now encouraged to use `BeanManager.getEvent()` instead._
-
-The method `BeanManager.fireEvent()` fires an event and notifies observers, according to <<observer_notification>>.
-
-[source, java]
-----
-public void fireEvent(Object event, Annotation... qualifiers);
-----
-
-The first argument is the event object.
-The remaining parameters are event qualifiers.
-
-If the runtime type of the event object contains a type variable, an `IllegalArgumentException` is thrown.
-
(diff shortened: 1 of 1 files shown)


mkouba:
-1

We should not remove entire sections - it's not consistent with what we did in CDI 1.1 and 1.2, see for example "@New qualified beans".

tremes:
Uf we should not likely remove all the text describing the original behaviour. I agree. But it seemed to me that this section `fireEvent() method` was introduced accidentally by the previous commit.

manovotn:
@tremes I think the only added text was about deprication of the `fireEvent` method.
Anyway I agree with Martin. We might not want to keep the whole text though. Can we just keep the first few lines up to (and excluding) the source code part? The detailed description of method args could be left out since we deprecate it.

antoinesd:
@mkouba it IS consistent with what we did in CDI 1.1 look at the end of sections: http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#bm_obtain_injectiontarget

IMO the `@New` section was let in place because nothing replaced it

mkouba:
Ok, fine. All the information is in javadoc anyway. +0 from me then ;-)

Emily-Jiang:
The javadoc should not be deleted as it is just deprecated not being deleted.

johnament:
@Emily-Jiang  the javadoc was not deleted.

Emily-Jiang:
I meant the whole section should not be deleted from the spec :o.

antoinesd:
@Emily-Jiang it was not deleted but replaced by one sentence like we did for `createInjectionTarget()` during CDI 1.1 : http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#bm_obtain_injectiontarget

Chapter 11 is already 35 pages long (nearly 20% of the spec), removing section for deprecated feature is one solution to avoid bloating the doc.

For the record, in Java EE, when a feature is deprecated it's not removed from the spec but its usage is not encouraged...

johnament:
+1 to this change  - if we're deprecating it, we shouldn't provide info to new users.

PR #352, opened 2016-12-02
CDI-655 Allow CDIProvider to specify the priority

antoinesd:
CDI-655 Allow CDIProvider to specify the priority

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/CDI.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/CDI.java
@@ -53,7 +53,10 @@ public abstract class CDI<T> implements Instance<T> {
 
     /**
      *
-     * @return the {@link CDIProvider} retrieved by serviceloader or setted by user
+     * Obtain the {@link CDIProvider} the user set with {@link #setCDIProvider(CDIProvider)}, or if it wasn't set, use the
+     * serviceloader the retrieve the {@link CDIProvider} with the highest priority.
+     *
+     * @return the {@link CDIProvider} set by user or retrieved by serviceloader
      */
     private static CDIProvider getCDIProvider() {
         if (configuredProvider != null) {
@@ -69,7 +72,7 @@ public abstract class CDI<T> implements Instance<T> {
             }
             configuredProvider = discoveredProviders.stream()
                     .filter(c -> c.getCDI() != null)
-                    .findAny().orElseThrow(() -> new IllegalStateException("Unable to access CDI"));
+                    .findFirst().orElseThrow(() -> new IllegalStateException("Unable to access CDI"));
             return configuredProvider;
         }
     }
@@ -97,7 +100,7 @@ public abstract class CDI<T> implements Instance<T> {
     private static void findAllProviders() {
 
         ServiceLoader<CDIProvider> providerLoader;
-        Set<CDIProvider> providers = new LinkedHashSet<>();
+        Set<CDIProvider> providers = new TreeSet<>(Comparator.comparingInt(CDIProvider::getPriority).reversed());
 
         providerLoader = ServiceLoader.load(CDIProvider.class, CDI.class.getClassLoader());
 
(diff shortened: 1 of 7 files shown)


antoinesd:
I wonder if we we should set a default value for `getPriority()` in `CDIProvider`.
Spec doc should be improved by resolving ticket [CDI-593](https://issues.jboss.org/browse/CDI-593)

johnament:
I don't see any benefit to this change.  This sounds like an implementation specific issue.  Why would a single implementation have multiple on the classpath?

mkouba:
@antoinesd 
> I wonder if we we should set a default value for getPriority() in CDIProvider.

I think we must provide a default priority value, otherwise we're not backward compatible.

@johnament @Emily-Jiang 
>  This sounds like an implementation specific issue. 
> Why would a single implementation have multiple on the classpath?

For example if using embedded servlet container (Weld SE +  Weld Servlet) - see also your use-case described in [WELD-2260](https://issues.jboss.org/browse/WELD-2260). Even if we merge your fix (thanks for the PR ;-) we still cannot use the optimized version of `CDIProvider` from Weld SE because Weld may not affect the lookup logic implemented in `javax.enterprise.inject.spi.CDI`.

johnament:
@mkouba your argument is a bit confusing.  CDI spec doesn't currently provide any servlet integration and we have said that SE and EE use cases are completely disjointed, no overlap.  I don't see this coming up.

mkouba:
@johnament If we don't specify this then implementations would have to add some hack (like I mentioned in [WELD-2260](https://issues.jboss.org/browse/WELD-2260?focusedCommentId=13330616&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13330616)) just to make it work properly in non-standardized environments (Servlets, OSGi, etc.) to support projects like Hammock. As I said before an implementation cannot modify `javax.enterprise.inject.spi.CDI`.

It's true there is also `CDI.setCDIProvider(CDIProvider)` but I take this method as a last resort.

Do you have a "real" problem with this rather minor change which could make integration in non-standardized environments easier?

antoinesd:
@mkouba @johnament perhaps the spec should mention that this feature could help support non-standardized environments. We already have places in the spec where we provide a way for implementations to provide their specific parameters.

johnament:
@mkouba I don't see how this makes integration into non-standard environments any easier.  As identified in WELD-2260, the problem had nothing to do with the order of CDIProvider, but instead ensuring that Weld Servlet used the same BeanManager (not CDIProvider!) as was bootstrapped in Weld SE.

mkouba:
@johnament Not really, we've fixed WELD-2260 in a different way but there are still related issues. E.g. whenever a user does `CDI.current()` a `CDIProvider` is looked up and if there are multiple impls (for example Weld SE and Servlet) there is no guarantee a user will get the right one, optimized for the given scenario (Weld SE in this case).

John, from my point of view this is really a minor change which does not affect most of the users at all. Do you have a strong argument against it? I don't understand your motivation here.

tremes:
I think it's ok, but I guess we should say what will happen when two providers have accidentally same priority.

antoinesd:
@tremes as always: non-portable results ;). You're right I add the mention.

mkouba:
@tremes @antoinesd I wouldn't use "non-portable" but "undefined" ;-) We use the same wording e.g. for decorators: _"The order of more than one decorator with the same priority is undefined."_

Emily-Jiang:
I think most scenarios will not use this. However, with this support, it is possible to overwrite a default cdi provider if they wish. If there is a user case for this support, I'm fine with it.

mkouba:
Note that it's always possible to skip the default providers (specified as a service provider) by means of `CDI.setCDIProvider(CDIProvider)`.

Emily-Jiang:
@mkouba true. Just try to find a use case for this support. Take your example of Servlet and SE CDI providers: In Servlet environment, do you give the Servlet 'CDI provider' a higher priority, while in SE, SE CDI provider a higher priority?

mkouba:
@Emily-Jiang In the example mentioned above both Weld SE and Servlet are used and since Weld only allows to "reuse" Weld SE container in Weld Servlet (see also http://docs.jboss.org/weld/reference/latest/en-US/html/environments.html#_weld_se_and_weld_servlet_cooperation) the Weld SE provider should have higher priority.

Emily-Jiang:
@mkouba thanks. understood now.

PR #354, opened 2016-12-06
CDI-658 Introduce ProducerConfigurator

mkouba:


api/src/main/java/javax/enterprise/inject/spi/ProcessProducer.java
```diff
@@ -73,6 +77,21 @@
...
+     * <p>
+     * Each call returns the same configurator instance within an observer notification.
+     * </p>
+     * 
+     * @return a non reusable {@link ProducerConfigurator} to configure the original
+     *         {@link javax.enterprise.inject.spi.Producer}.
+     * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 2.0
+     */
+    public ProducerConfigurator<X> configureProducer();
```
antoinesd: for consistency with other Lifecycle events where we provide access to configurators, thus method should be renamed to `setProducer()`

api/src/main/java/javax/enterprise/inject/spi/configurator/ProducerConfigurator.java
```diff
@@ -0,0 +1,95 @@
...
+    <U extends T> ProducerConfigurator<T> produceWith(Function<CreationalContext<U>, U> callback);
+
+    /**
+     * Set a callback to destroy the produced instance.
+     * 
+     * @param callback
+     * @return self
+     * @see Producer#dispose(Object)
+     */
+    ProducerConfigurator<T> disposeWith(Consumer<T> callback);
```
antoinesd: why not. But why only here and not `produceWith()` as well?

api/src/main/java/javax/enterprise/inject/spi/ProcessProducer.java
mkouba: `ProcessAnnotatedType.configureAnnotatedType()`, `ProcessBeanAttributes.configureBeanAttributes()` ...?

api/src/main/java/javax/enterprise/inject/spi/configurator/ProducerConfigurator.java
mkouba: I would leave it as it is for 2.0.

api/src/main/java/javax/enterprise/inject/spi/ProcessProducer.java
antoinesd: forget this, I'm lacking sleep...

manovotn:
Looks good

PR #355, opened 2016-12-06
CDI-652 add simple example for InterceptionFactory

tremes:


spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1652,4 +1652,16 @@ An `InterceptionFactory` can be obtain be calling `BeanManager.createInterceptio
 
 The container must provide a bean with scope `@Dependent`, bean type `InterceptionFactory` and qualifier `@Default` allowing producer methods to obtain this factory to produce their instance.
 
+Following example demonstrates producer method definition using `InterceptionFactory`:
+
```
antoinesd: you could perhaps add a few word to say that the returned value will have the interceptor binding `ActionBinding`

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1631,4 +1631,16 @@ An `InterceptionFactory` can be obtain be calling `BeanManager.createInterceptio
 
 The container must provide a bean with scope `@Dependent`, bean type `InterceptionFactory` and qualifier `@Default` allowing producer methods to obtain this factory to produce their instance.
 
+Following example demonstrates producer method definition using `InterceptionFactory`. Interceptor using `ActionBinding` interceptor binding will be associated with produced bean:
```
tremes: I am not sure I like the second sentence but so far I couldn't come up with anything better.

manovotn: I think your version is good. But if you still think it needs something, maybe transfer it to passive voice?
"Produced bean will have a single interceptor binding associated with it - `ActionBinding`."

tremes: Thanks for suggestions. Maybe following would be better:
`Produced bean will have a single interceptor associated by ActionBinding`, wdyt?

manovotn: That's not exact, is it? I mean you can have more interceptors linked to the bean via the same binding.

tremes: Yes you can but for the purpose of this example it's ok I guess. I wanted to emphasize that the binding is de facto the association which isn't so obvious to me from your proposal.

manovotn: Yea you are probably right, I am being too generic :)
It's ok for this example

PR #356, opened 2016-12-07
Update my previous wrong formatting and wording.

tremes:
I formatted service provider as a class, which is definitely wrong. Also it's better to start sentence with anything else than method name.


manovotn:
+1

PR #357, opened 2016-12-07
CDI-593 Mention javax.enterprise.inject.spi.Prioritized in spec text and improve its javadoc

antoinesd:
Added to Custom bean, interceptor and decoractor


api/src/main/java/javax/enterprise/inject/spi/Decorator.java
```diff
@@ -25,7 +25,10 @@
  * <p>
  * Represents an enabled {@linkplain javax.decorator decorator}.
  * </p>
- * 
+ * <p>
+ * Since CDI 2.0, an implementation this interface may implement {@link Prioritized} to be enabled for the entire application with a priority value.
```
manovotn: This sentence does not really make sense. I suggest following changes:
1) `an implementation this interface` -> `an implementation of this interface`
2) `to be enabled for the entire application with a priority value.` -> `in order to enable the decorator with given priority value for entire application`

Plus the same changes for `api/src/main/java/javax/enterprise/inject/spi/Interceptor.java `

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -47,6 +47,9 @@ An alternative may be given a priority for the application:
 * by placing the `@Priority` annotation on the bean class of a managed bean, or
 * by placing the `@Priority` annotation on the bean class that declares the producer method, field or resource.
 
+For a custom implementation of the `Bean` interface defined in <<bean>>, the container calls `isAlternative()` to determine whether the bean is an alternative and the bean must implement the <<prioritized, `Prioritized` interface>> to select this alternative for the entire application and give it a priority.
```
manovotn: I see several problems here. Firstly you probably meant to say that only if it is alternative than it should implement prioritized interface. Secondly `must implement` is IMHo wrong. What would mean any existing alternative custom bean would become invalid unless it implements prioritized. How about this?:
_Custom bean implementations which are also alternatives may implement <<prioritized, `Prioritized` interface>> in which case they will be enabled for entire application with given priority._

api/src/main/java/javax/enterprise/inject/spi/Decorator.java
tremes: +1

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
mkouba: +1 for Matej's wording and a separate sentence.

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -131,13 +131,28 @@ public interface ObserverMethod<T> {
 * `getObservedType()` and `getObservedQualifiers()` return the observed event type and qualifiers.
 * `getReception()` returns `IF_EXISTS` for a conditional observer and `ALWAYS` otherwise.
 * `getTransactionPhase()` returns the appropriate transaction phase for a transactional observer method or `IN_PROGRESS` otherwise.
-* `getPriority()` returns the priority that will be used by the container to determine the notification order.
+* `getPriority()` this method inherited from <<prioritized, `Prioritized` interface>> returns the priority that will be used by the container to determine the notification order.
```
mkouba: Shouldn't we mention the default value?

manovotn: What exactly do you mean? If users choose to implement this interface, they will have to add _some_ value. They needn't really care for default. Or am I missing something?

mkouba: public default int getPriority() {
        return DEFAULT_PRIORITY;
    }

api/src/main/java/javax/enterprise/inject/spi/BeanAttributes.java
```diff
@@ -67,6 +67,8 @@
 
     /**
      * Determines if the bean is an {@linkplain javax.enterprise.inject.Alternative alternative}.
+     *
+     * If it is an alternative it may may implement {@link Prioritized} to be selected for the entire application with a priority value.
```
mkouba: I don't think it's useless. An author of a custom bean will read this javadoc. But the wording should be something like:

A custom implementation of `Bean` may implement `Prioritized` in order to be selected for the application. `Prioritized.getPriority()` determines the priority used to resolve ambiguities.

api/src/main/java/javax/enterprise/inject/spi/Decorator.java
mkouba: ₊1

api/src/main/java/javax/enterprise/inject/spi/Interceptor.java
```diff
@@ -26,6 +26,9 @@
  * <p>
  * Represents an enabled {@linkplain javax.interceptor interceptor}.
  * </p>
+ * <p>
+ * Since CDI 2.0, an implementation this interface may implement {@link Prioritized} to be enabled for the entire application with a priority value.
```
mkouba: Same as for decorators...

api/src/main/java/javax/enterprise/inject/spi/Prioritized.java
```diff
@@ -19,20 +19,30 @@
 
 /**
  * <p>
- * Provides access to the priority value of prioritized elements.
+ * This interface allows some SPI implementation to change their priority programmatically.
  * </p>
  *
+ * <p>
+ * For instance The {@link ObserverMethod} interface extends this interface to set the observer priority.
```
mkouba: For instance, {@link ObserverMethod} extends this interface to define the observer priority.

api/src/main/java/javax/enterprise/inject/spi/Decorator.java
antoinesd: +1, far better understandable, but shouldn't it be: 

> ... with **the** given priority value ...

api/src/main/java/javax/enterprise/inject/spi/BeanAttributes.java
antoinesd: I put it here because it's the only place where we talk about alternative custom bean.
I agree with Martin and take his wording proposal.

spec/src/main/asciidoc/core/spi.asciidoc
antoinesd: Probably also the "non-portable behaviour" if the synthetic observer method is asynchronous

mkouba: Yes.

johnament: What do you mean by "if"? If a user implements ObserverMethod they are implementing Prioritized.

tremes:
+1

PR #358, opened 2016-12-07
CDI-663 clarify statement wrt repeating qualifiers/annotations.

tremes:


manovotn:
I agree with this solution. This makes the spec clear while preserving current behaviour.

mkouba:
Looks good.

PR #361, opened 2016-12-08
CDI-656 Clarify effect of @Priority on ordered observers

antoinesd:


spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -576,6 +576,7 @@ void afterLogin(@Observes @Priority(APPLICATION) LoggedInEvent event) { ... }
 
 The order of more than one observer with the same priority is undefined and the observer methods are notified therefore in a non predictable order.
 
+If a `@Priority` annotation is set on an asynchronous observer non-portable behaviour results.
```
mkouba: And what about:
If a `@Priority` annotation is declared on an event parameter of an asynchronous observer method, non-portable behavior results.

I would put this sentence next to `The priority of an observer method may be declared using the @Priority annotation.`, line 568. And maybe change the first sentence to `The priority of a synchronous observer method may be declared by annotating the event parameter with @Priority.`

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -566,6 +566,9 @@ If no exception is thrown by observers then the resulting CompletionStage is com
 
 Before the actual observer notification, the container determines an order in which the observer methods for a certain event are invoked.
 The priority of an observer method may be declared using the `@Priority` annotation.
```
mkouba: The priority of an observer method may be declared **by annotating the event parameter with `@Priority` annotation**.

Otherwise looks good ;-)

mkouba:
Looks good.

PR #363, opened 2016-12-09
CDI-495 What happens if an illegal bean type is found in the set of bean types

antoinesd:
illegal bean types in type set are ignored

### DIFF
--- a/spec/src/main/asciidoc/core/definition.asciidoc
+++ b/spec/src/main/asciidoc/core/definition.asciidoc
@@ -89,6 +89,8 @@ However, some Java types are not legal bean types :
 * A parameterized type that contains a wildcard type parameter is not a legal bean type.
 * An array type whose component type is not a legal bean type.
 
+If the set of bean types for a bean contains non-legal bean types, the container will ignore them during <<typesafe_resolution, typesafe resolution>>.
+
 Note that certain additional restrictions are specified in <<unproxyable>> for beans with a normal scope, as defined in <<normal_scope>>.
 
 [[restricting_bean_types]]
--- a/spec/src/main/asciidoc/core/definition.asciidoc
+++ b/spec/src/main/asciidoc/core/definition.asciidoc
@@ -89,8 +89,6 @@ However, some Java types are not legal bean types :
 * A parameterized type that contains a wildcard type parameter is not a legal bean type.
 * An array type whose component type is not a legal bean type.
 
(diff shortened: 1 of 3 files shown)


spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -89,6 +89,8 @@ However, some Java types are not legal bean types :
 * A parameterized type that contains a wildcard type parameter is not a legal bean type.
 * An array type whose component type is not a legal bean type.
 
+If the set of bean types for a bean contains non-legal bean types, the container will ignore them during <<typesafe_resolution, typesafe resolution>>.
```
mkouba: I think that we should rather state that illegal bean types are not added to the set of bean types. Note that we may need to ignore these types not only during typesafe resolution, i.e. proxy class generation...

I would also add a short note to all relevant sections in chapter 3, e.g. "3.1.2. Bean types of a managed bean".

antoinesd: I hesitated to put it in each part. If we go that way (putting in different part) we should remove it from here to avoid denormalisation of the spec. wdyt ?

mkouba: Yes, probably...

johnament: Agreed, unfortunately putting it in multiple places makes more sense than here.

spec/src/main/asciidoc/core/implementation.asciidoc
```diff
@@ -51,6 +51,8 @@ An application can prevent packages being split across jars by sealing the packa
 
 The unrestricted set of bean types for a managed bean contains the bean class, every superclass and all interfaces it implements directly or indirectly.
 
+If the resulting set of bean types for a managed bean contains <<legal_bean_types,non-legal bean types>>, these types are removed from the set of bean types.
```
manovotn: `non-legal` - it does not really make sense here. The opposite of legal is `illegal`.
Also the word illegal is already used in spec.

Or, if you don't like the word `illegal`, we may twist the sentence like this:
`The resulting set of bean types for a managed bean consist only of <<legal_bean_types,legal bean types>>, all other types are removed from the set of bean types.`

antoinesd: +1

manovotn:
LGTM

mkouba:
Looks good.

PR #364, opened 2016-12-12
ProducerConfigurator missing in chapter 11.6

tremes:


spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1558,6 +1558,19 @@ If any `ProcessProducer` method is called outside of the observer method invocat
...
+public interface ProducerConfigurator<T> {
+
+    <U extends T> ProducerConfigurator<T> produceWith(Function<CreationalContext<U>, U> callback);
+    ProducerConfigurator<T> disposeWith(Consumer<T> callback);
+    ProducerConfigurator<T> addInjectionPoint(InjectionPoint injectionPoint);
+    ProducerConfigurator<T> addInjectionPoints(InjectionPoint... injectionPoints);
+    ProducerConfigurator<T> addInjectionPoints(Set<InjectionPoint> injectionPoints);
+    ProducerConfigurator<T> injectionPoints(InjectionPoint... injectionPoints);
+}
+----
```
antoinesd: -1 to remove them from the spec doc, we need to have them officially specified. Javadoc is an helper but is not officially art of the spec.

tremes: @antoinesd so would you keep it as it is or will you add these java samples for each configurator?
Maybe I would add it for each.

PR #365, opened 2016-12-13
CDI-232 Loosen the requirement of Instance and Provider to only match injection points.

johnament:
I believe we spoke about this on some medium, I suspect IRC, some while ago.  Where we left is that the container only has to support beans being installed that match the injection points identified, not beans for Instances that had no injection targets.

As a result, I reworded the first line only to clarify this and rephrase it to make the reading into the bullets cleaner (in my opinion).

### DIFF
--- a/spec/src/main/asciidoc/core/injectionandresolution.asciidoc
+++ b/spec/src/main/asciidoc/core/injectionandresolution.asciidoc
@@ -751,14 +751,13 @@ An `UnsupportedOperationException` is thrown if the active context object for th
 
 ==== The built-in `Instance`
 
-The container must provide a built-in bean with:
+For each injection point of `Instance<?>` or `Provider<?>`, the container must provide a built-in bean that satisfies:
 
 * `Instance<X>` and `Provider<X>` for every legal bean type `X` in its set of bean types,
 * every qualifier type in its set of qualifier types,
 * scope `@Dependent`,
 * no bean name, and
 * an implementation provided automatically by the container.
-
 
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -751,15 +751,14 @@ An `UnsupportedOperationException` is thrown if the active context object for th
 
 ==== The built-in `Instance`
 
-The container must provide a built-in bean with:
+For each injection point of `Instance<?>` or `Provider<?>`, the container must provide a built-in bean that satisfies:
 
 * `Instance<X>` and `Provider<X>` for every legal bean type `X` in its set of bean types,
```
mkouba: CDI-232 is about this sentence that needs to be changed. 

I propose something like:

The container must provide a built-in bean:
* which is eligible for injection to any injection point with required type `Instance<X>` or `Provider<X>`, where `X` is any legal bean type, and any required qualifiers
* has `@Dependent` scope,
* has no bean name,
* an implementation is provided automatically by the container.

antoinesd: +1 for @mkouba proposal

johnament:
Ok, I'm thinking *any* is a good way to describe the qualifiers here.

mkouba:
Yep, I think it looks good now.

Emily-Jiang:
LGTM

manovotn:
+1 from me

PR #367, opened 2016-12-19
CDI-669 specify exception for subsequent calls of InterceptionFactory…

tremes:
….createInterceptedInstance

### DIFF
--- a/spec/src/main/asciidoc/core/spi.asciidoc
+++ b/spec/src/main/asciidoc/core/spi.asciidoc
@@ -1692,8 +1692,8 @@ These method should never be invoked upon bean instances. Otherwise, unpredictab
  It will bypass standard rules defined in <<unproxyable>>.
 * `configure()` returns an `AnnotatedTypeConfigurator` (as defined in <<annotated_type_configurator>>) initialized with the instance type to easily apply specific interceptor binding to apply when enhancing the instance.
 The method always return the same `AnnotatedTypeConfigurator`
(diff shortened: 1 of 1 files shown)


manovotn:
+1

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1693,7 +1693,7 @@ These method should never be invoked upon bean instances. Otherwise, unpredictab
 * `configure()` returns an `AnnotatedTypeConfigurator` (as defined in <<annotated_type_configurator>>) initialized with the instance type to easily apply specific interceptor binding to apply when enhancing the instance.
 The method always return the same `AnnotatedTypeConfigurator`
 * `createInterceptedInstance()` returns an enhanced version of the instance for which each method invocations will be a business method invocation.
```
johnament: May require follow up - does this imply the self-method calls limitation is lifted?

mkouba: I think this should be discussed in the context of the self-invocation issue: https://issues.jboss.org/browse/CDI-414.

mkouba: I would reword 
> `createInterceptedInstance()` returns an enhanced version of the instance for which each method invocations will be a business method invocation.

to something like:

> `createInterceptedInstance()` returns an enhanced version of the instance whose method invocations are treated as business method invocations as defined in 7.2. Container invocations and interception.

tremes:
Martin's suggestion incorporated.

PR #368, opened 2017-01-03
CDI-667 introduce new chapter for trimmed bean archive in Java EE.

tremes:
This is first proposal and I am open to any suggestions.


johnament:
+1

PR #370, opened 2017-01-03
CDI-578 reword original proposal.

tremes:
I tried to reword the original PR from Mark to include the original suggestions.

### DIFF
--- a/api/src/main/java/javax/enterprise/event/TransactionPhase.java
+++ b/api/src/main/java/javax/enterprise/event/TransactionPhase.java
@@ -21,13 +21,16 @@ package javax.enterprise.event;
  * Distinguishes the various kinds of transactional {@linkplain javax.enterprise.event.Observes observer methods} from regular
  * observer methods which are notified immediately.
  * </p>
- * 
  * <p>
  * Transactional observer methods are observer methods which receive event notifications during the before or after completion
  * phase of the transaction in which the event was fired. If no transaction is in progress when the event is fired, they are
  * notified at the same time as other observers.
+ * If the transaction is in progress, but {@link javax.transaction.Synchronization} callback cannot be registered due to the transaction being already
+ * marked for rollback or in state where {@link javax.transaction.Synchronization} callbacks cannot be registered, the {@link #BEFORE_COMPLETION},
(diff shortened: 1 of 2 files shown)


api/src/main/java/javax/enterprise/event/TransactionPhase.java
```diff
@@ -21,13 +21,16 @@
  * Distinguishes the various kinds of transactional {@linkplain javax.enterprise.event.Observes observer methods} from regular
  * observer methods which are notified immediately.
  * </p>
- * 
  * <p>
  * Transactional observer methods are observer methods which receive event notifications during the before or after completion
  * phase of the transaction in which the event was fired. If no transaction is in progress when the event is fired, they are
  * notified at the same time as other observers.
+ * If the transaction is in progress, but {@link javax.transaction.Synchronization} callback cannot be registered due to the transaction is already
```
manovotn: I would change the sentence a bit (`due to the transaction is` sounds weird). How about this:
`If the transaction is in progress, but {@link javax.transaction.Synchronization} callback cannot be registered due to the transaction being already marked for rollback or in state where {@link javax.transaction.Synchronization} callbacks cannot be registered, ... `

tremes: Agree that it sounds weird. I think we can go with your proposal.

api/src/main/java/javax/enterprise/event/TransactionPhase.java
```diff
@@ -45,13 +48,23 @@
      * <p>
      * Identifies a before completion observer method, called during the before completion phase of the transaction.
      * </p>
+     * <p>
+     * It will also get invoked if there is no transaction in progress, or the transaction is in progress,
```
mkouba: I'm not a native speaker but "get invoked" sounds wrong.

manovotn: I'd say it's OK as long as we talk about methods (which we do - it's OM here).
But we might also use something like "get notified" if that sounds better?

mkouba:
Looks good.

manovotn:
+1

PR #374, opened 2017-01-10
CDI-668 Most Example with @Priority provide non fully qualified constant name

antoinesd:
CDI-668 Most Example with @Priority provide non fully qualified constant name


mkouba:
Looks ready to merge.

PR #375, opened 2017-01-10
CDI-670 Clarify InterceptionFactory.ignoreFinalMethods() purpose and functioning

antoinesd:
CDI-670 Clarify InterceptionFactory.ignoreFinalMethods() purpose and functioning

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/InterceptionFactory.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/InterceptionFactory.java
@@ -83,12 +83,10 @@ public interface InterceptionFactory<T> {
 
     /**
      *
-     * <p>Instructs the container to ignore all non-static, final methods with public, protected or default visibility declared on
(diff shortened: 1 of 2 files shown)


api/src/main/java/javax/enterprise/inject/spi/InterceptionFactory.java
```diff
@@ -88,7 +88,7 @@
      *
```
mkouba: I would rather modify the first sentense to something like:

Instructs the container to ignore all non-static, final methods with public, protected or default visibility declared by any class in the type hierarchy of the intercepted instance during invocation of `createInterceptedInstance()`.

antoinesd: +1. The current sentence didn't fell accurate.

antoinesd:
New wording agreed on IRC

PR #377, opened 2017-01-12
Improve InterceptionFactory wording

mkouba:
This PR is an attempt to improve the wording of `InterceptionFactory` javadoc. I realize that the proposed wording is not perfect. Still I think it's more understandable. Also few typos are fixed.

### DIFF
--- a/api/src/main/java/javax/enterprise/inject/spi/InterceptionFactory.java
+++ b/api/src/main/java/javax/enterprise/inject/spi/InterceptionFactory.java
@@ -17,19 +17,20 @@
 
 package javax.enterprise.inject.spi;
 
+import javax.enterprise.context.Dependent;
 import javax.enterprise.context.spi.CreationalContext;
+import javax.enterprise.inject.Default;
 import javax.enterprise.inject.spi.configurator.AnnotatedTypeConfigurator;
 
 /**
- * <p>An {@link InterceptionFactory} adds the services available for instances created by the container to any instance.
- * It makes each method invocation in the instance a business method invocation as defined in section 7.2 of the specification document
- * It will bind business methods only to <tt>@AroundInvoke</tt> interceptor methods.
+ * {@link InterceptionFactory} allows to create a wrapper instance whose method invocations are intercepted by method
+ * interceptors and forwarded to a provided instance.
+ *
+ * <p>
+ * An implementation of {@link InterceptionFactory} may be obtained by calling
+ * {@link BeanManager#createInterceptionFactory(CreationalContext, Class)} to be used in the create method of a custom bean for
+ * example.
  * </p>
- *
- * <p>An implementation of {@link InterceptionFactory} may be obtain by
- * calling {@link BeanManager#createInterceptionFactory(CreationalContext, Class)}
- * to be used in the create method of a custom bean for example.</p>
- *
  *
  * <pre>
  * public class MyCustomBean implements Bean&lt;MyClass&gt; {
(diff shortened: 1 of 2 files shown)


antoinesd:
This wording is clearer but I thought we agreed to not use the term "proxy" here (probably why the initial wording was rather clumsy). I don't have a strong opinion on that point but It worth mentioning the original intent.

mkouba:
Hm, I don't remember the details. But one thing is the name of interface/methods and the other thing is the way a functionality is described. I don't think we create an "enhanced version of instance" though.

antoinesd:
Then you should also bring these modification to the spec (section 11.7) for consistency sake, IMO.

mkouba:
@antoinesd Yes of course, but I wanted to make an agreement first.

tremes:
I think it's not a problem to use term "proxy" here but maybe I would specify this term little bit like `client proxy` (which is not this case I guess) so `interception proxy`? 

Otherwise looks good.

mkouba:
AFAIK the original name of this interface was `InterceptionProxyFactory` but then during the F2F the EG decided to rename it :-)

antoinesd:
@mkouba, @tremes, renaming was done because the spec didn't mention proxy anywhere regarding interceptors. Interception is bind to business method definition. So we decided to avoid introducing the term proxy which was more implementation oriented.
As I said, that explains the difficulty I had to describe `InterceptionFactory`.

antoinesd:
So question is. Can you provide a better wording without using proxy ?

tremes:
@antoinesd Well one thing is the spec text itself and another thing is API. This PR doesn't change spec text and we use term `proxy` in API anyway e.g - https://github.com/cdi-spec/cdi/blob/master/api/src/main/java/javax/enterprise/inject/Instance.java#L225

antoinesd:
@tremes the term client proxy is well defined and used in the spec for bean instances in normal scope. I was pointing to the Interceptor feature.
Don't get me wrong: I agree that the current wording is not very good, but I was pointing to the EG intent that lead to this wording and interface renaming.

mkouba:
@antoinesd @tremes What about **wrapper**? I'll try to reword...

tremes:
@antoinesd ahh ok I see

antoinesd:
@mkouba @tremes in fact, if we want to stick to the 1.x spec wording: InterceptionFactory is a tool that transform method invocation into business method invocation for a given instance. So wrapper is probably a good match here.

mkouba:
> transform method invocation into business method invocation

Except that decorators are not supported.

api/src/main/java/javax/enterprise/inject/spi/InterceptionFactory.java
```diff
@@ -82,43 +80,58 @@
...
     /**
+     * Returns an {@link AnnotatedTypeConfigurator} initialized with the {@link AnnotatedType} created either for the class
+     * passed to {@link BeanManager#createInterceptionFactory(CreationalContext, Class)} or derived from the
+     * {@link InterceptionFactory} parameter injection point.
      *
-     * Returns an {@link AnnotatedTypeConfigurator} to allow addition of interceptor binding on the instance's methods.
-     * The matching annotatedType will be used to apply the interceptors when calling createInterceptedInstance method.
-     * Annotations that are not interceptor binding will be ignored.
+     * <p>
+     * The configurator allows to add or remove interceptor bindings used to associate interceptors with the proxy instance
```
antoinesd: ...wrapper instance...

mkouba:
I've also changed the title of the section because "Apply interceptor programmatically" sounds misleading to me.

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1672,46 +1672,45 @@ The container must provide implementation for all these configurators and make t
...
+If the provided instance was created by the container, this method does nothing and returns the provided instance.
 
-An `InterceptionFactory` can be obtain be calling `BeanManager.createInterceptionFactory` as defined in <<bm_obtain_interception_factory>>
+An `InterceptionFactory` can be obtained be calling `BeanManager.createInterceptionFactory()` as defined in <<bm_obtain_interception_factory>>.
 
-The container must provide a bean with scope `@Dependent`, bean type `InterceptionFactory` and qualifier `@Default` allowing producer methods to obtain this factory to produce their instance.
+The container must provide a built-in bean with scope `@Dependent`, bean type `InterceptionFactory` and qualifier `@Default`.
 
-Following example demonstrates producer method definition using `InterceptionFactory`. Produced bean will have a single interceptor associated by `ActionBinding`:
+If an injection point of type `InterceptionFactory` and qualifier `@Default` which is not a parameter of a producer method exists, the container automatically detects the problem and treats it as a definition error.
```
manovotn: While this is correct, it is rather difficult to read. How about this:
> If an injection point of type `InterceptionFactory` and qualifier `@Default` exists and it is not a parameter of a producer method, the container automatically detects the problem and treats it as a definition error.`

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1672,46 +1672,45 @@ The container must provide implementation for all these configurators and make t
 
 [[interception_factory]]
 
-=== Apply interceptor programmatically
-
-CDI 2.0 introduces the `javax.enterprise.inject.spi.InterceptionFactory<T>` interface to add services to an instance not created by the container by making each of its method invocation, a business method invocation as defined in <<biz_method>>.
-It will bind business methods only to `@AroundInvoke` interceptor methods.
+=== The `InterceptionFactory` interface
 
+CDI 2.0 introduces the `javax.enterprise.inject.spi.InterceptionFactory<T>` interface which allows to create a wrapper instance whose method invocations are intercepted by method interceptors and forwarded to a provided instance.
```
manovotn: Tiny note; I think there should be comma here:
`... interface, which allows to ...`

manovotn:
I think this one is ready for merge

PR #380, opened 2017-03-01
CDI-684 static observers on lifecycle events in extension are non-portable

antoinesd:


johnament:
+1

spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -1016,6 +1016,7 @@ The container is not required to deliver events fired during application initial
 
 The container instantiates a single instance of each extension at the beginning of the application initialization process and maintains a reference to it until the application shuts down.
 The container delivers event notifications to this instance by calling its observer methods.
+If an extension contains static observer method of a container lifecycle event or Object class, non-portable behavior results.
```
mkouba: I propose...

> If an extension declares a static observer method whose event parameter type:
> - is a container lifecycle event, or
> - is `java.lang.Object` and the event parameter has either no qualifiers or a single qualifier `@Any`,
>  
> non-portable behavior results.

If a `java.lang.Object` observer has any other qualifier, it's obviously not a container lifecycle event observer. See also "10.1. Event types and qualifier types":
> Every event has the qualifier @javax.enterprise.inject.Any, even if it does not explicitly declare this qualifier.

antoinesd: +1

PR #381, opened 2017-03-02
CDI-686 interface as InterceptionFactory's type parameter is not portbale

antoinesd:


spec/src/main/asciidoc/core/spi.asciidoc
```diff
@@ -697,6 +697,7 @@ The method `BeanManager.getInterceptionFactory()` returns an `InterceptionFactor
 <T> InterceptionFactory<T> createInterceptionFactory(CreationalContext<T> ctx, Class<T> clazz);
 ----
 
+If type parameter `T` is not a class non-portable behavior results.
```
manovotn: Why can't we just specifically mention interfaces?
E.g. `is not a class` -> `is an interface`
Also, there should be a comma before "non-portable" (`is not a class, non-portable behavior results.`)

mkouba: What about...
> If the actual type parameter of the method is not a Java class, non-portable behavior results.

I think we should also add a similar note to API, `javax.enterprise.inject.spi.InterceptionFactory<T>`.

antoinesd: What about Enum ?

manovotn: That would be truly weird use case. I mean can you even make a producer for enum so that you would want to intercept it?

tremes: Well the point is that how to exclude the enum so that it will result to non-portable behavior as well I guess. Stating `is an interface` is then not sufficient, but I am still not sure about better wording here.

manovotn: I know what Antoine meant. To me, trying to intercept produced Enum sounds just so bizarre I barely see it worth mentioning. But you might be right.

mkouba: Guys, we normally use "Java class" in the spec, see also my proposal below...

johnament: What happens if its an abstract class?

manovotn: Just tried it with Weld and it **works** like a charm.

antoinesd: @johnament abstract class is not a problem IMO. It's like applying an interceptor binding on abstract class method.

PR #385, opened 2017-03-21
CDI-690 Request Context Clarification

johnament:
Clarify in core when request context is active, and change verbiage a bit to align to common term of context.

### DIFF
--- a/spec/src/main/asciidoc/core/scopescontexts.asciidoc
+++ b/spec/src/main/asciidoc/core/scopescontexts.asciidoc
@@ -491,6 +491,16 @@ An event with qualifier `@Initialized(RequestScoped.class)` is synchronously fir
 An event with qualifier `@BeforeDestroyed(RequestScoped.class)` is synchronously fired when the request context is about to be destroyed, i.e. before the actual destruction.
 An event with qualifier `@Destroyed(RequestScoped.class)` is synchronously fired when the request context is destroyed, i.e. after the actual destruction.
 
+The request context is active:
+
+* during notification of an asynchronous observer method, and
+* during `@PostConstruct` callback of any bean.
+
+The request context is destroyed:
+
+* after the invocation of an asynchronous observer method completes, and
+* after the `@PostConstruct` callback completes, if it did not already exist when the `@PostConstruct` callback occurred.
+
 [[session_context]]
 
 ==== Session context lifecycle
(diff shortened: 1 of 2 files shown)


mkouba:
Looks good.

manovotn:
+1

johnament:
I updated the PR with the "destroyed" path as well.

antoinesd:
shouldn't we mention RequestContextController and manual activation? Otherwise looks good

johnament:
@antoinesd in this section?  Its talked about pretty extensively in `scopescontexts.adoc` about 200 lines up.  This section seems to be about cases where the context is automatically started.  It could be that we have to move that section down as well.

spec/src/main/asciidoc/core/scopescontexts.asciidoc
```diff
@@ -491,6 +491,16 @@ An event with qualifier `@Initialized(RequestScoped.class)` is synchronously fir
 An event with qualifier `@BeforeDestroyed(RequestScoped.class)` is synchronously fired when the request context is about to be destroyed, i.e. before the actual destruction.
 An event with qualifier `@Destroyed(RequestScoped.class)` is synchronously fired when the request context is destroyed, i.e. after the actual destruction.
 
+The request context is active:
+
+* during any asynchronous invocation of an event observer,
```
mkouba: To be precise: _"during notification of an asynchronous observer method"_

spec/src/main/asciidoc/core/scopescontexts.asciidoc
```diff
@@ -491,6 +491,16 @@ An event with qualifier `@Initialized(RequestScoped.class)` is synchronously fir
...
+* during `@PostConstruct` callback of any bean.
+
+The request context is destroyed:
+
+* after the asynchronous invocation of an event observer,
```
mkouba: _"after the invocation of an asynchronous observer method completes"_?

johnament: Agreed with both - but just to point out this is existing wording.  I can change it.

mkouba: Ok

antoinesd:
@johnament I was thinking about a mention like "request context can also be activated deactivated programatically as specified in <<section above>>". The idea being to have all the info on its lifecycle (even the non automatic one) in one place.

PR #390, opened 2018-06-25
CDI-699 AnnotationLiteral should use privileged actions for reflective operations

antoinesd:
Addition of privileged bloc in AnnotationLiteral

### DIFF
--- a/api/src/test/java/org/jboss/cdi/api/test/privileged/annotation/AnnotationLiteralTest.java
+++ b/api/src/test/java/org/jboss/cdi/api/test/privileged/annotation/AnnotationLiteralTest.java
@@ -1,0 +1,55 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2018, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.cdi.api.test.privileged.annotation;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+/**
+ *
+ * This class test usage of privileged bloc in {@link javax.enterprise.util.AnnotationLiteral}
+ * They launched with Java security Manager.
+ *
+ * @author Antoine Sabot-Durand
+ *
+ */
+public class AnnotationLiteralTest {
+
+    @Test
+    public void toStringShouldWork() {
+        Assert.assertNotNull(mal.toString());
+    }
+
+    @Test
+    public void annotationTypeShouldWork() {
+        Assert.assertEquals(MyAnnotation.class, mal.annotationType());
+    }
+
+    @Test
+    public void equalsShouldWork() {
+        MyAnnotationLiteral mal2 = new MyAnnotationLiteral("hello", 42);
+        Assert.assertEquals(mal, mal2);
+    }
+
+    @Test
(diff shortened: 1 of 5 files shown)


api/src/test/java/org/jboss/cdi/api/test/privileged/annotation/AnnotationLiteralTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+package org.jboss.cdi.api.test.privileged.annotation;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+/**
+ *
+ * This class test usage of privileged bloc in {@link javax.enterprise.util.AnnotationLiteral}
+ * They launched with Java security Manager.
```
mkouba: Where do we run this test with a `SecurityManager` used?

mkouba: Ah, you pushed one commit before this PR... https://github.com/cdi-spec/cdi/commit/c9aedeaadc9828cf2b5b8c3efc37f733a8f7569a

mkouba:
I think you should also remove https://github.com/cdi-spec/cdi/blob/master/api/pom.xml#L301.

mkouba:
Hm, it seems that `AnnotationLiteralTest` cannot reveal the problem with `java.lang.Class.getDeclaredMethods()` because (javadoc):
````
Throws:
SecurityException - If a security manager, s, is present and any of the following conditions is met:
* the caller's class loader is not the same as the class loader of this class and invocation of s.checkPermission method with RuntimePermission("accessDeclaredMembers") denies access to the declared methods within this class
* the caller's class loader is not the same as or an ancestor of the class loader for the current class and invocation of s.checkPackageAccess() denies access to the package of this class
```

manovotn:
The test configuration is really puzzling, I don't see why you firstly skip test execution and then re-enable it in both phases (e.g. [L301](https://github.com/cdi-spec/cdi/blob/master/api/pom.xml#L301), [L314](https://github.com/cdi-spec/cdi/blob/master/api/pom.xml#L314), [L327](https://github.com/cdi-spec/cdi/blob/master/api/pom.xml#L327)). If you remove all three, I think it will still execute - or is there any special reason for it?

api/src/test/java/org/jboss/cdi/api/test/privileged/annotation/AnnotationLiteralTest.java
antoinesd: Yes I hesitated but I have several PR to send with SM activated so as this addition doesn't change anything to release and test I found more convenient to have this on the common branch.

antoinesd:
@manovotn I need 2 test suites because the standard test are doing a lot of reflection and I couldn't activate SM for them. Adding a second test suite was necessary to activate SM for it. I have problem while overriding the standard test so I have to cancel it (`<skip>true</skip>`). But perhaps I can give another try...

antoinesd:
> The test configuration is really puzzling

I achieved to simplify surefire config and make tests launch clearer by removing `<skip>true</skip>` and override `default-test` suite instead of disabling it and create a new one.

struberg:
Isn't this an impl detail? Ofc it's the official spec api, but this doesn't change the contract. So nothing needs to be released spec wise, right?

manovotn:
> Isn't this an impl detail?

@struberg no it is not. User can have his own implementation of `AnnotationLiteral` which naturally extends this class. And on instantiation of that class the constructor on `AnnotationLiteral` is invoked. That in turn invokes `getMembers()` method which then invokes `getDeclaredMethods()` and that can cause issues with security manager.

struberg:
@manovotn yes, the api.jar should get fixed. But that can be done as a simple maintenance release. It doesn't require a new spec revision.

mkouba:
@struberg Yes, we target a "simple API maintenance release" named _CDI 2.0.SP1_; except that it's probably not so simple because of JCP rules etc. @antoinesd knows more.

antoinesd:
@struberg. It's a maintenance release. Doing it in the context of the JCP would have required to launch a MR (more than 60 days).

struberg:
oki @antoinesd and @mkouba I think we are on the same page. Wenn talking about a cdi-api 'Maintenance Release' I also didn't mean a CDI Spec MR.
In my eyes it's really just a bugfix release of the cdi-api.jar. Kind of 2.0.1. 

In any case nothing which requires to involve the EE EC for getting it to the people.

PR #391, opened 2018-06-26
CDI-727 CDI.current() should use privileged block

antoinesd:


antoinesd:
I was unable to recreate a similar error than in the ticket. We may consider adding a small jar with a fake CDI provider to be in the same condition.
I had to create another `SecurityActions` class for the api package: not very happy with that: suggestions are welcome

mkouba:
I don't think that `ServiceLoader.load()` requires a privileged block. See also my comment: https://issues.jboss.org/browse/CDI-727?focusedCommentId=13597453&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13597453

> I had to create another SecurityActions class for the api package.

No problem. That's a common practice.

antoinesd:
Yes, good point. I forgot that we switched to service loader from 1.2 to 2.0

manovotn:
>> I had to create another SecurityActions class for the api package.

> No problem. That's a common practice.

Hmm, just genuinely interested - why is it common practice? Just to prevent anyone using it? I thought these SM calls are basically caller-sensitive. So even if it was used by someone else, it shouldn't cause harm. But I guess I am missing something here?

mkouba:
@manovotn If you make the `SecurityActions` public then anyone can use it and thus "hijack" the permissions assigned to Weld.

manovotn:
Hmm that's exactly what I thought would be addressed by caller sensitivity, apparently I was wrong. Thanks for explanation.

hkalina:
@antoinesd I just tested this with EntityListenerBeanManagerInjectionTestCase in wildfly testsuite (with removed workaround and set using cdi-api:2.1-SNAPSHOT) and I confirm this fixes the bug. :+1:

antoinesd:
Thanks @hkalina. Do you have a suggestion to create a simple test to reproduce your issue? Mine is obviously too basic since it doesn't trigger any `AccessControlException` without privileged code bloc.

hkalina:
@antoinesd are you sure the test is running with security manager enabled? (is System.getSecurityManager() not null?)

PR #393, opened 2018-11-22
CDI-738 Generate Jakarta EE version of CDI API

antoinesd:
Changing GAV for Jakarta EE.
Usage of Jakarta EE staging repo for EL and Interceptor API (please use -Pstaging when building)
Publication targeted to Jakarta EE repository.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -25,10 +25,10 @@
 		<relativePath />
 	</parent>
 
-	<groupId>javax.enterprise</groupId>
-	<artifactId>cdi-api</artifactId>
+	<groupId>jakarta.enterprise</groupId>
+	<artifactId>jakarta.enterprise.cdi-api</artifactId>
 	<packaging>jar</packaging>
-	<version>2.1-SNAPSHOT</version>
+	<version>2.0.1-SNAPSHOT</version>
 
 	<name>CDI APIs</name>
 	<description>APIs for CDI (Contexts and Dependency Injection for Java)</description>
@@ -41,11 +41,6 @@
 			<distribution>repo</distribution>
 		</license>
 	</licenses>
-
-	<ciManagement>
-		<system>Hudson</system>
-		<url>http://hudson.jboss.org</url>
-	</ciManagement>
 
 	<issueManagement>
 		<system>JIRA</system>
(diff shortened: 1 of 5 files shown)


api/pom.xml
```diff
@@ -19,16 +19,16 @@
...
-		<version>36</version>
+		<groupId>org.eclipse.ee4j</groupId>
+		<artifactId>project</artifactId>
+		<version>1.0.5</version>
 		<relativePath />
 	</parent>
 
-	<groupId>javax.enterprise</groupId>
+	<groupId>jakarta.enterprise</groupId>
 	<artifactId>cdi-api</artifactId>
```
gunnarmorling: Shouldn't this be `jakarta.cdi-api`, following the pattern of the other specs at https://wiki.eclipse.org/New_Maven_Coordinates?

manovotn: Oracle's old artifacts had `javax` prefix in their `artifactId` whereas CDI never did.
Though I agree it makes sense to have similar syntax as all other Jakarta artifacts.

antoinesd: I get your point but it was the coordinate proposed by Dmitry in his first mail and other `enterprise` spec seems to use this as well as you check un the staging repo: https://oss.sonatype.org/content/groups/staging/jakarta/

manovotn: Not sure I understand you, all the artifacts in the [repo you mentioned](https://oss.sonatype.org/content/groups/staging/jakarta/) are using that prefix :)

antoinesd: But as other specs in `enterprise` hierarchy follow include `jakarta` in their named, we probably should follow this schema. that would give:

        <groupId>jakarta.enterprise</groupId>
        <artifactId>jakarta.enterprise.cdi-api</artifactId>

Wdyt?

gunnarmorling: Better double check with Dmitry then what he meant. In fact following the example of the other "enterprise" specs (not that I'd know what that even means), it even should be `jakarta.enterprise.cdi-api` probably.

For BV, we'll use jakarta.validation-api as the artifact id (we also just had validation-api before, but changed it to be consistent with the others).

gunnarmorling: Ha, cross-posted, but we arrived at the same conclusion 👍

manovotn: Yea, that sounds just about right

antoinesd: Done!

PR #394, opened 2019-02-15
Keep last valid discovered provider in cachedProvider field

doychin:
Use that provider until it is still valid and provides active cdi container.
Otherwise run discovery process again.


manovotn:
I would rather lean towards the solution where you expect `getCDI()` can throw exception since that honors the actual contract.

doychin:
I will update that PR

PR #405, opened 2019-08-03
Try working around current staging problems

starksm64:


spec/src/main/asciidoc/dictionary.txt
```diff
@@ -15,8 +15,8 @@ decorators
 DoD
 EE
 EG
-EJB
-EJBs
+Jakarta Eneterprise Bean
```
pzygielo: _Eneterprise_ doesn't look right...

PR #417, opened 2020-02-12
Clean Travis build

antoinesd:
Align Travis build on other Jakarta EE specs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,15 @@
+dist: trusty
 language: java
 sudo: false
-install: mvn -Pstaging install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
-script: mvn -Pstaging test -B
+
 jdk:
-  - oraclejdk8
+  - oraclejdk11
+  - openjdk11
 
 cache:
   directories:
-  - $HOME/.m2
+    - $HOME/.m2
 
-notifications:
-  email:
-    - cdi-dev@lists.jboss.org
-  irc:
-    channels:
-      - "chat.freenode.net#cdi-dev"
-    on_success: change  # options: [always|never|change] default: always
-    on_failure: always  # options: [always|never|change] default: always
-    on_start: never     # options: [always|never|change] default: always
+script:
+  - cd api
+  - mvn verify -B -V -Poss-release -Dgpg.skip=true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,21 @@
-dist: trusty
 language: java
 sudo: false
-
+install: mvn -Pstaging install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+script: mvn -Pstaging test -B
 jdk:
-  - oraclejdk11
+  - openjdk8
   - openjdk11
 
 cache:
   directories:
-    - $HOME/.m2
+  - $HOME/.m2
 
-script:
-  - cd api
-  - mvn verify -B -V -Poss-release -Dgpg.skip=true
+notifications:
+  email:
+    - cdi-dev@eclipse.org
+  irc:
+    channels:
+      - "chat.freenode.net#cdi-dev"
+    on_success: change  # options: [always|never|change] default: always
+    on_failure: always  # options: [always|never|change] default: always
+    on_start: never     # options: [always|never|change] default: always


.travis.yml
```diff
@@ -1,20 +1,15 @@
+dist: trusty
```
manovotn: I don't think you need this. It's yet another thing we will forget to change later on. Leaving it out means you will run on latest (Xenial) which is just fine for the versions you need.

.travis.yml
```diff
@@ -1,20 +1,15 @@
...
 language: java
 sudo: false
-install: mvn -Pstaging install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
-script: mvn -Pstaging test -B
+
 jdk:
-  - oraclejdk8
+  - oraclejdk11
```
manovotn: Why oracle version? Isn't openjdk good enough? From what I recall, there was absolutely no difference between those two since JDK 9.

.travis.yml
```diff
@@ -1,20 +1,15 @@
+dist: trusty
 language: java
 sudo: false
-install: mvn -Pstaging install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
-script: mvn -Pstaging test -B
+
 jdk:
-  - oraclejdk8
```
manovotn: Are we dropping `jdk8` altogether? I think we should keep that, [Xenial images can still run `openjdk8`](https://github.com/weld/core/blob/master/.travis.yml).

.travis.yml
```diff
@@ -1,20 +1,15 @@
...
-  irc:
-    channels:
-      - "chat.freenode.net#cdi-dev"
-    on_success: change  # options: [always|never|change] default: always
-    on_failure: always  # options: [always|never|change] default: always
-    on_start: never     # options: [always|never|change] default: always
\ No newline at end of file
+script:
+  - cd api
+  - mvn verify -B -V -Poss-release -Dgpg.skip=true
```
manovotn: Since this is for ordinary PR builds, why run with `oss-release` profile?

starksm64:
We need to keep the jdk8 build for Jakarta EE 9. This build is failing because we are not pulling in the dependent SNAPSHOTs. It could be due to the parent pom needing to be updated, or we need to activate the -Pstaging profile, or they may not have been staged yet. I'll update the parent and look into this further.


`[ERROR] Failed to execute goal on project jakarta.enterprise.cdi-api: Could not resolve dependencies for project jakarta.enterprise:jakarta.enterprise.cdi-api:jar:3.0.0-SNAPSHOT: The following artifacts could not be resolved: jakarta.el:jakarta.el-api:jar:4.0.0-SNAPSHOT, jakarta.interceptor:jakarta.interceptor-api:jar:2.0.0-SNAPSHOT, jakarta.inject:jakarta.inject-api:jar:2.0-SNAPSHOT: Could not find artifact jakarta.el:jakarta.el-api:jar:4.0.0-SNAPSHOT in sonatype-snapshots (https://oss.sonatype.org/content/repositories/snapshots/) -> [Help 1]
`

antoinesd:
Build breaks for 2 reasons:

1) The use of oracle jdk8 seems to be not supported by Travis anymore:

```
Installing oraclejdk8[0m
$ export JAVA_HOME=~/oraclejdk8
$ export PATH="$JAVA_HOME/bin:$PATH"
$ ~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"
Ignoring license option: BCL -- using GPLv2+CE by default
install-jdk.sh 2020-01-14
Expected feature release number in range of 9 to 15, but got: 8
```

2) The lack of Interceptor spec with `Jakarta` packages in release, staging or snapshots repository

for 1) I'll switch to `openjdk8` as @manovotn suggested. For 2) we should ask interceptor spec to release a snapshot or a RC

.travis.yml
antoinesd: You're right. I'll switch to open jdk 8.

manovotn: I meant to have both - `openjdk8` and `openjdk11`

antoinesd: yes, that's what I understood

Emily-Jiang:
> For 2) we should ask interceptor spec to release a snapshot or a RC
Interceptor is in wave1. I expect a RC to happen very soon.

manovotn:
This PR is outdated and stale. Travis config was already updated.
Closing this one.

PR #429, opened 2020-04-08
Fixed parent version number in cdi-spec pom file

moghaddam:
The parent version of the spec module is still referring to the old one. This PR fixes that.


pzygielo:
It seems to be fixed by #415 already, with spec/pom.xml using 3.0.0-SNAPSHOT. Please check/refresh your clone.

moghaddam:
Oh, you're right. My bad😟

Issue #425, opened 2020-02-18
CDI lite


Emily-Jiang:
In light of this [discussion](https://groups.google.com/forum/#!topic/microprofile/_vML3EzLKIE) on allowing widely adopted by compile-time runtimes, the following use case is requried.

As a implementor of CDI spec, I would like to support the CDI parts, which can be determained at compile time so that my end users can benefit from CDI and AOT.

aguibert:
I like the idea of a CDI Lite, but we should define it in terms of features/capabilities, rather than what the CDI implementers can figure out what to get working in AOT runtimes.

Perhaps we could start by making a list of features that are very computationally expensive or complex compared to the benefit they provide to end users?

For example, CDI Lite could have the default bean discovery mode be "annotated" and then "all" could be added by the "full" CDI spec, where it could become the default.

manovotn:
My personal 0.02$ are that I generally think we need a Lite version of CDI spec that will cut out some not-so-used parts and bits that aren't built time friendly. Something like that will help CDI stay up to date with rising technologies/frameworks in Java. However, we need to proceed carefully - first version would definitely have to be just a subset of existing features, the smaller the better, so that we don't include things that aren't necessary and might just keep getting in our way further down the line. Better to underspecify then to overspecify what we want to support, at least in my opinion :)

> CDI Lite could have the default bean discovery mode be "annotated"

Yeah, that's for sure, the `all` mode is just ridiculously expensive to process for the "good" that it brings.

Another candidates _for exclusion_ in Lite (just from the top of my head) would be things like decorators, less used scopes in microservice world (conversation/session). We'd definitely drop any EE integrations from Lite spec as those don't need to be present at all. And probably much more.

Emily-Jiang:
Here is the list of functionalities I think CDI core should provide:

- Relationships to other specifications: Relationship to Dependency Injection for 

- [ ] Relationship to Dependency Injection for Java

- [ ] Relationship to Java Interceptors


-

Cousjava:
I think that portable extensions shouldn't be part of core, however I do think that ConversationScoped and possibly custom scopes should be in core.

manovotn:
@Emily-Jiang Portable extensions are by nature runtime components; they are very much against any AOT concept. Take a look at this article by Antoine which says explains why - http://www.cdi-spec.org/news/2020/03/09/CDI_for_the_future/
[...]

PR #451, opened 2020-09-16
initial Build Compatible Extension API proposal

Ladicek:
EDITED by manovotn:

Issue for this can be found here - https://github.com/eclipse-ee4j/cdi/issues/452
The blog post can be seen here - http://www.cdi-spec.org/news/2020/09/15/CDI_Lite_extension/

Please do take a look at them as they describe this in depth as well as define a way to provide feedback (in a separate issue so that we don't clutter one issue/PR).

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationConfig.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationConfig.java
@@ -1,0 +1,23 @@
+package jakarta.enterprise.inject.build.compatible.spi;
+
+import jakarta.enterprise.lang.model.AnnotationAttribute;
+import jakarta.enterprise.lang.model.AnnotationInfo;
+import jakarta.enterprise.lang.model.declarations.ClassInfo;
+import java.lang.annotation.Annotation;
+import java.util.function.Predicate;
+
+// TODO better name?
+// TODO devise a builder-style API instead (see also Annotations)
+public interface AnnotationConfig {
+    void addAnnotation(Class<? extends Annotation> annotationType, AnnotationAttribute... attributes);
+
+    void addAnnotation(ClassInfo<?> annotationType, AnnotationAttribute... attributes);
+
+    void addAnnotation(AnnotationInfo annotation);
+
+    void addAnnotation(Annotation annotation);
+
+    void removeAnnotation(Predicate<AnnotationInfo> predicate);
+
+    void removeAllAnnotations();
+}
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Annotations.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Annotations.java
@@ -1,0 +1,84 @@
+package jakarta.enterprise.inject.build.compatible.spi;
+
+import jakarta.enterprise.lang.model.AnnotationAttribute;
+import jakarta.enterprise.lang.model.AnnotationAttributeValue;
+import jakarta.enterprise.lang.model.AnnotationInfo;
+import jakarta.enterprise.lang.model.declarations.ClassInfo;
+import java.lang.annotation.Annotation;
+import java.util.List;
+
+// TODO devise a builder-style API instead (see also AnnotationConfig)
+public interface Annotations {
+    AnnotationAttributeValue value(boolean value);
+
+    AnnotationAttributeValue value(byte value);
+
+    AnnotationAttributeValue value(short value);
+
+    AnnotationAttributeValue value(int value);
+
+    AnnotationAttributeValue value(long value);
+
+    AnnotationAttributeValue value(float value);
+
+    AnnotationAttributeValue value(double value);
+
+    AnnotationAttributeValue value(char value);
+
+    AnnotationAttributeValue value(String value);
+
+    AnnotationAttributeValue value(Enum<?> enumValue);
+
+    AnnotationAttributeValue value(Class<? extends Enum<?>> enumType, String enumValue);
+
+    AnnotationAttributeValue value(ClassInfo<?> enumType, String enumValue);
+
+    AnnotationAttributeValue value(Class<?> value);
+
+    AnnotationAttributeValue annotationValue(Class<? extends Annotation> annotationType, AnnotationAttribute... attributes);
+
+    AnnotationAttributeValue annotationValue(ClassInfo<?> annotationType, AnnotationAttribute... attributes);
+
+    AnnotationAttributeValue annotationValue(AnnotationInfo annotation);
+
+    AnnotationAttributeValue annotationValue(Annotation annotation);
+
+    AnnotationAttribute attribute(String name, boolean value);
+
+    AnnotationAttribute attribute(String name, byte value);
+
+    AnnotationAttribute attribute(String name, short value);
+
+    AnnotationAttribute attribute(String name, int value);
+
+    AnnotationAttribute attribute(String name, long value);
+
+    AnnotationAttribute attribute(String name, float value);
+
+    AnnotationAttribute attribute(String name, double value);
+
+    AnnotationAttribute attribute(String name, char value);
+
+    AnnotationAttribute attribute(String name, String value);
+
+    AnnotationAttribute attribute(String name, Enum<?> enumValue);
+
+    AnnotationAttribute attribute(String name, Class<? extends Enum<?>> enumType, String enumValue);
+
+    AnnotationAttribute attribute(String name, ClassInfo<?> enumType, String enumValue);
+
+    AnnotationAttribute attribute(String name, Class<?> value);
+
+    AnnotationAttribute arrayAttribute(String name, AnnotationAttributeValue... values);
+
+    AnnotationAttribute arrayAttribute(String name, List<AnnotationAttributeValue> values);
+
+    AnnotationAttribute annotationAttribute(String name, Class<? extends Annotation> annotationType,
+            AnnotationAttribute... attributes);
+
+    AnnotationAttribute annotationAttribute(String name, ClassInfo<?> annotationType, AnnotationAttribute... attributes);
+
+    AnnotationAttribute annotationAttribute(String name, AnnotationInfo annotation);
+
+    AnnotationAttribute annotationAttribute(String name, Annotation annotation);
+}
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AppArchive.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AppArchive.java
@@ -1,0 +1,162 @@
+package jakarta.enterprise.inject.build.compatible.spi;
+
+import jakarta.enterprise.lang.model.declarations.ClassInfo;
+import jakarta.enterprise.lang.model.declarations.FieldInfo;
+import jakarta.enterprise.lang.model.declarations.MethodInfo;
+import jakarta.enterprise.lang.model.types.Type;
+import java.lang.annotation.Annotation;
+import java.util.function.Consumer;
+
+// TODO remove entirely?
+public interface AppArchive {
+    ClassQuery classes();
+
+    MethodQuery constructors(); // no static initializers
+
+    MethodQuery methods(); // no constructors nor static initializers
+
+    FieldQuery fields();
+
+    /**
+     * The {@code exactly} and {@code subtypeOf} methods are additive.
+     * When called multiple times, they form a union of requested classes (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appArchive.classes()
+     *     .exactly(Foo.class)
+     *     .subtypeOf(Bar.class)
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for the {@code Foo} class and all subtypes of the {@code Bar} class.
+     * <p>
+     * The {@code annotatedWith} methods are additive.
+     * When called multiple times, they form a union of requested annotations (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appArchive.classes()
+     *     .annotatedWith(Foo.class)
+     *     .annotatedWith(Bar.class)
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all classes annotated either with {@code @Foo} or with {@code @Bar} (or both).
+     */
+    interface ClassQuery {
+        ClassQuery exactly(Class<?> clazz);
+
+        ClassQuery subtypeOf(Class<?> clazz);
+
+        ClassQuery annotatedWith(Class<? extends Annotation> annotationType);
+
+        void forEach(Consumer<ClassInfo<?>> consumer);
+    }
+
+    /**
+     * The {@code declaredOn} method is additive.
+     * When called multiple times, it forms a union of requested classes (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appArchive.methods()
+     *     .declaredOn(appArchive.classes().exactly(Foo.class))
+     *     .declaredOn(appArchive.classes().subtypeOf(Bar.class))
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all methods declared on the {@code Foo} class and on all subtypes of the {@code Bar} class.
+     * Note that this example can be rewritten as
+     * <pre>{@code
+     * appArchive.methods()
+     *     .declaredOn(appArchive.classes().exactly(Foo.class).subtypeOf(Bar.class))
+     *     .forEach(...)
+     * }</pre>
+     * which is probably easier to understand.
+     * <p>
+     * The {@code withReturnType} methods are additive.
+     * When called multiple times, they form a union of requested return types (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appArchive.methods()
+     *     .withReturnType(Foo.class)
+     *     .withReturnType(Bar.class)
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all methods that return either {@code Foo} or {@code Bar}.
+     * <p>
+     * The {@code annotatedWith} methods are additive.
+     * When called multiple times, they form a union of requested annotations (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appArchive.methods()
+     *     .annotatedWith(Foo.class)
+     *     .annotatedWith(Bar.class)
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all methods annotated either with {@code @Foo} or with {@code @Bar} (or both).
+     */
+    interface MethodQuery {
+        MethodQuery declaredOn(ClassQuery classes);
+
+        // equivalent to `withReturnType(Types.of(type))`
+        MethodQuery withReturnType(Class<?> type);
+
+        MethodQuery withReturnType(Type type);
+
+        // TODO parameters?
+
+        MethodQuery annotatedWith(Class<? extends Annotation> annotationType);
+
+        void forEach(Consumer<MethodInfo<?>> consumer);
+    }
+
+    /**
+     * The {@code declaredOn} method is additive.
+     * When called multiple times, it forms a union of requested classes (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appArchive.fields()
+     *     .declaredOn(appArchive.classes().exactly(Foo.class))
+     *     .declaredOn(appArchive.classes().subtypeOf(Bar.class))
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all fields declared on the {@code Foo} class and on all subtypes of the {@code Bar} class.
+     * Note that this example can be rewritten as
+     * <pre>{@code
+     * appArchive.fields()
+     *     .declaredOn(appArchive.classes().exactly(Foo.class).subtypeOf(Bar.class))
+     *     .forEach()
+     * }</pre>
+     * which is probably easier to understand.
+     * <p>
+     * The {@code ofType} methods are additive.
+     * When called multiple times, they form a union of requested field types (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appArchive.fields()
+     *     .ofType(Foo.class)
+     *     .ofType(Bar.class)
+     *     .forEach()
+     * }</pre>
+     * runs given code for all fields that are of type either {@code Foo} or {@code Bar}.
+     * <p>
+     * The {@code annotatedWith} methods are additive.
+     * When called multiple times, they form a union of requested annotations (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appArchive.fields()
+     *     .annotatedWith(Foo.class)
+     *     .annotatedWith(Bar.class)
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all fields annotated either with {@code @Foo} or with {@code @Bar} (or both).
+     */
+    interface FieldQuery {
+        FieldQuery declaredOn(ClassQuery classes);
+
+        // equivalent to `withReturnType(Types.of(type))`
+        FieldQuery ofType(Class<?> type);
+
+        FieldQuery ofType(Type type);
+
+        FieldQuery annotatedWith(Class<? extends Annotation> annotationType);
+
+        void forEach(Consumer<FieldInfo<?>> consumer);
+    }
+}
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AppArchiveBuilder.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AppArchiveBuilder.java
@@ -1,0 +1,10 @@
+package jakarta.enterprise.inject.build.compatible.spi;
+
+// TODO better name
+public interface AppArchiveBuilder {
+    void add(String fullyQualifiedClassName);
+
+    // TODO adds the type itself or not? (in theory yes, as subtyping is reflexive)
+    // TODO looks like it can't be implemented on top of Portable Extensions, so maybe remove?
+    void addSubtypesOf(String fullyQualifiedClassName);
+}
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AppArchiveConfig.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AppArchiveConfig.java
@@ -1,0 +1,66 @@
+package jakarta.enterprise.inject.build.compatible.spi;
+
+import jakarta.enterprise.lang.model.types.Type;
+import java.lang.annotation.Annotation;
+import java.util.function.Consumer;
+
+// TODO remove entirely, if we remove AppArchive/AppDeployment
+// TODO maybe AppArchiveConfig shouldn't extend AppArchive, and *ConfigQuery shouldn't extend *Query
+public interface AppArchiveConfig extends AppArchive {
+    @Override
+    ClassConfigQuery classes();
+
+    @Override
+    MethodConfigQuery constructors(); // no static initializers
+
+    @Override
+    MethodConfigQuery methods(); // no constructors nor static initializers
+
+    @Override
+    FieldConfigQuery fields();
+
+    interface ClassConfigQuery extends ClassQuery {
+        @Override
+        ClassConfigQuery exactly(Class<?> clazz);
+
+        @Override
+        ClassConfigQuery subtypeOf(Class<?> clazz);
+
+        @Override
+        ClassConfigQuery annotatedWith(Class<? extends Annotation> annotationType);
+
+        void configure(Consumer<ClassConfig<?>> consumer);
+    }
+
+    interface MethodConfigQuery extends MethodQuery {
+        @Override
+        MethodConfigQuery declaredOn(ClassQuery classes);
+
+        @Override
+        MethodConfigQuery withReturnType(Class<?> type);
+
+        @Override
+        MethodConfigQuery withReturnType(Type type);
+
+        @Override
+        MethodConfigQuery annotatedWith(Class<? extends Annotation> annotationType);
+
+        void configure(Consumer<MethodConfig<?>> consumer);
+    }
+
+    interface FieldConfigQuery extends FieldQuery {
+        @Override
+        FieldConfigQuery declaredOn(ClassQuery classes);
+
+        @Override
+        FieldConfigQuery ofType(Class<?> type);
+
+        @Override
+        FieldConfigQuery ofType(Type type);
+
+        @Override
+        FieldConfigQuery annotatedWith(Class<? extends Annotation> annotationType);
+
+        void configure(Consumer<FieldConfig<?>> consumer);
+    }
+}
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AppDeployment.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AppDeployment.java
@@ -1,0 +1,138 @@
+package jakarta.enterprise.inject.build.compatible.spi;
+
+import jakarta.enterprise.lang.model.declarations.ClassInfo;
+import jakarta.enterprise.lang.model.types.Type;
+import java.lang.annotation.Annotation;
+import java.util.function.Consumer;
+
+// TODO remove entirely?
+public interface AppDeployment {
+    BeanQuery beans();
+
+    ObserverQuery observers();
+
+    /**
+     * The {@code scope} methods are additive.
+     * When called multiple times, they form a union of requested scope types (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appDeployment.beans()
+     *     .scope(Foo.class)
+     *     .scope(Bar.class)
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all beans with the {@code @Foo} scope or the {@code @Bar} scope.
+     * <p>
+     * The {@code type} methods are additive.
+     * When called multiple times, they form a union of requested bean types (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appDeployment.beans()
+     *     .type(Foo.class)
+     *     .type(Bar.class)
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all beans with the {@code Foo} type or the {@code Bar} type (or both).
+     * Note that bean type is not just the class which declares the bean (or return type of a producer method,
+     * or type of producer field). All superclasses and superinterfaces are also included in the set of bean types.
+     * <p>
+     * The {@code qualifier} methods are additive.
+     * When called multiple times, they form a union of requested qualifiers (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appDeployment.beans()
+     *     .qualifier(Foo.class)
+     *     .qualifier(Bar.class)
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all beans with the {@code @Foo} qualifier or the {@code @Bar} qualifier (or both).
+     * <p>
+     * The {@code declaringClass} methods are additive.
+     * When called multiple times, they form a union of requested declaration classes (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appDeployment.beans()
+     *     .declaringClass(Foo.class)
+     *     .declaringClass(Bar.class)
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all beans declared on the {@code Foo} class or the {@code Bar} class.
+     */
+    interface BeanQuery {
+        BeanQuery scope(Class<? extends Annotation> scopeAnnotation);
+
+        BeanQuery scope(ClassInfo<?> scopeAnnotation);
+
+        BeanQuery type(Class<?> beanType);
+
+        BeanQuery type(ClassInfo<?> beanType);
+
+        BeanQuery type(Type beanType);
+
+        BeanQuery qualifier(Class<? extends Annotation> qualifierAnnotation);
+
+        BeanQuery qualifier(ClassInfo<?> qualifierAnnotation);
+
+        BeanQuery declaringClass(Class<?> declaringClass);
+
+        BeanQuery declaringClass(ClassInfo<?> declaringClass);
+
+        void forEach(Consumer<BeanInfo<?>> consumer);
+
+        void ifNone(Runnable runnable);
+    }
+
+    /**
+     * The {@code observedType} methods are additive.
+     * When called multiple times, they form a union of requested observer types (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appDeployment.observers()
+     *     .observedType(Foo.class)
+     *     .observedType(Bar.class)
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all observers that observe the {@code Foo} event type or the {@code Bar} event type.
+     * <p>
+     * The {@code qualifier} methods are additive.
+     * When called multiple times, they form a union of requested qualifiers (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appDeployment.observers()
+     *     .qualifier(Foo.class)
+     *     .qualifier(Bar.class)
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all observers with the {@code @Foo} qualifier or the {@code @Bar} qualifier (or both).
+     * <p>
+     * The {@code declaringClass} methods are additive.
+     * When called multiple times, they form a union of requested declaration classes (not an intersection).
+     * For example,
+     * <pre>{@code
+     * appDeployment.observers()
+     *     .declaringClass(Foo.class)
+     *     .declaringClass(Bar.class)
+     *     .forEach(...)
+     * }</pre>
+     * runs given code for all observers declared on the {@code Foo} class or the {@code Bar} class.
+     */
+    interface ObserverQuery {
+        ObserverQuery observedType(Class<?> observedType);
+
+        ObserverQuery observedType(ClassInfo<?> observedType);
+
+        ObserverQuery observedType(Type observedType);
+
+        ObserverQuery qualifier(Class<? extends Annotation> qualifierAnnotation);
+
+        ObserverQuery qualifier(ClassInfo<?> qualifierAnnotation);
+
+        ObserverQuery declaringClass(Class<?> declaringClass);
+
+        ObserverQuery declaringClass(ClassInfo<?> declaringClass);
+
+        void forEach(Consumer<ObserverInfo<?>> consumer);
+
+        void ifNone(Runnable runnable);
+    }
+}
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
@@ -1,0 +1,59 @@
+package jakarta.enterprise.inject.build.compatible.spi;
+
+import jakarta.enterprise.lang.model.AnnotationInfo;
+import jakarta.enterprise.lang.model.declarations.ClassInfo;
+import jakarta.enterprise.lang.model.declarations.FieldInfo;
+import jakarta.enterprise.lang.model.declarations.MethodInfo;
+import jakarta.enterprise.lang.model.types.Type;
+import java.util.Collection;
+
+/**
+ * @param <T> type of the inspected bean (that is, not the declaring class, but one of the types the bean has)
+ */
+public interface BeanInfo<T> {
+    // TODO remove the type parameter?
+
+    ScopeInfo scope();
+
+    Collection<Type> types();
+
+    // TODO method(s) for getting AnnotationInfo for given qualifier class?
+    Collection<AnnotationInfo> qualifiers();
+
+    /**
+     * Returns the class that declares the bean.
+     * In case of a bean defined by a class, that is the bean class directly.
+     * In case of a producer method or field, that is the class that declares the producer method or field.
+     * TODO null for synthetic beans, or return Optional?
+     *
+     * @return {@link ClassInfo} for the class that declares the bean
+     */
+    ClassInfo<?> declaringClass();
+
+    boolean isClassBean();
+
+    boolean isProducerMethod();
+
+    boolean isProducerField();
+
+    boolean isSynthetic();
+
+    MethodInfo<?> producerMethod(); // TODO null if not producer method, or return Optional?
+
+    FieldInfo<?> producerField(); // TODO null if not producer field, or return Optional?
+
+    boolean isAlternative();
+
+    int priority();
+
+    // EL name (from @Named), IIUC
+    String getName();
+
+    DisposerInfo disposer(); // TODO null if not producer method/field, or return Optional?
+
+    Collection<StereotypeInfo> stereotypes();
+
+    // TODO interceptors?
+
+    Collection<InjectionPointInfo> injectionPoints();
+}
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildCompatibleExtension.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildCompatibleExtension.java
@@ -1,0 +1,37 @@
+package jakarta.enterprise.inject.build.compatible.spi;
+
+/**
+ * Build compatible extensions are service providers for this interface, as defined in {@link java.util.ServiceLoader}.
+ * This means: they are classes that implement this interface, provide a {@code META-INF/services} file,
+ * and satisfy all other service provider constraints. Additionally, extensions should not be CDI beans
+ * and should not be used at application runtime.
+ * <p>
+ * Extensions can define arbitrary {@code public}, non-{@code static}, {@code void}-returning methods
+ * without type parameters, annotated with one of the extension annotations.
+ * <p>
+ * Extension processing occurs in 4 phases, corresponding to 4 extension annotations:
+ * <ul>
+ * <li>{@link Discovery @Discovery}</li>
+ * <li>{@link Enhancement @Enhancement}</li>
+ * <li>{@link Synthesis @Synthesis}</li>
+ * <li>{@link Validation @Validation}</li>
+ * </ul>
+ * <p>
+ * These methods can declare arbitrary number of parameters. Which parameters can be declared depends
+ * on the particular processing phase and is documented in the corresponding extension annotation.
+ * All the parameters will be provided by the container when the extension is invoked.
+ * <p>
+ * Extension methods can be assigned a priority using {@link jakarta.enterprise.inject.build.compatible.spi.ExtensionPriority @ExtensionPriority}.
+ * Note that priority only affects order of extensions in a single phase.
+ * <p>
+ * If the extension declares multiple methods, they are all invoked on the same instance of the class.
+ * <p>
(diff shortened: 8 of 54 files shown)


manovotn:
FTR, the CI is failing on JDK 11 javadoc build which seems to be stricter than its 8 counterpart.

ljnelson:
This looks interesting.  I assume there was a discussion about this on the mailing list, or in some other publicly-reachable forum? tradeoffs and architecture and design and whatnot?  Could you point me to that, please?

Ladicek:
Announcement and a more "deep dive" elaboration should be coming soon, and this is not a done deal, more like a starting point. Everything is up for scrutiny.

Emily-Jiang:
Did I miss something? I thought we need to work out what should be in CDI Lite first.

Emily-Jiang (changes requested):
Is this PR trying to address #425 ? If yes, we need to agree on the scope. If not, what issue is this PR for?

starksm64:
Yes, this will be a starting point for a proposal for #425. Antoine is putting out an initial blog and a prototype repository that @Ladicek has contributed for discussion. This PR is the API portion of that repo.

manovotn:
Just to reverberate what other said - this is supposed to be a starting point for a discussion, not a resolved thing, don't worry.
It is just a a bit of timezone difference and PTOs wreaking little havoc on synchronization efforts between several people :-)
There should be a blogpost about what it is as well as proper GH issue during hopefully during today.

manovotn:
Ok, the promised bits are here:
* issue - https://github.com/eclipse-ee4j/cdi/issues/452
* blogpost - http://www.cdi-spec.org/news/2020/09/15/CDI_Lite_extension/

Ladicek:
As @manovotn said, there was a little synchronization issue, mostly caused by me being on vacation. This PR, and the linked article, is supposed to be a starting point for a discussion on the Lite extension API. Important part of the Lite effort, but not all of Lite.

I'm sure you will agree that it's best to start a discussion with a concrete proposal, rather than some vague "we need to have extensions, how do you think they should look like?". I mentioned it above, and also in the article, but I'll repeat: everything is up for debate and change. This is not a "take it or leave it" situation, it's exactly the opposite.

graemerocher:
I see a lot of value in how the extension points work in terms of similar things that we do in Micronaut (adding annotations, modifying, removing etc.)

However, one issue I see with this proposal is it assumes the classes are available on the classpath for an extension. For example `annotatedWith(..)` seems to assume you have the annotation on the classpath (although there does seem to be a `ClassInfo` abstraction just no clarity on how to create it without a class). Using generic arguments for scanning again assumes you can reference the classes.

This is problematic as IMO one of the problematic areas in Quarkus is that it places the entire runtime classpath onto the build time enhancement classpath which is one of the things that annotation processors got right by ensuring the annotation processor classpath (where extensions would go) has an isolated classpath separate from compilation and runtime.

This is important because you want to ensure that dependency conflicts in extensions (which will happen) do not conflict and result in hard to debug compiler errors (nobody wants to remote debug a compiler).

Why is there an assumption in this proposed spec that build time enhancement is going to take place on the byte code and not on the source code? Processing the source code has significant advantages in that you can:

* Report errors back to the user when something is wrong pointing directly to the source element where a failure occurred
* Provide an implementation that does not require 2 step compilation (first compile then do byte code enhancement afterwards)
* Automatically support all moderns IDEs that already know how to handle annotation processors.

Any spec that is proposed should not assume the Quarkus approach which has many disadvantages IMO.

mkouba:
> For example annotatedWith(..) seems to assume you have the annotation on the classpath...

That's a good point. There should be an abstraction similar to `org.jboss.jandex.DotName` and `org.jboss.jandex.AnnotationInstance`.

> This is problematic as IMO one of the problematic areas in Quarkus is that it places the entire runtime classpath onto the build time enhancement classpath which is one of the things that annotation processors got right by ensuring the annotation processor classpath (where extensions would go) has an isolated classpath separate from compilation and runtime.

That's not entirely true - see https://quarkus.io/guides/class-loading-reference for more details. 

>  Processing the source code has significant advantages in that you can:

To be fair, it has significant disadvantages as well. Ideally, the API should not prefer one or the other.

graemerocher:
> To be fair, it has significant disadvantages as well. Ideally, the API should not prefer one or the other.

Clearly. Each approach has trade offs. Nevertheless my point remains a standard should not be Quarkus under another name and should allow for other approaches, otherwise it is yet another standard with significant overreach.

Ladicek:
I'm sure you will all agree that we should strive to keep this discussion strictly technical. Also I'd suggest to file a separate issue for this topic, as described in #452, because this is an important topic and if we help all discussions on one place, it will be a mess.

That said, I can tell that I certainly had annotation processors in mind at the beginning (even thought about using `javax.lang.model` as the metamodel), but I kinda forgot about them towards the end. I got focused on user friendliness and it's entirely possible that I made the proposal incompatible with annotation processors as a side effect. Please be assured that this wasn't intentional. I did write a few annotation processors in the past, but it's been a while and they were not terribly big, so I'm certainly no expert in this area. Any input (that is technical and free of accusations, please) is obviously very welcome.

Now -- I really thought that using generics for expressing basic type queries is compatible with annotation processors. I confess I didn't try. If not, we'll have to come up with something different.

Ah and one more thing: representations such as `org.jboss.jandex.DotName` or `javax.lang.model.element.Name` are terribly user unfriendly. I'd really like to stay away from them. So far, I didn't see a situation where plain old strings are not enough.

manovotn:
> Clearly. Each approach has trade offs. Nevertheless my point remains a standard should not be Quarkus under another name and should allow for other approaches, otherwise it is yet another standard with significant overreach.

Please let's avoid pointless accusations. We have tried to make the API as general as we could and if you did check Quarkus, you will know that it has very different 'extension scheme' from what we have brought up in this PR. 

In fact we started with just the API and only later got to creating a working Quarkus prototype in order to be able to play with the plain Lite extension API, test out how user friendly it is and see if you can re-write existing framework integration from Quarkus-specific extensions into actual pure CDI Lite extensions. 

That being said, it is very well possible that we missed a use case or two which is why we are looking for technical feedback and suggestions on how it should look like to support that given use case.
Feel free to go ahead and create and issue describing the use case you are missing and how you'd adapt the API to it. We definitely want to look into that!

n1hility:
@graemerocher FWIW we aren't thinking of (or really wanting) Quarkus to be the "RI", just one compatible implementation. Would be great to have a micronaut impl and an API that can support them both (along with other impls)

graemerocher:
> FWIW we aren't thinking of (or really wanting) Quarkus to be the "RI", just one compatible implementation. Would be great to have a micronaut impl and an API that can support them both (along with other impls)

Thanks @n1hility Nice to receive a rational response from a Red Hat engineer, and yes we are absolutely on the same page on that front. My desire is to support whatever CDI light is and that is a long term goal, but it has to be something that is supportable and not tied to a single implementation with a single architecture (byte code processing).

> Also I'd suggest to file a separate issue for this topic, as described in #452, because this is an important topic and if we keep all discussions on one place, it will be a mess.

I am not sure I understand, what topic are you referring to? The desire to have an API that everyone can support seems pretty relevant to this.

> Please let's avoid pointless accusations. I'm sure you will all agree that we should strive to keep this discussion strictly technical.

@Ladicek @manovotn Please clarify where I accused anyone of anything or where I deviated from technical discussion, it seems the both of you are a touch emotional. My concern as stated is merely that if the spec contains over reach, and it is my opinion that many specs unfortunately and irreversibly do (why is DI even included in JAX-RS for example) then it is not something that other implementations can viably support in the future and it will be the case that it is a spec with a single implementation (Quarkus) which has to be undesirable for all parties involved.

> Now -- I really thought that using generics for expressing basic type queries is compatible with annotation processors. I confess I didn't try. If not, we'll have to come up with something different.

Annotation processors can in fact reference classes, however you really don't want a massive annotation processor classpath so there needs to be a way to do the same thing without generics and not enforcing a user to depend on a concrete API. For example with annotation processors you can match annotation patterns which are strings like "foo.bar.*" 

> representations such as org.jboss.jandex.DotName or javax.lang.model.element.Name are terribly user unfriendly.

I agree those are unfriendly, although I understand the purpose as the name of a an inner class or array etc. is not the same source code vs byte code.

I would also say that having 2 ways to do things is confusing and just duplicating documentation effort. It would be better to just provide the fluent API and then be able to pass annotation / class names as strings if they don't exist on the classpath:

```java
    @Enhancement
    public void configure(AppArchiveConfig app) {
        app.classes()
            .subtypeOf("example.MyService")
            .configure()
            .stream()
            .filter(it -> !it.hasAnnotation(MyAnnotation.class))
            .forEach(it -> it.addAnnotation(MyAnnotation.class));
    }
```

The name `AppArchiveConfig` I would say is fairly unusual as well. Why is this an `AppArchive`? I know Quarkus builds this like Uber `Application` class but that doesn't mean other frameworks will take the same approach or use the same naming conventions.

rmannibucau:
Hi,

I read the blog post but I fail to see why we need yet another duplication of the API (we already have some duplicated~like API in the extension area).
Typically it already works at build time today, container can run them, capture the outcoming annotatedtypes (or even beans to be more optimized) and bypass extensions at run time == prebuilt CDI runtime mode.
All intermediates options are also available (dump scanning + reflection (made lazy at runtime) to speed up the run time boost and keep the full power of annotations, use full reflection in graalvm and keep all CDI SE/std features, etc...).
So currently I'm not sure if it will not split CDI communities and, to be very concrete, create a cdi-quarkus API since other vendors just made CDI supported in native mode.
I would be very cautious with such a profile which makes libraries working only in some CDI flavor and therefore reduce community - or more exactly create communities instead of having a community.
IMHO, this is not the point to have a standard API and once again, it does not bring anything for vendors and users in terms of new usages or use cases we can't do today AFAIK.

n1hility:
> > FWIW we aren't thinking of (or really wanting) Quarkus to be the "RI", just one compatible implementation. Would be great to have a micronaut impl and an API that can support them both (along with other impls)
> 
> Thanks @n1hility Nice to receive a rational response from a Red Hat engineer, and yes we are absolutely on the same page on that front. My desire is to support whatever CDI light is and that is a long term goal, but it has to be something that is supportable and not tied to a single implementation with a single architecture (byte code processing).

Awesome! Thank you for collaborating on this and sharing this valuable feedback and technical scrutiny, it will lead to a better standard and API. 


> Annotation processors can in fact reference classes, however you really don't want a massive annotation processor classpath so there needs to be a way to do the same thing without generics and not enforcing a user to depend on a concrete API. For example with annotation processors you can match annotation patterns which are strings like "foo.bar.*"

Ah I see. So in terms of referring to both strings and classes, in the scenario you have in mind, is it the extension author who decides (directly or indirectly) what's on that class path / dependency path is and whether they choose to refer to a type or not? So on the same CDI-Lite application, one extension in the application might just decide to refer to all strings, and another in the same might use types, possibly even the same type name just referenced differently?

n1hility:
> Typically it already works at build time today, container can run them, capture the outcoming annotatedtypes (or even beans to be more optimized) and bypass extensions at run time == prebuilt CDI runtime mode.

Unfortunately that would not be a compatible solution. The lifecycles of portable extensions and the runtime application are intertwined by design, and its common practice / expectation to pass state directly into the app. Since in a build time model extensions and the app run in different "VMs", that can't work. (Extensions are not designed to be serializable,  build and run don't even necessarily share the same class path, theres expectations around passing reflective types, etc)

manovotn:
> I am not sure I understand, what topic are you referring to? The desire to have an API that everyone can support seems pretty relevant to this.

We meant to have separate issues for separate discussion to avoid having several parallel discussions/ideas in one issue/PR.

> Please clarify where I accused anyone of anything or where I deviated from technical discussion, it seems the both of you are a touch emotional.

I had a similar impression from your initial comment. Looks like we both misunderstood each other :-)

> The name AppArchiveConfig I would say is fairly unusual as well. Why is this an AppArchive?

As was highlighted several times, names are up for changing. And this one in particular. We've had various ideas none of which really captured its purpose so this was the last attempt that just went public with the API. We need a better name but I don't really have any up my sleeve; it's supposed to be an entry point to the whole application, all possible classes that you might want to filter and transform for cases where going with just parameters in methods isn't enough.

> I would also say that having 2 ways to do things is confusing and just duplicating documentation effort. It would be better to just provide the fluent API and then be able to pass annotation / class names as strings if they don't exist on the classpath:

Hmm, you mean as in having the option to either have method parameters or just go via `AppArchiveConfig` (or however we rename it)? We actually had one, then the other and settled on both for convenience and because various use cases had various better fits. For complex cases, I'd definitely use `AppArchiveConfig`, for simply slapping annotation in top of one specific class, the parameter was way easier than performing my own filtering.

I do agree with the point of specifying classes as `String`, we didn't take that into consideration. I imagine we wouldn't be able to avoid something like the `DotName` abstraction in that case though...

rmannibucau:
> Unfortunately that would not be a compatible solution. The lifecycles of portable extensions and the runtime application are intertwined by design, and its common practice / expectation to pass state directly into the app. Since in a build time model extensions and the app run in different "VMs", that can't work. (Extensions are not designed to be serializable, build and run don't even necessarily share the same class path, theres expectations around passing reflective types, etc)

This is wrong, extension don't need to be serializable, the "post deployment state" (you can see it as the bean manager or annotated types + beans + config - interceptors, decorators, programmatic qualifiers etc...) must be and is. It is not a spec feature but a vendor feature since CDI is an API and does not assume anything between build and runtime specific envs - for the good, look at JPA which did assume with some annotation processor constraints and kind of failed to be used at build time in 90% of apps - but it just defines one API vendors integrate as they want while they respect the expected runtime. Once again it already works today.

Another big issue with such a "light" API is that it will quickly end up as requiring the same API as the standard one (I agree we can simplify it a bit due to the 1.0 -> 2.0 changes but this is not critical) except it will replace all the native reflection part by "copies" of these API (guess fields will get name to type as String, method will get name + parameters {String name,String type} + exceptions as strings etc....). Not a real gain. Now look at the runtime, if you get an interceptor, you still need reflection or to break the InvocationContext which would once again break library writing (or require yet another facade and drop CDI goal to be a primary API, in other words: CDI does not become the API but a technical stack which is a failure IMHO since it only makes sense as an API and not a vendor stack detail.

You drop the reflection but still need the meta to let the extension be useful so at the end you just redid what is done today: dump the state at build time and replace most of the reflection work done at runtime by a lookup in this state (a bit like CDS works).

mkouba:
> This is wrong, extension don't need to be serializable, the "post deployment state" (you can see it as the bean manager or annotated types + beans + config - interceptors, decorators, programmatic qualifiers etc...) must be and is. It is not a spec feature but a vendor feature since CDI is an API and does not assume anything between build and runtime specific envs...

@rmannibucau  Not really. See [11.5. Container lifecycle events](https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#init_events):
_"The container instantiates a single instance of each extension at the beginning of the application initialization process and maintains a reference to it until the application shuts down."_ In other words, the container must retain the extension instance used to deliver event notifications...

rmannibucau:
@mkouba yep but the same will apply to the light extensions - or will move to registered beans which is the same - so if you prefer this phrasing: "we can easily make today's spec build time compatible and portably" (today it is portable if the extension is serializable somehow, used some json but it was extension specific). So still, the step to make CDI "phasable" (build -> run) is easier and more consistent than the step to fork CDI core (extensions/lifecycle) to create another spec, this is the whole point. Today it is often compensated by having a build scanning  phase and dump it in a bean able to read this scanning result....but nothing to put in the spec IMHO, in particular in annotation processors which are always a pain for users from what I saw.

graemerocher:
> Ah I see. So in terms of referring to both strings and classes, in the scenario you have in mind, is it the extension author who decides (directly or indirectly) what's on that class path / dependency path is and whether they choose to refer to a type or not? So on the same CDI-Lite application, one extension in the application might just decide to refer to all strings, and another in the same might use types, possibly even the same type name just referenced differently?

@n1hility Yes, for example it is maybe ok to refer to references when you are processing a stable API JAR that contains only interfaces and no third party dependencies, but you wouldn't want to reference say Hibernate which has many third party dependencies and maybe introduce conflicts. 

However even in the case of a "stable" API you can have changes. For example look at the `javax.*` to `jakarta.*` transition. In Micronaut we have no problem handling this because we just map one API to the other without reference either directly (only user code should)

> `AppArchiveConfig` is supposed to be an entry point to the whole application

@manovotn So again this is an area where implementation may differ since for example Micronaut does not dictate your application entry point. We do build a registry of compile time computed beans but that is separate from the entry point. I would say if this is going to be generic a name like `ApplicationConfiguration` makes more sense.

manovotn:
> @manovotn So again this is an area where implementation may differ since for example Micronaut does not dictate your application entry point. We do build a registry of compile time computed beans but that is separate from the entry point. I would say if this is going to be generic a name like ApplicationConfiguration makes more sense.

It is in fact similar to existing CDI extensions and the usage of wildcard in lifecycle method parameters. For instance one pretty common (and poorly performing) case was to use `ProcessAnnotatedType<?>` and then do the whole filtering yourself. This class allows you to do very similar thing should you need that.

I don't see it as dictating any entry point. You needn't use this as one; you can use method parameters giving you precisely the classes/fields/method you want to modify.

Also, `ApplicationConfiguration` as a name isn't bad but it is also very broad. It isn't really whole application, it's just the CDI part of it that it grants you access to.

m0mus:
Mailing list discussion: https://www.eclipse.org/lists/cdi-dev/msg00048.html

Ladicek:
So what I wanted to move into a separate issue is concerns about annotation processors compatibility. There doesn't have to be a discussion that a CDI Lite extension mechanism, however we call it, must be implementable with annotation processors -- of course it must, that's a given. I would like to have a technical discussion about that topic and possible trade-offs it might bring. But if we have it here, well, then we have it here.

The present objection, if I understand correctly, is that putting the types [to which extensions refer] to compilation classpath can lead to compilation problems due to dependency misalignment. (Please correct me if I'm wrong.) To that, I'd say that you need to align those dependencies anyway, because you need to run the correct version of the extension. If you integrate Framework X Version 1.0 using an extension, you need to use an Extension For Framework X Version 1.0. Running an extension for a different version of the integrated framework is likely to lead to runtime problems, which I'm sure you'll agree is worse than compilation problems. So, again, those dependencies must be aligned anyway -- in which case, putting them to the compilation classpath seems like a minor thing to me. -- If there are other reasons to not put dependencies to compilation classpath, I'd like to learn about them.

(In otherwords, yes, I indeed made an assumption that classes referred to by an extension are available. I thought that's natural because of the above. I most definitely did not make an assumption that "all implementations of this API are Quarkus".)

Now, I can imagine a `String` based API, sure. For example, we could replace

```
@WithAnnotations(com.example.MyAnnotation.class) Collection<ClassConfig<? extends com.example.MySevice> service
```

with something like

```
@WithAnnotations("com.example.MyAnnotation") @SubtypesOf("com.example.MyService") Collection<ClassConfig> service
```

and so on. That could be conceptually relatively close to the current proposal. (Or we can have something that is conceptually very different from the present proposal. That's good too.) I'd argue that such API is harder to use ("stringly typed" vs. "strongly typed" and all that), but this is certainly a discussion we can have.

(When it comes to `AppArchive`, again, that doesn't come from Quarkus. I originally wanted to call it `BeanArchive`, because CDI already has a concept of bean archive and I thought it's very close. I later learnt that it's not, so I went with `AppArchive`, and there's a corresponding `AppDeployment`. I'd very much love to have a different name. `Application` and `ApplicationConfig` would be great, but I don't know how `AppDeployment` should be called in such case.)

graemerocher:
> To that, I'd say that you need to align those dependencies anyway, because you need to run the correct version of the extension.

There are multiple reasons to keep the compilation path and annotation processor path separate that have been established over many years since annotation processing is a mature model used by many frameworks and tools (in particular on mobile where Android SDK is almost entirely based on them).

Having a larger compilation classpath and/or annotation processor path will impact compiler performance since you have more jars to load against and scan. The Gradle engineers have done a lot of work on making annotation processing support incremental compilation and this again requires that the annotation processor path is kept to a minimum.

In terms of alignment normally it is better to have libraries separate their annotation jars (with a `-annotation` artifact that contains only the annotations) from API which is then further separated from the runtime dependencies.

It is not the case that you need the annotation processor classpath and the runtime classpath to be aligned, in fact Micronaut has dozens of modules none of which get included on the annotation processor classpath and is testament to the fact that this is in fact not necessary.

If you you decide ok I want to include the API on my extension classpath, extension authors would have to take doubly great care about transitives to ensure that 2 APIs from 2 different extensions don't transitively depend on conflicting versions of something like Jackson for example. Just look at all the transitives hibernate pulls in from dom4j, to javassist, bytebuddy etc. why would you want those on your compilation classpath and risk a potential conflict that would result in a user having to remote debug a compiler?

At least when conflicts occur at runtime you can use regular debugging techniques to analyze and resolve them.

Ladicek:
Thanks, that's very good info, it's great to learn about this. As I said, I'm no expert on annotation processors, so happy to have your input.

For an extension integrating a framework, transitive dependencies of the framework are typically not required, not even all the framework classes I think. I can see how assembling the right classpath can be a pain, but perhaps that can be automated somehow? (Not sure, pure speculation.)

When I mentioned runtime errors, I'm not thinking just classloading errors, but also logical errors stemming from the framework being incorrectly wired in. Interceptors not running, wrong types being injected, things like that.

m0mus:
Folks, I as suggesting to move this discussion to the mailing list. If multiple discussion places are used it will be very difficult to find what's been decided and when. IMO, mailing list is (and historically has been) a proper place for spec discussions.

m-reza-rahman:
I also believe the mailing list is the best place for this discussion. The GitHub discussion is really hard to follow or filter since I contribute to many things and essentially work asynchronously from my inbox.

Reza Rahman
Jakarta EE Ambassador, Author, Blogger, Speaker

Please note views expressed here are my own as an individual community member and do not reflect the views of my employer.

tomas-langer:
I have started a thread (+ associated doc) for discussing the future of CDI here: https://www.eclipse.org/lists/cdi-dev/2020/Oct/index.html to describe (and discover) the problems, reasons, and use cases that drive the need to change CDI spec.

manovotn:
We should use and work on both resources, the document has several parts that directly relate to what we tried to address here. Such as the extension SPI or the metamodel fot for AOT approach (i.e. reflection free).

api/src/main/java/cdi/lite/extension/beans/BeanInfo.java
```diff
@@ -0,0 +1,55 @@
...
+
+    boolean isClassBean();
+
+    boolean isProducerMethod();
+
+    boolean isProducerField();
+
+    boolean isSynthetic();
+
+    MethodInfo<?> producerMethod(); // TODO null if not producer method, or return Optional?
```
otaviojava, line 37:
```suggestion
java.util.Optional<MethodInfo<?>> producerMethod();
```

api/src/main/java/cdi/lite/extension/beans/BeanInfo.java
```diff
@@ -0,0 +1,55 @@
...
+
+    FieldInfo<?> producerField(); // TODO null if not producer field, or return Optional?
```
otaviojava, line 39:
```suggestion
java.util.Optional<FieldInfo<?>> producerField();
```

api/src/main/java/cdi/lite/extension/model/declarations/ClassInfo.java
```diff
@@ -0,0 +1,59 @@
...
+import cdi.lite.extension.model.types.Type;
+import cdi.lite.extension.model.types.TypeVariable;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * @param <T> the inspected class
+ */
+public interface ClassInfo<T> extends DeclarationInfo {
+    String name();
```
otaviojava, line 12:
```suggestion
    String getName();
```

otaviojava: I would recommend using the same standard of the class:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getName()

api/src/main/java/cdi/lite/extension/model/declarations/ClassInfo.java
```diff
@@ -0,0 +1,59 @@
...
+
+    String simpleName();
```
otaviojava, line 14:
```suggestion
    String getSimpleName();
```

otaviojava: The same reason:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getSimpleName()

api/src/main/java/cdi/lite/extension/model/declarations/ClassInfo.java
```diff
@@ -0,0 +1,59 @@
...
+
+    PackageInfo packageInfo();
```
otaviojava, line 16:
```suggestion
    PackageInfo getPackageInfo();
```

otaviojava: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getPackage()

api/src/main/java/cdi/lite/extension/model/declarations/ClassInfo.java
```diff
@@ -0,0 +1,59 @@
...
+
+    List<TypeVariable> typeParameters();
```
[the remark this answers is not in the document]
otaviojava: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getTypeParameters()

api/src/main/java/cdi/lite/extension/model/declarations/ClassInfo.java
```diff
@@ -0,0 +1,59 @@
...
+
+    Type superClass();
```
otaviojava, line 20:
```suggestion
    Type getSuperClass();
```

otaviojava: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getSuperclass()

api/src/main/java/cdi/lite/extension/model/declarations/ClassInfo.java
```diff
@@ -0,0 +1,59 @@
...
+
+    ClassInfo<?> superClassDeclaration();
+
+    List<Type> superInterfaces();
```
otaviojava, line 24:
```suggestion
    List<Type> superInterfaces();
```

otaviojava: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getInterfaces()

api/src/main/java/cdi/lite/extension/model/declarations/FieldInfo.java
```diff
@@ -0,0 +1,30 @@
+package cdi.lite.extension.model.declarations;
+
+import cdi.lite.extension.model.types.Type;
+
+/**
+ * @param <T> type of whomever declares the inspected field
+ */
+public interface FieldInfo<T> extends DeclarationInfo {
+    String name();
```
otaviojava, line 9:
```suggestion
    String getName();
```

otaviojava: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/reflect/Field.html#getName()

api/src/main/java/cdi/lite/extension/model/declarations/FieldInfo.java
```diff
@@ -0,0 +1,30 @@
...
+
+    Type type();
```
otaviojava, line 11:
```suggestion
    Type getType();
```

api/src/main/java/cdi/lite/extension/model/declarations/MethodInfo.java
```diff
@@ -0,0 +1,45 @@
...
+import cdi.lite.extension.model.types.Type;
+import cdi.lite.extension.model.types.TypeVariable;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * @param <T> type of whomever declares the inspected method or constructor
+ */
+public interface MethodInfo<T> extends DeclarationInfo {
+    String name();
```
otaviojava, line 12:
```suggestion
    String getName();
```

api/src/main/java/cdi/lite/extension/model/declarations/MethodInfo.java
```diff
@@ -0,0 +1,45 @@
...
+
+    List<ParameterInfo> parameters();
```
otaviojava, line 14:
```suggestion
    List<ParameterInfo> getParameters();
```

api/src/main/java/cdi/lite/extension/model/declarations/MethodInfo.java
```diff
@@ -0,0 +1,45 @@
...
+
+    Type returnType();
+
+    // TODO return Optional<Type> and only return non-empty if receiver parameter is declared,
+    //  or return Type and always return a receiver type, even if not declared (and hence not annotated)?
+    Optional<Type> receiverType();
+
+    List<Type> throwsTypes();
+
+    List<TypeVariable> typeParameters();
```
otaviojava, line 24:
```suggestion
    List<TypeVariable> getTypeParameters();
```

api/src/main/java/cdi/lite/extension/model/declarations/PackageInfo.java
```diff
@@ -0,0 +1,18 @@
+package cdi.lite.extension.model.declarations;
+
+// TODO is this useful? perhaps `ClassInfo.packageName` returning a `String` would be enough?
+public interface PackageInfo extends DeclarationInfo {
+    String name();
```
otaviojava, line 5:
```suggestion
    String getName();
```

rmannibucau:
Hi guys,

It looks like the proposed build time API is equivalent to the extension API so why not just running extensions at build time if it is the goal (since it has the same limitation at the end) and bypass them at runtime.
Whatever way you see it, this thread is not about really changing the spec in terms of features but just changing the impl which is enabled today and almost all vendors already do it (so it does not make sense to duplicate an existing API).
It fully works with CDI SE and does not need to bring to the spec build concerns which have big drawbacks for end users (like not working in IDE in 95% of cases even if they claim it should) and breaking common dev lifecycle.

Side note: once again also note that CDI is fully GraalVM friendly so this proposal does not help, it just limits CDI a lot and breaks all its ecosystem since not a single lib will work with this api anymore.

Indeed just my 2cts but I think it is worth investing all our time in something bringing at least a feature to end users.

Ladicek:
Hi @otaviojava, thanks for taking the time to fill in all those suggestions. I personally think that using a simple noun as a method name is a lot more readable than the verb phrase with `get` (see also Effective Java, 3rd edition, item 68). If we were to follow the API conventions of `java.lang.Class`, then we should prefer returning `null` over returning `Optional`, too. I'm still not quite sure what conventions to use, so this is definitely a good feedback, please keep it coming :-)

Hi @rmannibucau, few things re your comment. One, current CDI specification allows to (and people do) access extension instances at application runtime, so I don't really think you can just "run extensions at build time and bypass them at runtime". Moreover, the extension API relies heavily on reflection _and_ on having a [part of] CDI container running to deliver events, both of which you really want to avoid during build time. Which leads me to -- two, the point isn't to "move extensions to build time" or "make CDI [implementations] GraalVM friendly", both of which seem to be popular misconceptions. The point is to [allow implementations to] move _the entire bean discovery process_ to build time. Which of course includes extensions. This enables significant runtime speed and memory gains that you can't get otherwise. I think there's plenty of evidence out there that this is very useful for end users. But, three, everyone should of course feel free to propose their own enhancements. I'd be happy to discuss any proposal you come up with.

rmannibucau:
@Ladicek to be very concrete I see less issue to have a SerializableExtension (extends Extension { void write(OutputStream);void load(InputStream); }) which would filter extensions for the build time as described compared to current proposal which duplicates 1-1 java.lang.reflect and breaks CDI API for the whole ecosystem (in other words it is NOT a CDI enhancement but the creation of a new spec which should belong to another project IMHO). Note that it already works well while extensions are serializable which can be the sane limitation of such a feature compared to creating a while new API which breaks development and tools (btw give a try with lombok+mapstruct in the same project, you'll need unatural workarounds).
Also you say:

> Moreover, the extension API relies heavily on reflection and on having a [part of] CDI container running to deliver events, both of which you really want to avoid during build time. 

Which is not true since the proposed API is *exactly* that whatever you look at it, I just agree it is not CDI but another container spec.

Other thing is that

> The point is to [allow implementations to] move the entire bean discovery process to build time.

Is *already* doable and done by user vendor scanning services. Since the proposal is mainly about annotation propcessors or things like that, standardizing it will *not* help since it does not impact user API.

However note that your

> Which of course includes extensions.

*breaks* CDI which must be configurable at runtime, including in GraalVM or precomputed runtimes so here you will start to have ambiguities - and once again it would belong to another IoC like dagger.

> This enables significant runtime speed and memory gains that you can't get otherwise.

Do you have figures? Today a real CDI app - with extensions, interceptors and so on - startup is [3;10]ms max depending the computer (on an i7/SSD scanning is ~1ms for the same app), do you expect to gain 1ms? With a HTTP server the I/O already takes most of the time (~20ms) so if you add a database it is already pointless.

>  I think there's plenty of evidence out there that this is very useful for end users.

Do you have a few? But if you review all the JPA, lombok and mapstruct annotation processor usages over stackoverflow, twitter etc you will see there are plenty of evidence of drawback and reject of that by users.

So overall, being able to capture the deployment before the runtime can be a feature even if I don't see it required nor expected since 1. it will limit CDI and break extension runtime capacities, 2. does not bring real performance boost in GraalVM mode. In JVM mode you can get a perf boost but it means you replace the O(N) extension by pre-generated beans (all containers can exclude undesired extensions through their SPI). This lead me to the fact you probably want to consider *one* of both options:

1. the serializable extension mecanism
2. adding a Filter (Predicate<Extension>) SPI to be able to control which extensions are in or not

Both options sound saner to me than duplicating java.lang.reflect and preventing people to use CDI extensions.

graemerocher:
@rmannibucau It is odd that you are bringing this up here when there have been discussion and discussion over these various topics on the mailing list and at the various open spec discussion meet-ups that happen weekly. 

I recommend you participate there since many of the things you raise have already been addressed, multiple times.

With regards to your comment first off the build time classpath != runtime classpath and for build time you want to isolate a smaller classpath which just includes build time extensions and never load users classes. Loading user classes and extension classes inside a build step or compiler step is asking for trouble and hence the proposal to avoid java.lang.reflect

It is troublesome for a user to debug dependency conflicts at runtime that occur such as a NoSuchMethodError, the last thing you want as a user experience is for the user to have to attach a remote debug to their compiler.

The proposal so far as is not preventing anyone from doing anything. If you want full runtime extension support just use an implementation of CDI that provides that, if you don't then users have choice to rely just on build time services to provide what they need. 

The proposals here are trying to help CDI by improving adoption and compatibility so that more frameworks (and two of the most popular ones used by Java developers to build modern applications ie Quarkus and Micronaut) can support a profile or version of CDI that works at build time.

Without efforts to standardise this then CDI will continue to be broken and fractured anyway since neither Quarkus nor Micronaut will be able to achieve spec compliance. If that is what the CDI community as a whole wants then so be it, and CDI can continue to fade into irrelevance as a standard.

otaviojava:
@Ladicek 
Thank you, I'll organize myself to participate in the meetings as well.
I'll try hard to be useful as possible.
I believe that it returns null instead of optional because the Class is older than Java 8.
My point is, once both Java SE and Jakarta EE are platforms, I believe it should have a closer line up.

Ladicek:
I think @graemerocher summed it well, but I'll add that the gains we seek to enable are also important for regular JVMs (where, granted, they are easier to see). Again, this isn't only about AOT compilation to native binaries.

graemerocher:
@rmannibucau You seem to be expert on Micronaut and Quarkus' architecture already and somehow know for a fact that these projects could have supported existing runtime extensions if they wanted to. If you had been party to the discussion on the mailing list ( and I suggest you go back and read the threads) you will have seen and read where it has been explained multiple times that the existing runtime extension model is a non-starter for build time approaches.

Using reflection is a non-starter, unless as you say you come up with a solution that provides a bad user experience. So your suggestion is that we should standardise on an existing API whereby it will be confusing to the user what and what is not possible at build time and whereby the build step will load user code and result in potentially difficult to debug problems in order to maintain strict backwards compatibility? 

I can tell that it is not even viable from a Micronaut perspective and will let the Quarkus folks express their opinions on that.

rmannibucau:
@graemerocher all the justifications I read - and I spent time trying to see the justification since it is *done* - are not technical but philosophical or developer driven. I can hear bypassing reflection can be a target but it does not mean changing current model + annotation processor driven experience is way way worse than reflection at build time. Lastly the most blocker point for me is to create a new IoC under CDI umbrella which shouldn't even be an option for anyone IMO. If you manage to fix these two blockers - which BTW requires very little changes to the API compared to current proposal - I would support it, but in current form the proposal does not make any sense for users, only for vendors which shouldn't be until jakarta decides to be an implementation and no more a spec which should be stated explicitly.

graemerocher:
@rmannibucau All I am hearing from your side is a lot of opinion and eagerness to stick the status quo which makes absolute sense given you work for a vendor.

The experience Micronaut users have with Micronaut and its popularity (nearly 5k GitHub stars vs 50 for OpenWebBeans?) does not reflect your statement that "annotation processor driven experience is way way worse than reflection at build time" besides that the spec revision does not mandate the use of annotation processors (Quarkus in fact doesn't use them).

Having said that the process to standardise this is not set in stone and as it stands still quite far from reality, plus it is not going to be dictated by you or me but by a group of individuals working together to see what is possible. I suggest you join that process and try to make a contribution to it, rather than just making statements that are based in your reality but not necessarily the reality that users have with these respective tools.

rmannibucau:
@graemerocher please note I don't work for a vendor, not sure where you took that but it is wrong. Also don't use github stars as a criteria since it is - as you know - completely biased, in particular since primary work area for a lot of ASF projects is not github,  and if you want to use that criteria anyway, add all containers using openwebbeans as well to be fair since micronaut and OWB are not strictly the same kind of project in terms of scope. Lastly note that I'm not opposed to the proposal as "it", I just strongly think it is against CDI and does not belong to CDI at all, it is trivial now (@eclipse) to create a new project importing the CDI API you need as a dependency and adding this new API. Will enable you, quarkus and whoever wants, to use it, let users validate it is working or not and later on merge it with CDI if desired but as of today I see two big blockers and danger for CDI to import it with this design. Don't forget the overlap between this API and CDI is very little compared to the overall CDI API. Indeed CDI can be broken in "extension", "runtime", "se", "build" subspecs but it will have the same pitfalls I described so I think it is saner to play with this build proposal outside of CDI *for now*.

graemerocher:
@rmannibucau well that is exactly what is happening, there is no expectation for this standard to be formalised this year. The Quarkus and Micronaut teams are working towards building proof of concept reference implementations that validate the concepts proposed here completely outside of the main CDI specification. It will be lucky if we complete that this year and if we do it can be revisited what if anything can be standardised.

Standardising without validating by creating 2 implementations is madness. This is all still a long way off being a reality so there is no need to panic in the short term.

As for the Github stars, that was in defence if your unfounded statement that annotation processors provide a bad user experience which is contrary to the evidence out there where many many developers worldwide have great user experiences with things like dagger, mapstruct, micronaut and yes even Lombok. The fact that people are asking questions about these things on stack overflow is a sign they are actually being used by people which is a good thing. If there were no questions that is when you have to be worried.

rmannibucau:
@graemerocher ok so why not creating a jakarta-sandbox @eclipse and continue there? Would let this bugtracker dedicated to alreayd targettde discussions, no?

About github stars: it is not unfounded but JPA + lombok (independently) feedbacks for years (last ~7 continuous years from my experience). Experience is generally ok-ish while you have a single one but as soon as you have 2 changing the model (like generating sources or even modify classes to add container SPI/extension points) it breaks this fluent experience and it is likely such an annotation processor generates some source at some point. Just look the mapstruct-lombok SPI introduction to have an idea of that, hurted for years and was only recently fixed. The fact they are used is because often you don't have the choice (clearly with mapstruct and jpa for the metamodel), in most cases if it can be avoided it is. The issue about the question is that it just does not work compared to a runtime solution and makes time lost and setup more complex requiring to not use the IDE anymore or loose some of its feature. This is my big worry point about build time solutions (even quarkus solution which is not an annotation processor can have issues in IDE). Concretely as soon as you quit compile -> run test flow you can have troubles. Annot processors seems promishing since it respects that but their model and lifecycle induce other issues so overall runtime if metadata are not already generated solutions (which kind of combine best of both worlds) are the best compromise as of today - can change if javac enables to add custom build phases or post compile phases portably maybe but it is not yet a feature.

graemerocher:
Lombok is a special case and shouldn't be grouped with other annotation processors because it violates the API provided by Java for annotation processors by using internal classes and mutating the underlying AST of the Java and Eclipse compilers. This approach was simply looking for trouble so hence why I don't group that processor together with others and hence why developers have run into issues with.

Others like mapstruct, dagger and micronaut have no such issues. In the case of Micronaut we don't even generate Java source code. 

As for your questions, these are suggestions that you should absolutely bring to the group of the mailing list. Currently as far as I can see this is merely a PR in a private repository. There is no threat of being merged and it is being used for discussion.

struberg:
> @rmannibucau It is odd that you are bringing this up here when there have been discussion and discussion over these various topics on the mailing list and at the various open spec discussi

@graemerocher Romain, I and a bunch of other people gave this feedback since almost a year. Then I was accused to only ship stawmen arguments, etc. I replied less often, because all is said. Point is: it is technically not really necessary to kill the existing CDI API as many things it provides can be moved to build time. Apache OpenWebBeans start up in 5 ms with build time extensions. Helidon solved the problem as well. So why would it be a good idea to break backward compatibility with former versions of CDI? Could be done of course, but it's not CDI anymore.

rmannibucau:
@graemerocher I agree that lombok and a bunch of other processors are "special" (even vioate the annot proc contract) but it is the world we are in and it will not be fixed anytime soon so a new annot proc would be broken as well, you can hope prople only use the stack you provide but it is generally combined and at some point broken this way so it should be accepted IMHO and therefore the solution not considered as sane About the repo: PR in a private repo is fine until it becomes a PR which gets started to be reviewed (as any PR should) so maybe close the PR for now if it is how you see things?

Ladicek:
I have marked the PR as a draft, because it certainly isn't supposed to be merged as is, but it is also supposed to be discussed.

Though at this point, I might have also just closed this, because the current proposal that we debate on the calls is somewhat different already. I'll think about that, but in the meantime, I'll keep it as a draft. I agree that makes more sense.

Ladicek:
I have pushed a thoroughly revised initial proposal for Build Compatible Extensions API, corresponding to the current state of the https://github.com/Ladicek/StillDI project, updated for the `jakarta` namespace. I will shortly update StillDI to use this new API (up to now, it's been based on CDI 2.0 and hence the `javax` namespace).

I have also went through all the files in this PR and added a number of TODO comments -- mostly to highlight all the places where I wasn't sure of something, or where I know there's more work required. I have, for now, intentionally abstained from making larger changes, even though I'm convinced of some (e.g. removal of `AppArchive[Config]` and `AppDeployment`). This is because I believe (and we discussed this on the weekly CDI call) that at this point, we need to have some baseline that other implementers can start using.

The current number of TODOs is 88, and they are far from minutiae (there's e.g. a TODO for the name "build compatible extensions"), which should clearly show that this isn't me throwing complete stuff over the wall and asking everyone else for ratification, but a genuine call for participation. Constructive feedback would be very much welcome (with understanding that present discussion has many examples of constructive, non-constructive and counter-constructive feedback, so let's try to be better this time).

manovotn (approved):
LGTM.

Let's get this into the repo so that we can roll out some Alpha and start improving it iteratively.

Ladicek:
Agree. Merging this is the beginning, not the end. Thanks!

graemerocher:
@Ladicek @Ladicek if there is a way to install it locally it is probably fine without an alpha so we can iterate. Is it just `mvn install`?

manovotn:
@graemerocher `mvn install` will compile and install the artifact into your local Maven repository. Which means your local build will work but it is still pretty inconvenient for CI code in GH repos etc.

Anway, I would welcome an Alpha release even for Weld. It shouldn't be an issue either; I just wanted to wait until today meeting to see if anyone has issues with it. If not, I should be able to release the artifact tomorrow :-)

manovotn:
BTW I think any committer of this project should be able to access and execute the Jenkins job that does a release (actually two jobs, staging and then repo release), so you wouldn't even be blocked waiting on me should you need more iterations out there.

graemerocher:
ok sounds reasonable

rmannibucau:
Hi,
Is it possible to NOT get alpha under jakarta but org.eclipse.jakarta.experimental package maybe?
Really want to avoid for it to look like something which will go to next release while there is no agreement nor technical need of that.

manovotn:
> Is it possible to NOT get alpha under jakarta but org.eclipse.jakarta.experimental package maybe?

Not really, experimental packages tend to do more harm than use. Both, for implementors and anyone actually trying to early adapt and test new APIs. Besides, we are pretty sure an API like this will be needed and present with next version of CDI for Lite. Of course this isn't a final version but we need a draft out so that implementors can try our its limits and usability.

>  Really want to avoid for it to look like something which will go to next release while there is no agreement nor technical need of that.

So, I suppose by the so called "no agreement" you mean CDI Lite and the presence of build compatible extensions within it. This has been agreed on by committers (and community) in a vote started by this email - https://www.eclipse.org/lists/cdi-dev/msg00361.html

Furthermore, this SPI is being iterated on continuously during CDI meetings and was also discussed multiple time on PRs and on mailing list. Both future implementors (and not just them) agree that we need an SPI like this to supplement CDI extensions in a build time environment. In fact, this SPI was drafted exactly because there was a technical need for an equivalent of portable extensions that would work in build time environment - e.g. that would work in Lite.

rmannibucau:
> we are pretty sure an API like this will be needed and present with next version of CDI for Lite

There is clearly no agreement on that point and a soft one can have been found due to the investment requirement to push against "big vendors" but at the end it is not a good decision for the community.

Technically it is proven that native applications are not going to be mainstream because:

1. In java land it is a small amount of application which are concerned
2. It does not match EE/Jakarta design much (short lifetime like CLI apps)
3. It is a huge build investment which rarely pays in the cloud or on premise (setup and build cost is way >> gained run in 100% of cases since the gain only pays in lambda case which is not targeted by EE nor adopted)
4. As soon as you have a remote connection (database for ex) it is not worth at all,
5. It brings build time constraint in a runtime spec which is broken by design for jakarta scope since jakarta must not provide build tools - to dev apps,
6. It does not enable any use case since CDI is build time friendly *already* as proven by most vendors and its design enables to not use reflection under a few constraints *already* - same as this lite proposal - so a duplicated API does not enable anything for end users except it adds noise and prevent to use the existing ecosystem of the related spec (you can't use a CDI extension with CDI lite which from user point of view is insane).

So this flavor of CDI "Lite" is just unexpected and a lack of knowledge of what is already there in 99% of apps and therefore shouldn't hit jakarta package IMHO.

manovotn:
> There is clearly no agreement on that point 

There is and your denying of it won't change it, sorry.

> So this flavor of CDI "Lite" is just unexpected and a lack of knowledge of what is already there in 99% of apps and therefore shouldn't hit jakarta package IMHO.

Right, which is why all the committers acknowledged the need for it and voted +1?

> Technically it is proven that native applications are not going to be mainstream because:

Your knowledge of the whole Java ecosystem and your near-prophetical abilities to foresee the future course and needs of development are truly astonishing but I don't see any hard evidence supporting them so I won't even comment on them one by one.

The project still aims to support Jakarta EE as it is (a runtime oriented setup) but that doesn't need to prevent it from having variant(s) that can be used outside of it. CDI can evolve is multiple ways and in order to do so, an agreement of majority of committers is needed which, after a very long period of discussions, was achieved.

Ladicek:
I'd like to point out one thing, too. It seems to be a popular misconception that the main goal of Lite is to enable native compilation. It is not. The main goal is to enable moving bean discovery and related tasks (such as running extensions) to build time. This is a popular implementation strategy for AtInject (see e.g. Dagger) and I can't see why it shouldn't be possible for CDI.

rmannibucau:
@manovotn :
> all the committers acknowledged the need for it and voted +1

Well it is clearly not the case and I didn't get so positive feedbacks from most people I asked about so once again it is a very biased vision we should find a way to make it less vendor oriented.


> your near-prophetical abilities to foresee the future course

I'm actually speaking of what happent, and that the enthusiasm of the beginning are now down, nothing prophetical compared to this proposal of new API without a need.

> The project still aims to support Jakarta EE as it is (a runtime oriented setup) but that doesn't need to prevent it from having variant(s) that can be used outside of it. 

Actually no since CDI support will need to cover both - this is why I asked why not doing another project since the proposal is CDI unrelated at the end. Once again I'm not against trying this, I am against doing it in CDI scope.

@Ladicek :
> The main goal is to enable moving bean discovery and related tasks (such as running extensions) to build time.

Then you don't need any API since it is already doable with all implementation without a single change. So you are right there is no reason it is not done for CDI 2.0 but no need to reinvent the API from scratch too.

ljnelson:
> all the committers acknowledged the need for it and voted +1

That is incorrect.  For the record I deliberately did not vote +1 for many of the reasons @rmannibucau outlines above.  (The outcome of the vote, of course, is not in question.)

Respectfully: let's keep the tone professional here.

manovotn:
> That is incorrect.

Not really, there is lazy consensus applied to votes, as described here - https://github.com/eclipse-ee4j/cdi/wiki#decision-making-process
Anyhow, I meant all actively participating in vote anyway and I get your point.

However, the vote is there precisely for people to express their thoughts, regardless of whether they vote one way or another. And with all respect, it would be valuable if people did that instead of coming up weeks later on PRs which are not nearly as visible as mailing list not to mention by that time the vote has concluded. It is hard to incorporate such feedback is all I am saying :-)

n1hility:
I think we should move on from the issue 

> > all the committers acknowledged the need for it and voted +1
> 
> That is incorrect. For the record I deliberately did not vote +1 for many of the reasons @rmannibucau outlines above. (The outcome of the vote, of course, is not in question.)

Overall the vote was very positive +12 / (+6 committers), and no -1s. It's beneficial to open the door to new implementations, there is no real negative to those that aren't interested in Lite, and the spec restructuring is going to lead to an overall better / cleaner cleaner spec.  Sure there may not be perfect unanimous agreement but that's ok. If we always agreed the same on everything we wouldn't have a standard in the first place, since everyone would have agreed on the same implementation. A good standard will allow for flexibility, and I think this was accomplished.  We should work together to move forward and continue the spec's improvement.

rmannibucau:
Guys, just move forward *outside* CDI in a eclipse-ee4j/cdi-lite or whatever you want to name it but putting it inside CDI brings a ton of issues as soon as you start to use *CDI* as a spec (kind of the same as at inject implementations have together) which is more than unlikely to serve anyone at all except in terms of communication which is pretty bad for end users.
So yes move forward if you want but please keep the high value of EE inheritance which is to NOT break users when not needed.

PR #455, opened 2020-10-15
Adding support for PPC64LE CI

ezeeyahoo:
Hi, 

Please include these changes that enable CI for PPC64LE.

Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,18 @@
 language: java
 sudo: false
+
+matrix:
+  include:
+    - arch: ppc64le
+      addons: [ apt: [ packages: maven ]]
+      jdk: openjdk8
+    - arch: ppc64le
+      addons: [ apt: [ packages: maven ]]
+      jdk: openjdk11
+
 install: mvn -Pstaging install -DskipTests=true -B -V
 script: mvn -Pstaging test -B
+
 jdk:
   - openjdk8
   - openjdk11


manovotn:
Hi,

would you please care to explain why we need to test CDI specification with this architecture?
I don't see any benefit - or to put it another way, I don't see a reason why it should fail on that architecture as opposed to standard linux build that we already do.

ezeeyahoo:
@manovotn 
This project happens to be a part of the Red Hat distributions, including their ppc64le builds. We are enabling upstream CI for those to help reduce the cost of maintaining/debugging issues and to ensure the port stays current on ppc64le. We are trying to make sure that top of tree is always functional on ppc64le. So, the goal is to enable and validate on ppc64le - the Linux on Power Little Endian releases.

Now if we look at PPC build log here, it is not mirror of AMD64 build log. Moreover, I will be writing build script for building on local/VM systems  https://github.com/ppc64le/build-scripts.

Therefore, it would make sense to add this architecture.

manovotn:
Sorry, I am still not seeing the link here.
CDI is a community project which is consumed by all kinds of architectures and systems but it remains the responsibility of those projects and products to do the appropriate testing for architectures they want to support.
We cannot cover every possible matrix combination of architecture and java version on all eclipse projects. That would unnecessarily drain the CI with zero or low benefit - especially given that we are talking a Java/Jakarta specification which is mostly a matter of API and documentation.

I suggest we close this PR, cc @starksm64

Issue #461, opened 2020-12-17
A manually set configuredProvider can be lost for good. 


benjamin-confino:
The changes made in #445 means that a manually set configuredProvider can be lost for good if it ever returns null and it is registered in the ServiceLoader.

manovotn:
Fixed by https://github.com/eclipse-ee4j/cdi/pull/462

PR #462, opened 2020-12-17
Throw an exception without setting configuredProvider to null when ge…

benjamin-confino:
…tCDI returns null

Signed-off-by: Benjamin Confino <benjamic@uk.ibm.com>

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
@@ -44,6 +44,7 @@ import java.util.TreeSet;
 public abstract class CDI<T> implements Instance<T> {
 
     private static final Object lock = new Object();
+    private static volatile boolean providerSetManually = false;
     protected static volatile Set<CDIProvider> discoveredProviders = null;
     protected static volatile CDIProvider configuredProvider = null;
 
@@ -77,7 +78,11 @@ public abstract class CDI<T> implements Instance<T> {
             if (configuredProvider != null && configuredProvider.getCDI() != null) {
                 return configuredProvider;
             }
-        } catch (IllegalStateException ignored) {
+        } catch (IllegalStateException e) {
+            //if the provider is set manually we do not look for a different provider.
+            if (providerSetManually) {
+                throw e;
+            }
         }
         configuredProvider = null;
         // Discover providers and cache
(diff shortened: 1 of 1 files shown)


benjamin-confino:
This resolves #461 see also the discussion in #445

Emily-Jiang (approved):
LGTM!

api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
```diff
@@ -44,6 +44,7 @@
 public abstract class CDI<T> implements Instance<T> {
 
     private static final Object lock = new Object();
+    private static boolean providerSetManually = false;
```
manovotn, line 47: I think this should be `volatile`.

api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
```diff
@@ -79,6 +80,10 @@ private static CDIProvider getCDIProvider() {
             }
         } catch (IllegalStateException ignored) {
         }
+        //configuredProvider cannot be null if providerSetManually is true. If it somehow is an NPE is appropriate.
```
manovotn, line 83:
```suggestion
        //configuredProvider cannot be null if providerSetManually is true. If it somehow is, an ISE is appropriate.
```

api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
```diff
@@ -79,6 +80,10 @@ private static CDIProvider getCDIProvider() {
...
+        if (providerSetManually && configuredProvider.getCDI() == null) {
```
manovotn, line 84: You could add this logic directly into the `catch` block and just check for `providetSetManually`.

manovotn (changes requested):

api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
benjamin-confino: I've made the change.

api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
benjamin-confino: Good idea!

benjamin-confino:
I have uploaded changes based on suggestions from @manovotn

manovotn (approved):

benjamin-confino:
Hello. There has not been any movement on this for a while. Is anything else needed or can this be merged?

PR #463, opened 2021-01-11
Fix XML descriptor in documentation example

Thihup:
Signed-off-by: Thiago Henrique Hupner <thihup@gmail.com>


manovotn (approved):
Thanks for contributing :)

PR #466, opened 2021-02-03
Refer to Jakarta DI instead of JSR330

Cousjava:


spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -619,7 +619,7 @@ public class CoffeeShop
 }
 ----
 
-Note that to ensure compatibility with other JSR-330 implementations, all pseudo-scope annotations except `@Dependent` *are not* bean defining annotations.
+Note that to ensure compatibility with other Jakarta Dependency Injection implementations, all pseudo-scope annotations except `@Dependent` *are not* bean defining annotations.
```
manovotn, line 622: This should be a full name meaning - `Jakarta Contexts and Dependency Injection`

manovotn (changes requested):

spec/src/main/asciidoc/core/definition.asciidoc
Cousjava: No, this is a correct full name. https://jakarta.ee/specifications/dependency-injection/ is the replacement for JSR-330, Jakarta Context and Dependency Injection was JSR-299,JSR-346 and JSR-365.

manovotn: My bad, misread the JSR numbers. You are correct.

manovotn (approved):

PR #470, opened 2021-03-18
cdi-468 Split Spec doc to introduce CDI Lite

antoinesd:
This PR solves #cdi-468

Signed-off-by: antoinesd <antoine@sabot-durand.net>


antoinesd:
I still have some modifications to add before a first review

spec/src/main/asciidoc/preface.asciidoc
```diff
@@ -11,6 +17,10 @@ include::license-{license}.asciidoc[]
 
 Jakarta Contexts and Dependency Injection 3.0 (link:https://jakarta.ee/specifications/cdi/3.0/[CDI 3.0]) is an update to Jakarta Contexts and Dependency Injection 2.0 (link:https://jakarta.ee/specifications/cdi/3.0/[CDI 2.0]).
 
+Contexts and Dependency Injection 4.0 introduce a new feature with the CDI Lite specification.
+
+CDI Lite is a subset of CDI that doesn't require a CDI container at runtime and thus is a good candidate to develop cloud native or resources constraint applications.
```
Ladicek, line 22: It does require CDI container at runtime :-) What it doesn't require to be at runtime is certain parts of CDI initialization, such as bean discovery and extension processing. Those can be moved to application build.

manovotn:
> I still have some modifications to add before a first review

I am aware of this but I still got to glance at it and thought I might leave some notes for you if you don't mind :)

FTR any chapter numbers I mention below are from your snapshot.

 * We need to add something akin to `Packaging and deployment` to Lite part which will describe the way discovery works for Lite
   * I think the only proposal we acknowledged as working ATM is the one I wrote to the mailing list, but even if you leave it blank for now, we should have the chapter in place not to forget it
 * Chapter `2.1.5. Default bean discovery mode` is linked to the point above; I agree we want `annotated` but I don't think we found a way to achieve that in a non-breaking way without stating it in `beans.xml`
 * Any and all mentions of `beans.xml` used for enablement have to be moved into Full (probably a separate chapter that will mention that you can use `beans.xml` for local enablement?)
   * I saw this in chapter about alternatives (`2.3.1.1 Declaring selected alternatives`) and interceptors (`2.6.4 Interceptor enablement and ordering`)
 * Parts of `3.1. Inheritance and specialization` need to be moved to Lite; namely the inheritance still holds true for Lite, it is only the specialization that belongs to Full
 * `2.4.7.4. Conversation context lifecycle` defines conversation scope/context which should belong to Full
   * Same goes for `@SessionScoped` - according to [this doc](https://docs.google.com/document/d/1jS7EbHnyBBvZZxa56LmthKeoixm2FEHmqHwITGRLljQ/edit#heading=h.k6r699yh0bj7), we wanted to have it optional (basically not part of Lite by default)
   * If the above is true, then perhaps the whole section `2.4.6. Passivation and passivating scopes` probably doesn't make much sense in Lite spec part?
 * [Optional] Adapt code snippets in Lite part to only use scopes we allow there; a lot of examples have session/conversation scopes

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -106,8 +61,6 @@ A bean is said to be _enabled_ if:
 
 Otherwise, the bean is said to be disabled.
 
-Note that <<new>> defines a special rule that determines whether a `@New` qualified bean is enabled or disabled.
-This rule applies as only to `@New` qualified beans, as an exception to the normal rule defined here.
 
 [[inconsistent_specialization]]
```
manovotn, line 66: This should also be under Full in case we move specialization there.

spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
```diff
@@ -0,0 +1,128 @@
+[[injection_and_resolution_full]]
+
+== Dependency injection and lookup in {cdi_full}
+
+
+[[selection_full]]
+
+=== Modularity in {cdi_full}
```
manovotn, line 8: IMO a huge part of the text from the original `Modularity` section should be here - we don't want to talk about explicit bean archive or alternatives selected per archive in Lite. WDYT?

spec/src/main/asciidoc/preface.asciidoc
manovotn: Agreed, plus I think this part should probably be under `Major changes` section.
In fact, this `Foreword` section doesn't bring much valuable information - why don't we just merge it into `Major Changes` altogether?

manovotn (changes requested):
Added some comments - I will send a PR against your branch with the suggestions I am making here (majority of them).
We also need a placeholder chapter for bean archives (packaging and deployment basically) in Lite which is TBD but we mustn't forget about it.

There are few more point that my PR won't address and that I wanted to bring attention to:
* Specialization section should IMO be under `Full` according to what we agreed on
* CDI Full has tons on very small chapters, most of which deal with limitations and deployment exceptions for decorators (and other single feature related topics). I was thinking if it wouldn't be more readable if we were to just create, for example, a decorator chapter and list all the limitations there?
* `Modularity` in Lite currently speaks about explicit bean archives and enabling alternatives per archive. This is IMO wrong and should be mostly moved into Full. WDYT?
* Lite needs a section about `Packing and deployment`. It will re-use part of the text from Full (such as how discovery works for implicit) but even with just

manovotn:
@antoinesd I sent the above suggested changes as a PR to your repo - https://github.com/antoinesd/cdi/pull/6

spec/src/main/asciidoc/core/definition_full.asciidoc
```diff
@@ -0,0 +1,80 @@
+[[concepts_full]]
+
+== Extended Concepts for {cdi_full}
```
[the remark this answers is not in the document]
otaviojava: CDI full is the default CDI, isn't it?
Maybe, just define the CDI lite

spec/src/main/asciidoc/preface.asciidoc
otaviojava, line 22:
```suggestion
CDI Lite is a subset of CDI that uses the bean discovery process in the build time instead of runtime. Thus is an excellent candidate to develop cloud-native or resources constraint applications.
```

spec/src/main/asciidoc/preface.asciidoc
```diff
@@ -22,6 +32,8 @@ This document is organized in 4 parts:
 
 * An introduction (this part), which is not part of the specification but introduces CDI concepts and give examples.
 * The core CDI specification: <<part_1>>.
+** The CDI Lite specification which contains CDI features that contains features that could be implemented at build time
```
otaviojava, line 35:
```suggestion
CDI Lite specification contains a subset of CDI features to explore the bean discovery in the build time.
```

manovotn:
We will also need to state that we don't support `@Interceptors` in CDI Lite (as previously discussed in various places).
Otherwise we would need to support it as we only inherit what interceptors spec says.

spec/src/main/asciidoc/core/scopescontexts.asciidoc
```diff
@@ -187,7 +187,6 @@ The `@Dependent` scope is always active.
 
 Many instances of beans with scope `@Dependent` belong to some other bean and are called _dependent objects_.
 
-* Instances of decorators and interceptors are dependent objects of the bean instance they decorate.
```
Azquelt, line 190: Should this line remain here, but refer only to interceptors?

(also the corresponding line in scopescontexts_full.asciidoc)

manovotn: Yeah, this should still be in place, good catch!

spec/src/main/asciidoc/core/definition_full.asciidoc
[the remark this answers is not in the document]
manovotn: I would restrain from saying what is default - that depends on your environment and implementation that you pick up.

manovotn:
Addressed other comments present here. Also rebased onto master which contained some doc changes causing clashes.

Moved specialization under Full while leaving inheritance rules in Lite.
Changed `Modularity` to not mention per-archive enablement of alternatives and also remove notion of explicit bean archives - both were moved to Full section.

If you find anything unclear or discover any typos, misplaces chapter etc, let me know and I'll fix that.
Meanwhile, I am trying to change the Full section in some way that would make it more readable (e.g. something like condense all decorator related bits into one chapter).

manovotn:
I've added a commit that tries to make the `Full` part a but more readable. Feedback is welcome including a different way to handle this altogether, this is just my first idea on how to improve it.

What I've done:
* Put all decorator-related limitation into a paragraph under decorators
* Take several other rules from `Full` and put then back in `Lite` but state explicitly that given behavior is related to `Full`

This drastically reduces the amount of paragraphs and sections needed under `Full` so that this section can instead fully focus on specifying what features it offers on top of Lite.

manovotn:
Added a commit splitting interceptor chapter into Lite and Full to separate things such as per-archive activation. I have also added a statement into Lite section that eliminates support of `@Interceptors` in Lite as we only wanted to support resolution based on interceptor bindings.

Ladicek:
I did a comprehensive review of the entire spec and made quite a bit of changes, mostly related to the Lite/Full split. I pushed a commit with all those changes.

I have also liberally (yet carefully) added `TODO` comments all over the place, so that unfinished parts are easy to find.

Ladicek:
For the record, I also maintain a set of HTML diffs of the spec here: https://ladicek.github.io/cdi-spec/ I'm going to add a diff for my changes later today -- generating the diff takes some time.

spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -619,6 +636,8 @@ public class CoffeeShop
 Note that to ensure compatibility with other Jakarta Dependency Injection implementations, all pseudo-scope annotations except `@Dependent` *are not* bean defining annotations.
 However, a stereotype annotation including a pseudo-scope annotation *is* a bean defining annotation.
 
+// TODO we might be able to make `@Singleton` a bean defining annotation in Lite? depending on how we decide on the beans.xml thing
```
manovotn, line 639: I don't think you can do that. Because in Full it cannot be bean defining annotation and that would create inconsistency between Lite and Full.

spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -717,6 +736,8 @@ A stereotype may also specify that:
 
 A bean may declare zero, one or multiple stereotypes.
 
+// TODO how about we finally allowed declaring `@Priority` on stereotypes?
```
manovotn, line 739: What is the use case? You actually want to avoid having (for instance) multiple interceptors that all have the same priority because then you don't know which one goes first. And in case of alternatives, it is probably even an error to have two with identical priority.

spec/src/main/asciidoc/core/inheritance.asciidoc
```diff
@@ -29,6 +29,8 @@ Then the second bean inherits, and may not override, the qualifiers and bean nam
 The second bean is able to serve the same role in the system as the first.
 In a particular deployment, only one of the two beans may fulfill that role.
 
+Specialization is only present in {cdi_full} and is specified therein.
```
manovotn, line 32: Maybe add a link to the chapter?

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -732,6 +744,7 @@ The following built-in annotations define a `Literal` static nested class to sup
 * `jakarta.enterprise.inject.Any`
 * `jakarta.enterprise.inject.Default`
 * `jakarta.enterprise.inject.Specializes`
+// TODO move to Full?
```
manovotn, line 747: I noticed these as well. IMO if the user gets to consume the exact same CDI API jar, then there is little reason to extract this from spec as those literals are simply present, even though their respective annotations are ignored by Lite.

Or you could just separate these two and add them right below, stating that these are related to CDI Full?

spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
```diff
@@ -70,49 +88,70 @@ The container automatically detects inconsistent specialization and treats it as
...
 ==== Inter-module injection in {cdi_full}
 
-In addition to rules defined in <<inter_module_injection>>, when running in {cdi_full}, the following apply:
+In addition to rules defined in <<inter_module_injection>>, the following rules apply.
 
 A bean is also _available for injection_ in a certain module if:
 
-* the bean is not a decorator
+* the bean is not a decorator,
+* the bean is either not an alternative, or the module is a bean archive and the bean is a selected alternative of the bean archive.
```
manovotn, line 96: +1 for overriden becuase with the last sentence you added, it becomes very cryptic to figure out the rules.

spec/src/main/asciidoc/architecture.asciidoc
```diff
@@ -7,10 +7,10 @@ This specification defines a powerful set of complementary services that help to
 * A sophisticated, typesafe _dependency injection_ mechanism, including the ability to select dependencies at either development or deployment time, without verbose configuration
```
Emily-Jiang, line 7: I suggest to define cdi full and cdi liter concept first in this chapter and then refer to it.

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
```diff
@@ -0,0 +1,268 @@
+[[programmatic_access]]
+
+== Programmatic access to container
+
+CDI 3.0 used to provide a single `BeanManager` object, which allows access to many useful operations.
```
Emily-Jiang, line 5: This is not only true for CDI 3.0 though. I would say: The CDI version prior to 4.0

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
```diff
@@ -0,0 +1,268 @@
...
+In CDI 4.0, this is split into `BeanManagerLite` and `BeanManager`.
+
+`BeanManagerLite` provides access to a subset of `BeanManager` features which can be implemented in more restricted environments;
+It is available in {cdi_lite} environments.
+
+`BeanManager` extends `BeanManagerLite` and provides the remaining features.
```
Emily-Jiang, line 11: the remaining features -> more capabilities

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
```diff
@@ -0,0 +1,268 @@
...
+It is available in {cdi_full} environments.
```
Emily-Jiang, line 12: {cdi_full} environments environments -> a {cdi_full} environment

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
```diff
@@ -0,0 +1,268 @@
...
+
+In {cdi_lite} environment, attempting to obtain a `BeanManager` or invoking methods on it results in non-portable behavior.
+
+[[beanmanager]]
+
+=== The `BeanManagerLite` object
+
+The interface `jakarta.enterprise.inject.spi.BeanManagerLite` provides operations for obtaining contextual references for beans, along with many other operations of use to applications.
```
Emily-Jiang, line 20: Can BeanManagerLite be looked up via jndi lookup at runtime?

spec/src/main/asciidoc/core/definition.asciidoc
Ladicek: Fair point.

spec/src/main/asciidoc/core/definition.asciidoc
Ladicek: https://issues.redhat.com/browse/CDI-695 :-)

spec/src/main/asciidoc/core/inheritance.asciidoc
Ladicek: Can do. I think back-links from Full to Lite are very important -- the other way, not so much.

spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
Ladicek: I think a similarly cryptic sentence is present in the original, but I see what you mean :-)

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
Ladicek: If you could suggest a way how to phrase this without the historical reference, that would be best. I admit I didn't give it much thought, I'll think about it again.

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
Ladicek: I believe we should avoid references to JNDI, as well as other Jakarta EE concepts, in Lite.

Ladicek: (In fact, I briefly thought we should move that JNDI thing to the EE section. I don't know why it's present in "Core CDI", it's clearly EE only and for example can't be used in SE.)

Ladicek:
FTR, the HTML diff for the latest commit in this PR is now available at https://ladicek.github.io/cdi-spec/

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
manovotn: I think I raised this question at some earlier comment on the BM Lite PR as well. Technically speaking you can get the BML via JNDI simply because you can get ordinary BM (which extends BML).

That being said, IMO we don't want any specific notion of Lite supporting JNDI in any way.

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
manovotn: The plural form here is correct because there are two CDI Full environments - SE and EE

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
manovotn: `Prior to CDI 4.0, there used to be a single ....`?

spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
manovotn: Yeah,... I never said the original was better :D

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
Ladicek: I'll try to rephrase this paragraph today, there are too many comments so I clearly have to do a better job :-)

manovotn:
As discussed, we want to avoid having `Lite` as part of CDI API names. So I've changed the current proposal of `BeanManagerLite` to `BeanContainer` (see https://github.com/eclipse-ee4j/cdi/pull/471) and I have also changed all references in this PR to keep it consistent.

Ladicek:
Looking at the diff I published yesterday, I have noticed some typos -- some of then I fixed yesterday, some of them I fixed right now. I also tried to rephrase that one paragraph under "Programmatic access to container".

Latest diff is available at https://ladicek.github.io/cdi-spec/diff4.html

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
```diff
@@ -24,10 +23,11 @@ Thus, any bean may obtain an instance of `BeanContainer` by injecting it:
 
 [source, java]
 ----
-@Inject BeanContainer manager;
+@Inject BeanContainer container;
 ----
 
 The operations of `BeanContainer` may be called at any time during the execution of the application.
+// TODO Full has restrictions on when BeanManager methods can be called, do we want to reflect them here in some way?
```
manovotn, line 30: How about we instead change the `The BeanManager object` and mention that those rules apply to both, BM and `BeanContainer`?

Ladicek: That should be implied by

> In {cdi_full} environment, `BeanContainer` is subject to the same rules as `BeanManager`.

What I'm more thinking about here is: if we're saying that in Lite, `BeanContainer` can be invoked at any time, that's a superset of what's allowed in Full. Which might be problematic at least from the "Lite is a strict subset of Full" perspective.

Ladicek: Or at least, it might _look_ like a superset. That will depend on how we define the container lifecycle in Lite, which isn't there yet. (In Full, it's defined in the SPI section.)

manovotn: > BeanContainer can be invoked at any time

That's not what you wrote there though. You stated `... at any time during the execution of the application` which means after the CDI container is bootstrapped, doesn't it? In other words, in Lite you cannot even try to use `BeanContainer` earlier.
And so long as you cannot use `BeanContainer` where you couldn't use BM (from within extensions), you should be good.

Or am I missing something?

Ladicek: Again, "at any time during the execution of the application" depends on how we define lifecycle, which we didn't do yet. If we define it in a way that all initialization phases are considered _before_ application execution, then sure.

I added the TODO here mainly as a reminder, to revisit this section after we define lifecycle. I could have worded it more explicitly, for sure -- sorry about that.

antoinesd:
> I have summarized all the TODO comments that are currently present in this PR here: #482 I believe that this PR should be merged as is and all the issues summarized in #482 should be addressed separately.

It's ok for me. Waiting for @manovotn agreement

manovotn (approved):
Sure, fine by me!

PR #473, opened 2021-04-13
Removal of deprecated bits for next major version

### LINKED ISSUE #472: Removal of deprecated bits for next major version
There are certain bits in the specification that have been carried over as deprecated legacy for years.
I think for future versions we can easily shake them off.

I discovered the following deprecated parts that I think we can remove:
* `@New` qualifier
  * Deprecated since CDI 1.1 and replaced by `@Dependent`
* `Bean.isNullable()` method
  * Deprecated since CDI 1.1
* `BeanManager.fireEvent()` method
  * Deprecated since CDI 2.0 and replaced by `BeanManager.getEvent()`
* `BeanManager.createInjectionTarget(AnnotatedType)` method
  * Deprecated since CDI 1.1 and replaced by `BeanManager.getInjectionTargetFactory(AnnotatedType)`
* `BeforeBeanDiscovery.addAnnotatedType(AnnotatedType)` method
  * Deprecated since CDI 1.1 and replaced by `BeforeBeanDiscovery.addAnnotatedType(AnnotatedType, String)`
* Remove the notion of a mandatory configuration switch for all container to support different understanding of missing `beans.xml`
  * This behavior change exists ever since CDI 1.1
  * Stated in the following part of specification - https://jakarta.ee/specifications/cdi/2.0/cdi-spec-2.0.html#bean_archive
  * `For compatibility with Contexts and Dependency 1.0, products must contain an option to cause an archive to be ignored by the container when no beans.xml is present.`
  
Obviously, for all of the above changes might be needed in both, API and specification text.

manovotn:
Fixes #472 

The issue^ explains the content of this PR.
Each removed functionality is represented by a single commit.


api/src/main/java/jakarta/enterprise/inject/Instance.java
```diff
@@ -71,11 +71,6 @@
  * Instance&lt;PaymentProcessor&gt; anyPaymentProcessor;
  * </pre>
  * 
- * <p>
- * Finally, the {@link New &#064;New} qualifier may be used, allowing the application to obtain a
- * {@link New &#064;New} qualified bean:
- * </p>
- *
```
Ladicek, line 78: Probably want to remove the next paragraph as well?

Ladicek: I mean, not a paragraph, but a `<pre>` block.

manovotn: Yep thanks, missed that.
And there is similar case in `injectionandresolution` file - will address both!

manovotn: Should be good now :)

Ladicek (approved):
I didn't check if there's anything more that needs to be removed, I trust you were diligent :-) The changes here LGTM.

PR #476, opened 2021-05-10
Small improvements to use Java 8 API

otaviojava:


Emily-Jiang (approved):
LGTM!

manovotn (commented):
I don't have anything against removing unused imports.

However, I don't see any benefit in using lambdas. What's the point/reason for it? Why not extract it into a util class or just keep an anonymous one?
Similarly, the method reference versus lambda; is there any gain in doing it this way? From what I can tell, it should do the exact same thing.
In other words, I see such change as meaningless, unless somehow justified.

Emily-Jiang:
> I don't have anything against removing unused imports.
> 
> However, I don't see any benefit in using lambdas. What's the point/reason for it? Why not extract it into a util class or just keep an anonymous one?
> Similarly, the method reference versus lambda; is there any gain in doing it this way? From what I can tell, it should do the exact same thing.
> In other words, I see such change as meaningless, unless somehow justified.

Personally I think what @otaviojava has done simplifies the code and utilise some Java 8 features to declutter the code - removing the anonymous class to make code some readable.

Ladicek:
I personally don't think this simplifies anything, because lambdas like

```java
() -> {
}
```

are almost as harmful to readability as anonymous classes. It really should be `() -> {}`.

Is that mandated by some code style?

That said, I've got nothing against using lambdas or method references.

What's more interesting to me is -- I'm all for removing unused imports, but why are imports reordered? Is that, again, mandated by some code style?

I mean -- I'm all for having a unified code style, preferably mandated by Checkstyle or Eclipse Formatter or something like that -- but is that what's happening here?

arjantijms:
>  preferably mandated by Checkstyle or Eclipse Formatter or something like that 

https://github.com/eclipse-ee4j/ee4j/tree/master/codestyle

This code style can be used automatically in a maven build, see e.g.

https://github.com/eclipse-ee4j/servlet-api/blob/master/api/pom.xml#L123

Ladicek:
I have also just noticed https://github.com/eclipse-ee4j/cdi/tree/master/api/src/ide/eclipse and I wonder what should CDI use then :-)

arjantijms:
> I wonder what should CDI use then :-)

My guess would be the EE4J one ;) The jboss community one seems to have transferred from its previous location at Red Hat / Jboss.

PR #499, opened 2021-06-10
Correction to wording of EventMetadata#getInjectionPoint

manovotn:
This is an oversight from https://github.com/eclipse-ee4j/cdi/pull/473 where the wording was incorrectly changed.

The only case in which `null` was to be returned was `BM.fireEvent()`.
We even had a TCK that asserted just that, see https://github.com/eclipse-ee4j/cdi-tck/blob/3.0.2/impl/src/main/java/org/jboss/cdi/tck/tests/event/metadata/EventMetadataTest.java

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/spi/EventMetadata.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/EventMetadata.java
@@ -51,8 +51,7 @@ public interface EventMetadata {
     public Set<Annotation> getQualifiers();
 
     /**
-     * Get the {@link InjectionPoint} representing the injected {@link Event} instance which fired the event, or
-     * <code>null</code> if it was fired from {@link BeanManager#getEvent()}
+     * Get the {@link InjectionPoint} representing the injected {@link Event} instance which fired the event
      *
      * @return InjectionPoint of the Event
      */


api/src/main/java/jakarta/enterprise/inject/spi/EventMetadata.java
```diff
@@ -51,8 +51,7 @@
     public Set<Annotation> getQualifiers();
 
     /**
-     * Get the {@link InjectionPoint} representing the injected {@link Event} instance which fired the event, or
-     * <code>null</code> if it was fired from {@link BeanManager#getEvent()}
+     * Get the {@link InjectionPoint} representing the injected {@link Event} instance which fired the event
```
Ladicek, line 54: `BeanContainer#getEvent()` still exists, no?

manovotn: Yes, but we only added this in the aforementioned PR which was a mistake (`getEvent()` was there even then and we never expected `null` in such case).
See https://github.com/eclipse-ee4j/cdi/pull/473/files#diff-b3a9e85ffddc1930defb35eaf952d3d26a75763ed5b0a1756e4ae35aca825338R55

Ladicek: Ah ah I see! Thanks!

Ladicek (approved):

api/src/main/java/jakarta/enterprise/inject/spi/EventMetadata.java
Ladicek: Side note: what's the expected value when `BeanManager.getEvent().fire()` is used? Is there a TCK test for that?

manovotn: I didn't see any test for that and what we currently return (as in Weld) is an `InjectionPoint` belonging to the bean manager from which you retrieved the event.
Which IMO makes sense when you think about it. Thought it is probably not a valuable IP to inspect which is why I don't think we need TCK for it. At most we could check it is non-null value?

PR #501, opened 2021-06-28
Add getBeanContainer() method to SeContainer for consistency

graemerocher:
A new method called getBeanContainer() was added to the CDI type. For consistency this PR adds the same method to the SeContainer interface.


manovotn (approved):
Nice catch, I totally forgot about this one while filing the original PR!

PR #503, opened 2021-06-28
Create GH actions file

### LINKED ISSUE #502: Update CI setup
This project used Travis but apparently never migrated to the new variant which means the CI now doesn't work.

We should fix that.
Personally, I'd just add GH actions for which I can send a PR.

manovotn:
Fixes #502


manovotn:
Actually I am not sure what JDKs are needed for EE 10. I know it's no longer 8 and has to be 11. Beyond that, I am not sure.
Therefore, I will keep the setup with JDK 11 only ATM.

.github/workflows/ci-actions.yml
```diff
@@ -0,0 +1,30 @@
...
+    name: "Build CDI - JDK ${{matrix.java.name}}"
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java:
+          - {
+            name: "11",
+            java-version: 11,
+          }
```
Ladicek, line 18: You can do just

```yaml
jobs:
  build:
    strategy:
      fail-fast: false
      matrix:
        java:
        - 11
```

and later

```yaml
jobs:
  build:
    steps:
    - name: Set up JDK ${{ matrix.java }}
      uses: actions/setup-java@v2
      with:
        distribution: adopt
        java-version: ${{ matrix.java }}
```

(Ignore the `setup-java@v2` difference, I'm copying from a project where I use that, as it allows to choose the `distribution`.)

Ladicek: But what you have now is of course also fine.

Ladicek (approved):

.github/workflows/ci-actions.yml
manovotn: Ok, I have copied this from a setup that I was using elsewhere. Don't recall exactly why I did it this way but I think I wasn't able to reference the version from job name (which might have been a bug back then)?
Can change it to what you are proposing.

Ladicek: I don't really mind :-)

PR #504, opened 2021-06-28
Cleanup AnnotationInfo/Annotation Target

graemerocher:
I am starting to progress the build extensions impl and I think we should agree on a few things before I try put effort into API cleanup so I created this draft for discussion:

* The rest of the CDI spec uses `get*` methods for accessors so I think we should remain consistent with the rest of the API otherwise it seems a bit odd that the new model API doesn't follow the convention of the rest of the spec
* In general I think we should do away with type coercion methods like `asType`, `asDeclaration` and the rest. It seems like an artefact of Jandex but Java has a type system with `instanceof` and recent Java has pattern matching so it seems completely unneeded and error prone

If we agree that we should stick to `get*` I can progressively try and cleanup and javadoc more of this API to try to add some clarity to it.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/lang/model/AnnotationInfo.java
+++ b/api/src/main/java/jakarta/enterprise/lang/model/AnnotationInfo.java
@@ -13,14 +13,14 @@ public interface AnnotationInfo {
      *
      * @return target of this annotation
      */
-    AnnotationTarget target();
+    AnnotationTarget getTarget();
 
     /**
      * Declaration of this annotation's type.
      *
      * @return declaration of this annotation
      */
-    ClassInfo<?> declaration();
+    ClassInfo<?> getDeclaration();
 
     /**
      * Fully qualified name of this annotation.
@@ -28,8 +28,8 @@ public interface AnnotationInfo {
      *
      * @return fully qualified name of this annotation
      */
-    default String name() {
-        return declaration().name();
+    default String getName() {
+        return getDeclaration().name();
     }
 
     /**
@@ -39,7 +39,7 @@ public interface AnnotationInfo {
      * @return whether this annotation is repeatable
      */
     default boolean isRepeatable() {
-        return declaration().hasAnnotation(Repeatable.class);
+        return getDeclaration().hasAnnotation(Repeatable.class);
     }
 
     /**
@@ -57,14 +57,14 @@ public interface AnnotationInfo {
      * @param name attribute name
      * @return value of this annotation's attribute with given {@code name}
      */
-    AnnotationAttributeValue attribute(String name);
+    AnnotationAttributeValue getAttribute(String name);
 
     default boolean hasValue() {
         return hasAttribute("value");
     }
 
-    default AnnotationAttributeValue value() {
-        return attribute("value");
+    default AnnotationAttributeValue getValue() {
+        return getAttribute("value");
     }
 
     /**
@@ -72,5 +72,5 @@ public interface AnnotationInfo {
      *
      * @return all attributes of this annotation
      */
-    Collection<AnnotationAttribute> attributes();
+    Collection<AnnotationAttribute> getAttributes();
 }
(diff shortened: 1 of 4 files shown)


Ladicek:
There's a lot to cleanup for sure, so thanks for starting that! I didn't have much time for CDI Lite recently, sorry about that.

For an unrelated reason, I recently wrote a small annotation processor, so I had to deal with the `javax.lang.model` API, which is similarly structured, but lacking the `is*` / `as*` methods. I found it's a massive pain when I have to go look which subtypes are there, and explicit casting interrupts the code flow. So ... I personally think these should stay. (Maybe with different names.)

Now, if we had sealed types and pattern matching today, that would be a different story. (Note that the meaning of "today" is not yet clear in our context. If Jakarta EE 10 is based on Java 11, too bad. If 17 is selected as a base, then that's nice, but still no pattern matching for switch, so no exhaustiveness checks :-/)

The language model API is quite detached from the rest of CDI (and, as we discussed earlier, is present in CDI because CDI is the first to need it, but it should be CDI-agnostic), so I don't think consistency with the rest of the CDI API is terribly important there. Specifically for the `get*` prefix, I'm deferring to Effective Java, 3rd edition, item 68 :-) Also, the existing API doesn't use `Optional` and instead returns `null` everywhere, so using that also introduces inconsistency :-)

The new extension API, that's a different thing IMHO, and we should be consistent with the rest of CDI there.

I guess if I were to summarize...

For the language model:
- I'm not attached to `is*` / `as*`, but I like them. I'm ready to let them go if that's how everyone feels.
- I really don't like using the `get` prefix for accessors (even if it's perhaps inconsistent with other parts of CDI API).
- I agree we should use `Optional` (even if it's perhaps inconsistent with other parts of CDI API).

For the new extension API, I agree we should be consistent with the rest of the CDI API. So if they use `get*`, we do too. If they return `null` instead of `Optional`, we do too.

graemerocher:
> For an unrelated reason, I recently wrote a small annotation processor, so I had to deal with the `javax.lang.model` API, which is similarly structured, but lacking the `is*` / `as*` methods. I found it's a massive pain when I have to go look which subtypes are there, and explicit casting interrupts the code flow. So ... I personally think these should stay. (Maybe with different names.)

Can you provide an example of what is complicated? You have things like `ExecutableElement`, `VariableElement` etc. in `java.lang.model`. Having said that in Micronaut we have an abstraction over this called `MethodElement`, `FieldElement` etc. and just do `instanceof` checks. This hasn't presented any problems, so I am curious to know what issues you face.


> The language model API is quite detached from the rest of CDI (and, as we discussed earlier, is present in CDI because CDI is the first to need it, but it should be CDI-agnostic), so I don't think consistency with the rest of the CDI API is terribly important there. Specifically for the `get*` prefix, I'm deferring to Effective Java, 3rd edition, item 68 :-) Also, the existing API doesn't use `Optional` and instead returns `null` everywhere, so using that also introduces inconsistency :-)

Ok so is it envisioned that this new model package will eventually become an independent spec? Note that I changed to `Optional` to propose moving to `Optional` since it is in general more `null` safe if that is not interesting then ok, but we should agree and what pattern to follow (return `null` and ensure what `null` means in all javadoc documentation on return types)

With regards to Effective Java item 68 it states:

> Methods that return a non-boolean function function or attribute of the object are usually named with a noun, a noun phrase or a verb phrase beginning with the verb `get`

and

> Methods that performs some actions are generally named with a verb

Both of which this API seems to violate since methods like `kind()` perform no action 😉

graemerocher:
The other thing I will mention is we have sub interfaces in build time extensions API like `ClassConfig` that subclass things like `ClassInfo` thus there will be 2 conventions present (one in model and one in cdi build time extensions) if we decide to deviate which would be unfortunate IMO

Ladicek:
Yea I think `ClassConfig` extending `ClassInfo` was a mistake, honestly.

Ladicek:
Ah and `kind` is a noun in my book (or an adjective), never heard of it as a verb :-)

graemerocher:
Oh it is definitely a noun, hence why I expected it to fit in the category of ` beginning with the verb get` as per that extract from the aforementioned book

PR #506, opened 2021-07-27
remove App{Archive,ArchiveConfig,Deployment}, rename AppArchiveBuilder and ContextBuilder

Ladicek:
The `AppArchive`, `AppArchiveConfig` and `AppDeployment` types
were redundant. Moreover, they looked like they can access whole
world, which prevents certain kinds of implementations. (Some
effort was made to design the API in these types so that whole
world access is not necessary, but it's a leaky abstraction.)
For those reasons, these types are removed altogether.

The `AppArchiveBuilder` type is renamed to `ScannedClasses`,
which is more descriptive (and shorter). The `addSubtypesOf`
method is removed, because it can't be implemented in certain
environments (notably, Portable Extensions).

The `ContextBuilder` type is renamed to `ContextConfig`, because
it isn't really used in the usual builder way and is quite close
to the other existing `*Config` types.

Additionally, some documentation is improved, especially for
types in the `@Discovery` phase.


Ladicek:
This is just a first round of improvements I have in mind. More will follow for sure.

Since @graemerocher started improving the language model part, I started looking at the actual extension API, so that we don't clash (hopefully :-) ).

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/MetaAnnotations.java
```diff
@@ -3,16 +3,48 @@
...
+     * @param config allows transforming annotations on the {@code annotation}
+     */
     void addStereotype(Class<? extends Annotation> annotation, Consumer<ClassConfig<?>> config);
 
-    // includes defining the scope annotation
-    ContextBuilder addContext();
+    /**
+     * Registers custom context as configured by the returned {@link ContextConfig}.
+     */
+    ContextConfig addContext();
```
Ladicek, line 49: At this point, maybe we should just have `addContext(Class<? extends AlterableContext>)` (which is the only mandatory part) and leave `ContextConfig` purely for the optional stuff?

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/ScannedClasses.java
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * Adds all subclasses of the class with given name to the set of classes to be scanned during bean discovery.
+     *
+     * @param className binary name of the class, as defined by <cite>The Java&trade; Language Specification</cite>;
+     * in other words, the class name as returned by {@link Class#getName()}
+     */
+    // TODO adds the type itself or not? (in theory yes, as subtyping is reflexive)
+    // TODO looks like it can't be implemented on top of Portable Extensions, so maybe remove?
+    void addSubtypesOf(String className);
```
graemerocher, line 24: Can we remove this method? It seems to violate that only annotated types should be added

Ladicek: Yea we really should. I find that useful in certain scenarios, but it can't be implemented on top of Portable Extensions either, so it should go. I'll amend the PR.

Ladicek: Done.

graemerocher (approved):

graemerocher:
Great this makes life much simpler. You working the AnnotationBuilder PR now?

Ladicek:
I do, yes. Should be up shortly -- I have to resist the urge to change too many things at once :-)

PR #508, opened 2021-07-29
remove Annotations and add AnnotationBuilder

Ladicek:
The `Annotations` API was a provisional solution for creating
annotation instances. This commit replaces it with a simpler,
more traditional builder API: `AnnotationBuilder`. Considering
that `AnnotationInfo.target()` was dropped, the builder can
simply produce `AnnotationInfo` instances, which allows
simplification of other APIs that expect users to pass
annotation instances (annotation transformation and synthetic
component registration).

Additionally, this commit tries to establish a convention
for documenting `null` handling:

- if the user is forbidden to pass `null`, we say "must not
  be null" and we don't have to document that a NPE will be
  thrown;
- if the API guarantees that `null` will never be returned,
  we simply say "never null".


api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilderFactory.java
```diff
@@ -0,0 +1,15 @@
...
+import java.lang.annotation.Annotation;
+
+/**
+ * Service provider interface that supports creating {@link AnnotationBuilder}.
+ * Should not be called directly by users; the static methods on {@link AnnotationBuilder} are preferred.
+ */
+public interface AnnotationBuilderFactory {
+    AnnotationBuilder create(Class<? extends Annotation> annotationType);
+
+    AnnotationBuilder create(ClassInfo<?> annotationType);
```
graemerocher, line 14: Needs javadoc

api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
```diff
@@ -167,4 +169,14 @@ public BeanContainer getBeanContainer() {
...
+    /**
+     * Get the Build Compatible Extension {@link AnnotationBuilderFactory} for the current CDI container.
+     * <p>
+     * This is present here to reuse existing service provider discovery mechanism.
+     * Should not be called directly by users; the static methods on {@link AnnotationBuilder} are preferred.
+     *
+     * @return the {@link AnnotationBuilderFactory}
+     */
+    // TODO this is easiest, but also exposes AnnotationBuilderFactory prominently, which we don't want
+    public abstract AnnotationBuilderFactory getAnnotationBuilderFactory();
```
graemerocher, line 181: Perhaps just with service loader inside the `AnnotationBuilderFactory` interface it self:

```java
interface AnnotationBuilderFactory {
      AnnotationBuilderFactory INSTANCE = SomePackagePrivateClass.resolveFactory()
      ...
```

api/src/main/java/jakarta/enterprise/lang/model/AnnotationMemberValue.java
```diff
@@ -2,261 +2,318 @@
...
 
     /**
-     * Return the enum type that an enum value represents.
+     * Returns the enum type of the annotation member value.
      *
-     * @return A {@link jakarta.enterprise.lang.model.declarations.ClassInfo} representing the type of an enum value.
+     * @return a {@link ClassInfo} representing the enum type
+     * @throws IllegalStateException if this annotation member value is not an enum value
      */
     ClassInfo<?> asEnumClass();
```
graemerocher, line 278: I was wondering if this is necessary since in Micronaut we don't actually store the enum type and only the constant value so it would be complicate matters, given the user can already lookup types via the `Types` impl ?

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilder.java
```diff
@@ -0,0 +1,764 @@
...
+
+    /**
+     * Adds an annotation array-valued annotation member called {@code value}.
+     *
+     * @param values the annotation array, must not be {@code null} or contain {@code null}
+     * @return this {@code AnnotationBuilder}
+     */
+    default AnnotationBuilder value(Annotation... values) {
+        return member(AnnotationInfo.MEMBER_VALUE, values);
+    }
```
graemerocher, line 415: These methods will require implementors to transform `Annotation` instances to `AnnotationInfo` internally presumably via reflection, which is probably doable but more overhead

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilder.java
```diff
@@ -0,0 +1,764 @@
...
+ * Expected usage is:
+ * <ol>
+ *     <li>create the builder using {@link #of(Class)} or {@link #of(ClassInfo)};</li>
+ *     <li>use the {@code value()} and {@code member()} methods to define annotation members;</li>
+ *     <li>call {@link #build()} to create an {@link AnnotationInfo}.</li>
+ * </ol>
+ * One builder instance should not be used to create multiple annotations.
+ */
+// TODO not sure if all the methods taking ClassInfo are needed, maybe they're not
+public interface AnnotationBuilder {
```
graemerocher, line 22: This looks great!

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilder.java
Ladicek: I added this to let users use the existing `AnnotationLiteral` subclasses. Indeed it requires reflection, but I found it's not that hard to implement -- here's how I transform an `Annotation` to Jandex `AnnotationInstance`: https://github.com/Ladicek/quarkus-fork/blob/experiment-cdi-lite-ext-v5/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/cdi/lite/ext/AnnotationsReflection.java

I won't insist on keeping it, but I do think it's convenient.

api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
Ladicek: Yea, I just didn't want to reimplement all the service discovery stuff with priorities -- which we probably would have to do, since the `CDI` class does ti too.

api/src/main/java/jakarta/enterprise/lang/model/AnnotationMemberValue.java
Ladicek: I noticed you dropped this method and I thought it's probably a mistake, so I added it back. What I _didn't_ notice is that you added a parameter to `<E extends Enum<E>> E asEnum(Class<E> enumType)` :-) IMHO, that should just be `<E extends Enum<E>> E asEnum()`. The enum type is available on the declaration of the annotation member, isn't it? As in, you should always be able to find it even if you don't remember it?

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilderFactory.java
```diff
@@ -0,0 +1,15 @@
...
+import jakarta.enterprise.lang.model.declarations.ClassInfo;
+
+import java.lang.annotation.Annotation;
+
+/**
+ * Service provider interface that supports creating {@link AnnotationBuilder}.
+ * Should not be called directly by users; the static methods on {@link AnnotationBuilder} are preferred.
+ */
+public interface AnnotationBuilderFactory {
+    AnnotationBuilder create(Class<? extends Annotation> annotationType);
```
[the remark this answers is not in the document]
Ladicek: Added.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilderFactory.java
Ladicek: Added.

api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
Ladicek: I mean -- yea, I'll probably just do it, it's just copy'n'paste :-)

api/src/main/java/jakarta/enterprise/lang/model/AnnotationMemberValue.java
graemerocher: Looks like as of today we just store enum values as strings. See https://github.com/micronaut-projects/micronaut-core/blob/a949a27c69ff00fe50fa751adf0501540aee79b2/inject-java/src/main/java/io/micronaut/annotation/processing/JavaAnnotationMetadataBuilder.java#L637-L641

I guess we would have to rework that code to store the actual enum values, although I am not sure how possible that is, we would I guess need to store the underlying `VariableElement` from the annotation processing API

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilder.java
graemerocher: Ok it is probably fine.

Ladicek: If you later find out it's problematic, we can always remove it. It's strictly convenience, not a requirement.

api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
Ladicek: Done that. I'm now thinking if `AnnotationBuilder` should perhaps belong to `jakarta.enterprise.lang.model`, since it's fully self-contained. Probably don't have to solve this in this PR, can move later if we find that makes more sense.

api/src/main/java/jakarta/enterprise/lang/model/AnnotationMemberValue.java
Ladicek: Alright that sounds weird. I don't know how you [intend to] implement this, but if your implementation of `AnnotationMemberValue` retains a reference to whatever you use to implement `AnnotationInfo`, you should be able to lookup the declaration of the annotation member and from there, lookup the enum type. Am I missing something?

graemerocher: yeah probably there is a way, will try work it out

graemerocher (approved):

api/src/main/java/jakarta/enterprise/lang/model/AnnotationMemberValue.java
Ladicek: OK, thanks. I added a TODO to maybe remove the `Class` parameter from `asEnum`. Let's consider this an open question for now.

Also I realized what was the true nature of your question about `AnnotationMember` / `AnnotationMemberValue`. Currently, `AnnotationMember` is only used on one place: `AnnotationInfo.members()` return `Collection<AnnotationMember>`. That method could easily return `Map<String, AnnotationMemberValue>`, so we could remove `AnnotationMember` altogether. I added a TODO for that as well. (I assume we would then rename `AnnotationMemberValue` to something shorter, perhaps reusing the name `AnnotationMember`; I personally don't like `AnnotationValue`, because that's ambiguous.)

graemerocher: So one issue with `asEnum()` with no class parameter at least for annotation processors is that the enum type might not be on the annotation processor classpath and only be on the user class, so it may be impossible to materialize it into an enum value.

I guess in that case we could just throw an exception and say to use `asEnumConstant()` in general though using generic type arguments without a way to infer the type seems odd to me. For example this isn't type safe:

```
Foo foo = member.<Foo>asEnum();
```

And calling asEnum() without specifying the type doesn't compile automatically, or at least IntelliJ doesn't know what to do with it.

graemerocher: Oh and +1 to remove `AnnotationMemberValue` and just have `AnnotationMember`

Ladicek: Yea I thought `asEnum` might be problematic, I originally had a TODO for that method: `// TODO should this be present?`

Does adding the `Class` parameter help with the classpath issues? If so, I'm fine with leaving it there (and indeed it also help with type inference).

OK, I'll submit another PR to change `AnnotationInfo.members()` to return `Map<String, SomeAnnotationMemberValueType_ProbablyAnnotationMember>` and remove `AnnotationMember` as it currently stands. I've have to figure out one other thing now, but I should get to it later today.

graemerocher: Yes having the `Class` parameter is better as the API user knows they have to have the class before they can the enum. That is why I added it in the other PR

Ladicek: Here, the API user is the extension author. That is typically not the same person as the extension user.

I'd rather remove `asEnum` completely than having it sometimes failing with classloading errors, to be honest :-)

graemerocher: Right, my point was if the extension author has to supply a `Class` parameter, they know that they have to configure their classpath to make sure the `Class` they are requesting is on the same classpath as the extension.

```
FooEnum e = info.asEnum(FooEnum.class);
```

In the above case `FooEnum` will not be referencable if it is not on the classpath of the extension and the extension author would get a compilation error.

Ladicek: OK, if you're fine with it, then I am :-) Let's keep `asEnum` as it is now, then. I'll remove the TODO comment.

PR #512, opened 2021-08-05
Make addContext accept the annotation type

graemerocher:
AlterableContext doesn't take a generic argument and there is no way to know the
 actual scope annotation without instantiating it so this change makes the scope
 annotation required.﻿

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/ContextConfig.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/ContextConfig.java
@@ -7,29 +7,26 @@ import jakarta.enterprise.context.spi.AlterableContext;
  * Allows configuring custom scope and context. The only mandatory method is
  * {@link ContextConfig#implementation(Class)}, other methods are optional.
  * If they aren't called, correct value is deduced from the context implementation class.
+ *
+ * @param <A> The annotation type
  */
-public interface ContextConfig {
+public interface ContextConfig<A extends Annotation> {
     /**
      * If implementation class is defined using {@link #implementation(Class)},
      * this method doesn't have to be called. If it is, it overrides the scope annotation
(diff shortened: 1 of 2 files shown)


Ladicek:
My original idea was that either the scope annotation is provided explicitly by the extension author (calling `ContextConfig.scope(CustomScope.class)`), or it's discovered by instantiating the context class -- which we would require to have a public zero-parameter constructor.

Granted, it's not nice, so we could require the extension author to call both `scope` and `implementation`. We would still have to require that the context class has a public zero-parameter constructor, otherwise there's no way for the container to obtain the context object, but it's much nicer conceptually.

Either way, I don't think we have to carry around the scope annotation type -- we'd just have to restate the requirements given in the javadocs. WDYT?

graemerocher:
The issue is how do you at build time discover beans that are annotated with `@CustomScope` if you don't know the annotation? Unless it is required that the annotate in question is meta annotated with `@Scope`

Ladicek:
Sure, and there's nothing in the current API preventing that. As it is now, `ContextConfig.implementation()` is the only method that must be called. If `ContextConfig.scope()` isn't called, you would have to find the annotation type by instantiating the context class. I agree that is ugly, and we can make `ContextConfig.scope()` also mandatory.

I agree the `ContextConfig` API isn't exactly nice, but IMHO, turning `scope()` into a getter, while the others remaing setters, isn't much nicer.

Maybe we could just have this on `MetaAnnotations`:

```java
void addContext(Class<? extends Annotation> scopeAnnotationType, Class<? extends AlterableContext> contextImplementationClass); // looks at scope annotation to figure out if it's a normal scope

void addContext(Class<? extends Annotation> scopeAnnotationType, boolean isNormalScope Class<? extends AlterableContext> contextImplementationClass);
```

and be done with it?

graemerocher:
+1 for the methods

PR #519, opened 2021-09-08
Begin addressing spec todos

starksm64:
There are two actual changes:
- add information about use restrictions to BeanContainer methods
- Move the built-in Event passivation requirements to events_full

The other changes add links to the existing items for a todo.

The main question is whether the approach used to move spec content from core to full is what we want. I duplicated the section heading, and the add a statement along the lines of 'in addition to <<core_section_ref>>, in {cdi_full} ...'


spec/pom.xml
```diff
@@ -14,8 +14,8 @@
 	<description>CDI Specification documentation</description>
 
 	<properties>
-		<asciidoctor-maven.version>1.5.6</asciidoctor-maven.version>
-		<asciidoctorj-pdf.version>1.5.0-alpha.14</asciidoctorj-pdf.version>
+		<asciidoctor-maven.version>2.2.1</asciidoctor-maven.version>
+		<asciidoctorj-pdf.version>1.6.0</asciidoctorj-pdf.version>
```
Ladicek, line 18: This will break, other Asciidoctor configuration changes are required. At least based on reading the documentation (which I did for the SpecRef prototype). I specifically mean these changes: https://github.com/Ladicek/cdi-spec/blob/bb7cafddb2364dd78151a25c6a8d5a576baca0d0/spec/pom.xml#L212-L219 (without the `sourcemap` attribute, that was only for my prototype).

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
```diff
@@ -27,7 +27,20 @@ Thus, any bean may obtain an instance of `BeanContainer` by injecting it:
...
+* `getBeans(Type, Annotation...)`,
+* `resolve(Set)`,
+* `resolveInterceptors(InterceptionType, Annotation...)`,
+* `resolveObserverMethods(Object, Annotation...)`.
+
+and if the following operations are called before the `AfterDeploymentValidation` event is fired:
+
+* `createInstance()`,
+* `getReference(Bean, Type, CreationalContext)`,
+* `getInjectableReference(InjectionPoint, CreationalContext)`.
```
Ladicek, line 43: That's why I added the comment. Lite doesn't know anything about `AfterBeanDiscovery` or `AfterDeploymentValidation`.

starksm64: Hmmm, the current BeanContext interface which I thought Lite was supporting does list these in the IllegalStateException documentation of these methods?

spec/pom.xml
starksm64: Curious, I was having problems with this even building the pdf under SE 11 with the current version. I'll try just updating the asciidoctorj-pdf.

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
Ladicek: Right, we only updated 1 or 2 places in the API javadocs, so they mostly don't reflect the existence of CDI Lite. Yet. I hope :-)

spec/pom.xml
Ladicek: No, I'm saying the `sourcemap` attribute is not necessary to migrate to new Asciidoc. The `imagesdir` and `source-highlighter` attributes have to be changed. I also added `sourceDirectory`, which isn't strictly necessary, but without it, you get semi-warnings about non-existing directories.

manovotn:
@starksm64 @Ladicek what is the state of this PR? GH is showing me conflicts and it mostly adds links to TODOs (some of which might have been removed and are the source of conflicts?). Should we try to fix it or just scratch it and start addressing the issues as we have them in the GH tracker?

Ladicek:
Yea I think this should be scratched. It doesn't do much and it would most likely be easier to start over.

manovotn:
I agree. Closing this and we can start working on separate issues.

PR #522, opened 2021-09-13
fix SyntheticObserverBuilder copy&paste mistake

Ladicek:


Ladicek:
This is a little embarrassing, but wasn't found in review either :laughing:

manovotn (approved):

PR #524, opened 2021-09-13
Allow using @Priority within stereotypes.

### LINKED ISSUE #495: allow stereotypes to include @Priority
https://issues.redhat.com/browse/CDI-695

I believe this is useful in a limited number of cases, which are nevertheless pretty common. Typical example would be marking an alternative implementation of some bean as a test double -- in such case, you always want it enabled.

manovotn:
Fixes #495

### DIFF
--- a/spec/src/main/asciidoc/core/definition.asciidoc
+++ b/spec/src/main/asciidoc/core/definition.asciidoc
@@ -731,12 +731,11 @@ A stereotype encapsulates any combination of:
 A stereotype may also specify that:
 
 * all beans with the stereotype have defaulted bean names, or that
-* all beans with the stereotype are alternatives.
+* all beans with the stereotype are alternatives, or that
+* all beans with the stereotype have predefined `@Priority`.
 
 
 A bean may declare zero, one or multiple stereotypes.
-
-// TODO how about we finally allowed declaring `@Priority` on stereotypes?
 
 [[defining_new_stereotype]]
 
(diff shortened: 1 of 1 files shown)


starksm64 (approved):
LGTM

spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -839,6 +838,28 @@ We may specify that all mock objects are alternatives:
 public @interface Mock {}
 ----
 
+[[priority_stereotype]]
+
+===== Declaring stereotype with `@Priority`
+
+A stereotype may declare a `@Priority` annotation which functions as a means of enabling and ordering affected alternatives, interceptors and decorators.
```
Ladicek, line 845: Decorators are Full-only, but it probably doesn't make sense to move that single thing out of Lite. Maybe just something like

> affected alternatives, interceptors, and decorators (in {cdi_full})

Note I added an Oxford comma :-) I don't think the spec uses it consistently, but in this case, it makes it visually more clear that "in CDI Full" only applies to decorators.

spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -839,6 +838,28 @@ We may specify that all mock objects are alternatives:
...
+
+If there are two different stereotypes declared by the bean that declare different priority values, then there is no default priority and the bean must explicitly declare a `@Priority` annotation. If it does not explicitly declare priority, the container automatically detects the problem and treats it as a definition error.
```
Ladicek, line 847: Avoid passive if possible? :-)

```suggestion
If a bean declares two different stereotypes that declare different priority values, the bean must explicitly declare a `@Priority` annotation. If the bean does not explicitly declare priority, the container automatically detects the problem and treats it as a definition error.
```

spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -839,6 +838,28 @@ We may specify that all mock objects are alternatives:
...
+Following sample shows a stereotype that can be used to mark bean as globally enabled alternative:
+
+[source, java]
+----
+@Alternative
+@Priority(Interceptor.Priority.APPLICATION + 5)
+@Stereotype
+@Target(TYPE)
+@Retention(RUNTIME)
+public @interface TestReplacementStereotype {}
```
Ladicek, line 860: We could just call it `@Mock`, following on the example from previous section? Or `@TestDouble`, which is essentially the same thing (or a more general term covering mocks, stubs, spies, fakes and I don't know what).

manovotn:
@Ladicek updated as per suggestions

Ladicek (approved):

PR #527, opened 2021-09-14
API changes for allowing @Priority on stereotypes

Ladicek:


Ladicek:
This is a follow-up on #524. I realized we should probably do this sooner rather than later.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
```diff
@@ -32,18 +32,35 @@
...
+     * @return whether this stereotype is meta-annotated {@link jakarta.annotation.Priority @Priority}
+     */
+    boolean hasPriority();
+
+    /**
+     * Returns the priority value this stereotype declares using the {@link jakarta.annotation.Priority @Priority}
+     * meta-annotation. All alternatives and interceptors with this stereotype will be enabled for the application
+     * and ordered using this priority value (unless they declare priority explicitly).
+     * <p>
+     * Returns {@code 0} if this stereotype is not meta-annotated {@code @Priority}.
```
manovotn, line 53: Theoretically, `0` is a valid priority number. Shouldn't we throw an exception if its not meta-annotated with priority?
Or at least change the signature to `Integer` and return `null`?

manovotn (commented):
Good catch, I didn't realize I should change it here as well.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
Ladicek: I can do that, but for some reason, I thought that `0` is often used as a default if something doesn't have a priority defined (see e.g. https://github.com/weld/api/blob/5.0.Alpha3/weld/src/main/java/org/jboss/weld/inject/WeldInstance.java#L103). That's why I opted for `int` and a second method `hasPriority`.

Emily-Jiang: how about returning -1 instead?

Emily-Jiang: Actually, returning negative number aligning with the Priority javadoc
`Priority values should generally be non-negative, with negative values reserved for special meanings such as "undefined" or "not specified". `

manovotn: > Priority values should generally be non-negative, with negative values reserved for special meanings such as "undefined" or "not specified".

True, however, I'd still advocate that returning `null` or throwing exception is friendlier and more indicative of the fact that there is no priority present (which is not the same as not specified). Besides, having the `boolean` method to check for presence of priority should be good enough to avoid bumping into the exception by accident :)

Ladicek: This made me realize that `BeanInfo` currently doesn't cover the case when the bean is, in fact, an interceptor -- which also has a priority. This is actually quite related to this. I'll try to address it together with this one, tomorrow.

manovotn: @Ladicek all custom beans can implement `Prioritized` interface, see https://jakarta.ee/specifications/cdi/2.0/cdi-spec-2.0.html#prioritized

Ladicek: That's essentially equivalent to annotating the bean class `@Priority`. I don't see how that's relevant outside of alternatives, interceptors and decorators, but I'll make the javadoc a bit more general.

Ladicek: I pondered over this and in the end, I have to agree with @manovotn that using `Integer` as the return type, and `null` as the return value in case no `@Priority` is declared, is probably best.

Using `0` is wrong, because `@Priority` explicitly allows a priority value of `0`.

Using a negative number is bad either, because it doesn't allow distinguishing the case when `@Priority` is absent from the case when `@Priority` is present, but has this special value -- which the `@Priority` javadoc clearly allows, in special cases, and lets other specifications define those special cases.

Ladicek: Done.

Ladicek:
I have another set of changes that allow directly representing interceptors (`InterceptorInfo extends BeanInfo`). I will submit them in another PR.

manovotn (approved):

graemerocher (approved):

Ladicek:
Thanks!

PR #529, opened 2021-09-20
improve extension API

Ladicek:
- explicitly specified that CDI implementations don't have to accept
  custom implemetations of the `jakarta.enterprise.lang.model` and
  `jakarta.enterprise.inject.build.compatible.spi` interfaces
- moved to Jakarta Annotation 2.1.0-B1, removed `@ExtensionPriority`
  and specified that `@Priority` should be used on extension methods
- specified that `AnnotationBuilder.build` will throw if some members
  of the annotation type were not defined
- adjusted `AnnotationBuilder` so that its methods no longer accept
  varargs; arrays must be passed in explicitly to avoid ambiguity
  in case of single-element arrays
- added `InterceptorInfo` (extending `BeanInfo`) to provide
  information about interceptors, and allowed `@Processing` methods
  to declare parameters of type `InterceptorInfo`
- adjusted `@Enhancement` so that parameters of type  `DeclarationInfo`
  and `DeclarationConfig` can no longer be declared; it is not clear
  what is the full set of matching values
- replaced the use of `Map<String, Object>` in synthetic components
  functions by a dedicated `Parameters` interface
- adjusted synthetic component builders to allow defining a parameter
  of an enum type, as well as using `ClassInfo` for defining a parameter
  of a `Class` type
- improved `ParameterizedType` to provide access to the type (`ClassType`)
  of the generic class, in addition to the declaration (`ClassInfo`)
- replaced abbreviated forms "isn't", "doesn't" and "can't"
  with their longer forms "is not", "does not" and "cannot"


api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilder.java
```diff
@@ -16,6 +16,9 @@
  * <li>call {@link #build()} to create an {@link AnnotationInfo}.</li>
  * </ol>
  * One builder instance should not be used to create multiple annotations.
+ * <p>
+ * Note that values of <em>all</em> members of given annotation type must be defined
+ * before calling {@code build()}. If they are not, {@code build()} will throw an exception.
```
Ladicek, line 21: Should we also explicitly specify that defining members that are _not_ present on the annotation type is ignored? Or maybe it should lead to an exception?

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilder.java
```diff
@@ -67,7 +70,7 @@ default AnnotationBuilder value(boolean value) {
      * @param values the boolean array, must not be {@code null}
      * @return this {@code AnnotationBuilder}
      */
-    default AnnotationBuilder value(boolean... values) {
+    default AnnotationBuilder value(boolean[] values) {
```
Ladicek, line 73: I'm not 100% sure about this change, but when I tried to use the API, it was actually a bit surprising to me. I started wondering whether we should do something about single-element arrays, maybe automatically convert the type or so. In the end, I thought getting rid of  varargs and forcing explicit arrays is a nice way out of it.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildCompatibleExtension.java
```diff
@@ -22,13 +22,20 @@
...
+ * only affects order of extensions in a single phase.
  * <p>
  * If the extension declares multiple methods, they are all invoked on the same instance of the class.
  * <p>
  * Extension classes can be annotated {@link SkipIfPortableExtensionPresent @SkipIfPortablExtensionPresent}
  * when they are supposed to be ignored in presence of a given portable extension.
+ * <p>
+ * CDI implementations are not required to accept custom implementations of any {@code jakarta.enterprise.lang.model}
+ * or {@code jakarta.enterprise.inject.build.compatible.spi} interface. In other words, users may only use instances
+ * of these interfaces that they previously obtained from the corresponding API. If not, non-portable behavior results.
```
Ladicek, line 38: Couldn't think of a better place. We should put it to the spec text for sure, but I thought the API javadoc should have it as well.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/InterceptorInfo.java
```diff
@@ -0,0 +1,50 @@
...
+ * Interceptors are managed beans annotated {@link jakarta.interceptor.Interceptor @Interceptor}.
+ * An interceptor declares a set of {@linkplain jakarta.interceptor.InterceptorBinding interceptor binding annotations},
+ * used to associate the interceptor with target beans. It also declares at least one interceptor method:
+ * <ul>
+ * <li>for business method invocations: {@link jakarta.interceptor.AroundInvoke @AroundInvoke},</li>
+ * <li>for constructor invocations: {@link jakarta.interceptor.AroundConstruct @AroundConstruct},</li>
+ * <li>for lifecycle callbacks: {@link jakarta.annotation.PostConstruct @PostConstruct},
+ * {@link jakarta.annotation.PreDestroy @PreDestroy}.</li>
+ * </ul>
+ * Finally, an interceptor that is <em>enabled</em> also declares {@link jakarta.annotation.Priority @Priority}.
```
Ladicek, line 18: @manovotn Do you think this description makes sense? I wrote it from the CDI and especially CDI Lite perspective, knowing that interceptors don't always have to be beans, but they kinda have to in CDI Lite. So I'm wondering a little if this may be confusing.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Parameters.java
```diff
@@ -0,0 +1,88 @@
...
+ * {@code int.class}), or using the corresponding wrapper type ({@code Integer.class}).
+ * The return value is always of the wrapper type.
+ * <p>
+ * Class-typed parameters are available as instances of {@link Class}, even if an instance
+ * of {@code ClassInfo} was passed to the builder.
+ * <p>
+ * Similarly, annotation-typed parameters are always available as instances of the annotation
+ * type, even if an instance of {@code AnnotationInfo} was passed to the builder.
+ */
+public interface Parameters {
```
Ladicek, line 31: The original `Map<String, Object>`-based API isn't exactly pleasant when you've got more than just 2 or 3 parameters. So I thought a dedicated API would be better (and would have much smaller surface).

Not sure if this style (`get(String, Class)`) is better than the other suggested style (`getInt(String)`, `getString(String)`, `getEnum(String)`, `getAnnotation(String)` etc.).

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
```diff
@@ -273,6 +273,28 @@
...
+    /**
+     * Adds an enum-valued parameter to the map of construction/destruction parameters.
+     * The parameter map is passed to the {@linkplain SyntheticBeanCreator creation} and
+     * {@linkplain SyntheticBeanDisposer destruction} function when the bean is instantiated/disposed.
+     *
+     * @param key the parameter key, must not be {@code null}
+     * @param value the parameter value
+     * @return this {@code SyntheticBeanBuilder}
+     */
+    SyntheticBeanBuilder<T> withParam(String key, Enum<?> value);
```
Ladicek, line 285: Enums are pretty common, so I thought we should have a top-level way of passing enum-typed parameters. I don't think we need to have all primitive types on the top level (we have `boolean`, `int`, `long` and `double` and I think that's enough). Passing annotations is always possible, and that allows expressing everything [we want to allow].

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
```diff
@@ -283,7 +305,20 @@
...
+     * {@linkplain SyntheticBeanDisposer destruction} function when the bean is instantiated/disposed.
+     * <p>
+     * When looked up from the parameter map in the creation/destruction function, the value will be
+     * an instance of {@link Class}, <em>not</em> a {@code ClassInfo}.
+     *
+     * @param key the parameter key, must not be {@code null}
+     * @param value the parameter value
+     * @return this {@code SyntheticBeanBuilder}
+     */
+    SyntheticBeanBuilder<T> withParam(String key, ClassInfo value);
```
Ladicek, line 321: I recall you asked for this @graemerocher? And even if not, I thought that makes sense.

api/src/main/java/jakarta/enterprise/lang/model/types/ParameterizedType.java
```diff
@@ -4,20 +4,20 @@
...
     /**
-     * Returns the {@linkplain ClassInfo declaration} of the generic class that was parameterized with a list
+     * Returns the {@linkplain ClassType type} of the generic class that was parameterized with a list
      * of type arguments to form this parameterized type.
      *
-     * @return the {@linkplain ClassInfo declaration} of the generic class, never {@code null}
+     * @return the {@linkplain ClassType type} of the generic class, never {@code null}
      */
-    ClassInfo declaration();
+    ClassType genericClass();
```
Ladicek, line 20: This comes from me thinking how to describe what a type is. Parameterized type isnt' created by applying a list of type arguments to a class, but to a class _type_. So I thought that should be a first-class concept here. Allowing access to the declaration still makes sense, so that is retained.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Enhancement.java
```diff
@@ -11,7 +11,6 @@
  * <p>
  * Methods annotated {@code @Enhancement} must define exactly one parameter of one of these types:
  * <ul>
- * <li>{@link DeclarationConfig} or {@link jakarta.enterprise.lang.model.declarations.DeclarationInfo DeclarationInfo}</li>
```
Ladicek, line 14: I tried to implement this and found that it isn't exactly clear _what_ should be passed to an `@Enhancement` method that declares a parameter of type `DeclarationInfo` or `DeclarationConfig`.

Say I have this:

```java
@Enhancement
@ExactType(MyService.class)
void enhance(DeclarationConfig decl) {
    ...
}
```

How is that method supposed to be called? Once with the `MyService` class, and then once for each field in the class, and then once for each method in the class?

What if it uses `DeclarationInfo`? Should it be, in addition to above, also once for each parameter of each method in the class? Note that we have `ParameterInfo`, but don't have `ParameterConfig`.

I now think this shouldn't be allowed. It should always be either `Class{Info,Config}`, or `Field*`, or `Method*`.

Ladicek:
Added comments where I thought some discussion is warranted.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Enhancement.java
graemerocher: I agree this shouldn't be allowed

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Parameters.java
graemerocher: FWIW this is what we have in Micronaut that is equivalent https://github.com/micronaut-projects/micronaut-core/blob/19fef5801f31a729d88d8df47c3410fbfa442eda/core/src/main/java/io/micronaut/core/annotation/AnnotationValueResolver.java#L27

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildCompatibleExtension.java
graemerocher: I assume you mean annotation values with no default values?

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilder.java
graemerocher: I assume you mean annotation values with no default values?

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilder.java
Emily-Jiang: I prefer varargs instead of the array. In the previous form, you can pass one value while the proposed change needs to convert the single value to an array.

Ladicek: That is in fact incorrect, and shows exactly the same confusion I found myself in.

Note that we have overloaded methods `value(boolean)` and `value(boolean...)`. If you call `value(true)`, the overload resolution process will always select the first method (see https://docs.oracle.com/javase/specs/jls/se17/html/jls-15.html#jls-15.12.2: _This ensures that a method is never chosen through variable arity method invocation if it is applicable through fixed arity method invocation_).

Varargs here make sense if _more than 1_ argument is passed, or in case of an empty array. But for a single-argument array, you would always have to write `new boolean[] { true }`, which is inconsistent and, as demonstrated above, confusing.

Ladicek: Another method to eliminate the confusion would be, of course, to eliminate the overloads, so that the varargs accepting method would be called differently. I don't have a good name, personally.

manovotn: > Another method to eliminate the confusion would be, of course, to eliminate the overloads

That's what I was thinking when I saw your comment. Couldn't we have a `value` method name for single value and `values` for varargs? Or something like `singleValue` and `value`. Basically, I'd just look for some names so that we are able to keep varargs variants in place.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Enhancement.java
manovotn: +1, using `DeclarationInfo` and `DeclarationConfig` should throw an exception

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/InterceptorInfo.java
manovotn: @Ladicek I think it is mostly fine. Interceptors are beans and for purposes of Lite, your description fits. Just two minor notes:

> It also declares at least one interceptor method:

`It also declares at most one of each interceptor methods:` 
(as declaring same interception method multiple times should be considered an error)

> Finally, an interceptor that is <em>enabled</em> also declares {@link jakarta.annotation.Priority @Priority}.

`Finally, an interceptor is by default disabled. In order to enable it, it should declare {@link jakarta.annotation.Priority @Priority}.` 
(just to state what the default is)

manovotn (commented):
Dropped two separate comments but overall it looks good.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilder.java
Emily-Jiang: The name `values` makes sense.

Ladicek: The name `values` doesn't make sense, because the method defines a value of the annotation member called `value`.

Having only `value(boolean...)` would be possible, but would require implementations to find what's the type of the annotation member, which necessitates verifying that given annotation member even exists, which is something I'd like to avoid. But maybe there's a way I don't see.

Emily-Jiang (changes requested):
I think a lot of changes is about updating varargs to array, which I think it is not a good change.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilder.java
Ladicek: To clarify further.

One reason why I don't want to force implementations to find types of annotation members is that I think it should be possible to define values for annotation members that don't exist (those values would be ignored). The reason is hypothetical forward compatibility. It should be possible to define a value for an annotation member that doesn't exist on an older version of the annotation, but does exist on a newer version of the annotation.

Maybe that's not entirely useful and we should fail in such situation -- I can see benefits in both ways and don't really mind. (I just pushed a change where I explicitly specify that nonexisting members are ignored. I can amend it again to say the opposite.)

Another reason is that I think callers should be _explicit_ about which type of the annotation member they are defining. On other places, we moved away from type coercion, I don't think we should add it here. When I read `value(true)`, I need to know immediately what type is it -- I don't want to look into the annotation definition to figure out if it's singular or array. Principle of locality.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/InterceptorInfo.java
Ladicek: Done.

I vaguely recall some spec saying that an interceptor must declare at least one interceptor method, but I can't find that anymore :-) So I went with "at most one interceptor method for each interception type".

I reworded the sentence about `@Priority` to better convey what you suggest, but still fit into the overall text flow.

I also added a note that this description applies to Lite, and there are more ways to declare an interceptor in Full.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilder.java
graemerocher: personally I think we should have `value(boolean)` and `value(boolean[])` it is not worth the hassle to support varargs and yes we 100% need the ability to specify meta-members that don't actually exist

Ladicek: Just out of curiosity @graemerocher, what is your reason to support specifying annotation members that don't exist? I admit mine is just the hypothetical forward compatibility -- I'd love to know if you have more.

graemerocher: We use it in Micronaut Data. We precompute the SQL/JPA-QL query and store it a meta annotation member that has actual real equivalent

Ladicek: That's an interesting one, thanks for sharing!

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/AnnotationBuilder.java
Ladicek: Done.

api/src/main/java/jakarta/enterprise/lang/model/types/ParameterizedType.java
Ladicek: I realized that I got this one wrong. The type arguments are not applied to the class _type_; it's applied to the _type constructor_ that is introduced by the class declaration. I need to think more about this :-)

graemerocher (approved):

Ladicek:
@Emily-Jiang I merged this PR, because it contains many other valuable changes. I'd be happy to debate the merits of varargs vs. plain arrays separately, if you're interested, and am open to changing it back if we find a good solution.

Emily-Jiang:
@Ladicek can you explain why array is more preferred than varargs? I think it might be better if you explained and then merged rather than the other way round.

Ladicek:
I shared my reasoning in the comments extensively, but I'll repeat it for you in a condensed form.

If we have `value(boolean)` and `value(boolean...)`, we also have an ambiguity for a single-element array, which needs to be resolved by the caller by creating an array explicitly.

If we only have `value(boolean...)`, we force implementations to look up the type of the annotation member, which is undesirable.

We don't have a good alternative name for the array-accepting method; `values` is wrong, because the method is used to set the `value` annotation member (similarly for the `member` method -- `members` is also wrong, because it's still for a single member).

Having `value(boolean)` and `value(boolean[])` is perhaps not so nice when you can write down the array as a sequence of literals (can't estimate how common that is, but you sometimes need to _compute_ the array, in which case it doesn't matter), but it makes it immediately obvious what method is being called and what is the intended type.

Emily-Jiang:
@Ladicek thanks for the further explanation! In my previous comments, I noticed not only a number of `values` being impacted but also `members`. 

> If we only have value(boolean...), we force implementations to look up the type of the annotation member, which is undesirable.
I think forcing implementations doing something extra does give end uers a nicer API though.

Ladicek:
Yes, the same reasoning applies to the `member` methods. I keep referring to `value(boolean)` / `value(boolean...)`, but the same problem exists for `member(String, int)` / `member(String, int...)` and all the other combinations.

> I think forcing implementations doing something extra does give end uers a nicer API though.

My opinion is that this also leads to worse API. With explicit array creation, the code author must be explicit about the type of the annotation member. This means that if you read the code, you know exactly what type the member is supposed to have. If that information is implicit, you have to look at the annotation type when reading the code. Further, it would prevent certain idioms, such as keeping additional data in "non-existing" annotation members.

PR #531, opened 2021-09-21
Adjust wording of specification and API around @Priority usage on ste…

manovotn:
…reotypes.

Follow up on #524 

This PR aligns the wording closely to what spec already uses. Instead of specifically listing options, it simply allows usage of this stereotype on beans.

### DIFF
--- a/spec/src/main/asciidoc/core/definition.asciidoc
+++ b/spec/src/main/asciidoc/core/definition.asciidoc
@@ -840,7 +840,7 @@ public @interface Mock {}
 
 ===== Declaring stereotype with `@Priority`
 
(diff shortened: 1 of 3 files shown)


spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -840,7 +840,7 @@ public @interface Mock {}
 
 ===== Declaring stereotype with `@Priority`
 
-A stereotype may declare a `@Priority` annotation which functions as a means of enabling and ordering affected alternatives, interceptors, and decorators (in {cdi_full}).
+A stereotype may declare a `@Priority` annotation which functions as a means of enabling and ordering affected beans. (in {cdi_full}).
```
Ladicek, line 843: Probably want to remove the `(in {cdi_full}).` suffix :-)

Ladicek (approved):

PR #532, opened 2021-09-22
add BeanConfigurator.priority

Ladicek:
This is to allow setting a priority for synthetic beans created
using `BeanConfigurator`.

I briefly considered if `priority` should be added to
`BeanAttributesConfigurator` as well, but that would have
a ripple effect. It also doesn't reflect the present reality.
A custom `Bean` implementation can be added a priority
by implementing the `Prioritized` interface. Adding only
`BeanConfigurator.priority` is a perfect mirror of that
in the configurator API.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/spi/configurator/BeanConfigurator.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/configurator/BeanConfigurator.java
@@ -316,4 +316,20 @@ public interface BeanConfigurator<T> {
      */
     BeanConfigurator<T> alternative(boolean value);
 
+    /**
+     * Set the priority of the configured bean.
+     * By default, the configured bean does not have a priority.
+     * <p>
+     * This is equivalent to putting the {@link jakarta.annotation.Priority Priority}
+     * annotation to an actual bean class or making a custom
+     * {@link jakarta.enterprise.inject.spi.Bean Bean} class implement
+     * {@link jakarta.enterprise.inject.spi.Prioritized Prioritized}.
+     * <p>
+     * This method has no effect if the configured bean is not an alternative.
+     *
+     * @param priority the priority value
+     * @return self
+     */
+    BeanConfigurator<T> priority(int priority);
+
 }


Ladicek:
This alters the Portable Extensions API, so I'm not adding Graeme for review. The only relationship to Lite is that this addition will make it possible to completely implement the Build Compatible Extensions API on top of Portable Extensions.

api/src/main/java/jakarta/enterprise/inject/spi/configurator/BeanConfigurator.java
```diff
@@ -316,4 +316,18 @@
...
     BeanConfigurator<T> alternative(boolean value);
 
+    /**
+     * Set the priority of the configured bean.
+     * By default, the configured bean does not have a priority.
+     * <p>
+     * This is equivalent to putting the {@link jakarta.annotation.Priority Priority}
+     * annotation to an actual bean class or making a custom
+     * {@link jakarta.enterprise.inject.spi.Bean Bean} class implement
+     * {@link jakarta.enterprise.inject.spi.Prioritized Prioritized}.
```
manovotn, line 326: I would add a note saying something along these lines:
> This method has no effect on custom beans which are not alternatives.

manovotn (commented):
Added small comment, otherwise LGTM.

Notably, this already exists in Weld, see https://github.com/weld/api/blob/master/weld/src/main/java/org/jboss/weld/bootstrap/event/WeldBeanConfigurator.java#L135-L144

api/src/main/java/jakarta/enterprise/inject/spi/configurator/BeanConfigurator.java
Ladicek: Done.

manovotn (approved):

manovotn:
@Ladicek Btw we should create a tracking issue for TCK test or send one right away (we could probably use existing weld test) so that we don't forget.

starksm64:
I added a TCK issue

Ladicek:
Whoops! Yes, good point. @manovotn can you please point me to the existing Weld test? I'll try submitting a PR to the TCK.

manovotn:
@Ladicek this one - https://github.com/weld/core/tree/master/tests-arquillian/src/test/java/org/jboss/weld/tests/alternatives/customBeanPriority

PR #534, opened 2021-09-27
Programmatic lookup improvements, introduce Handle interface.

### LINKED ISSUE #521: Programmatic lookup improvements
Both the reference implementation (Weld) and ArC have an enhanced version of `javax.enterprise.inject.Instance` that provide features that might be worth adding to the spec.

| Enhancement | Description |
| --- | --- |
| [`Instance#getHandler()`](https://github.com/weld/api/blob/4.0/weld/src/main/java/org/jboss/weld/inject/WeldInstance.java#L65-L76) (name TBC) | Allows the client to obtain a "contextual reference handler" which not only holds the contextual reference but also makes it possible to inspect the _metadata_ (i.e. `Bean<?>`) of the relevant bean and to destroy the underlying contextual instance. |
| [`Instance#handlers()`](https://github.com/weld/api/blob/4.0/weld/src/main/java/org/jboss/weld/inject/WeldInstance.java#L78-L88) (name TBC) | Returns an `Iterable` of handlers described above. |
| [`select()`](https://github.com/weld/api/blob/4.0/weld/src/main/java/org/jboss/weld/inject/WeldInstance.java#L119-L134) method that accepts `java.lang.reflect.Type` | This allows for generic selection of instances which can be handy while dealing with third party beans through extensions. However, it's not type-safe. |
| Caching the result of `Instance#get()`| See https://quarkus.io/guides/cdi-reference#caching-the-result-of-programmatic-lookup |

See also Weld docs about `WeldInstance` -  [Enhanced version of jakarta.enterprise.inject.Instance](https://docs.jboss.org/weld/reference/latest/en-US/html_single/#_enhanced_version_of_jakarta_enterprise_inject_instance).

manovotn:
Resolves #521 

BTW, I will also prepare a draft of a TCK test based on what we had in Weld.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/Instance.java
+++ b/api/src/main/java/jakarta/enterprise/inject/Instance.java
@@ -21,6 +21,7 @@ import java.lang.annotation.Annotation;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
+import jakarta.enterprise.inject.spi.Bean;
 import jakarta.enterprise.util.AnnotationLiteral;
 import jakarta.enterprise.util.TypeLiteral;
 import jakarta.inject.Provider;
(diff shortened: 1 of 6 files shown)


manovotn:
@Ladicek @mkouba I didn't include the [`getPriorityComparator()`](https://github.com/weld/api/blob/4.0/weld/src/main/java/org/jboss/weld/inject/WeldInstance.java#L98-L108) method in this PR. If you think it should be included, do let me know please :)

api/src/main/java/jakarta/enterprise/inject/Instance.java
```diff
@@ -227,4 +228,88 @@ default boolean isResolvable() {
      */
     void destroy(T instance);
 
+    /**
+     * Obtains an initialized contextual reference handle for a bean that has the required type and qualifiers and is
+     * eligible for injection. Throws exceptions if there is no such bean or more than two beans.
```
Ladicek, line 233: More than _two_?

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -687,6 +690,53 @@ The bean instance passed to `destroy()` should be a dependent scoped bean instan
...
+It is an abstraction allowing inspection of bean metadata via `Bean<?>` objects.
+Handles have to resolve their contextual references lazily, i.e. when their `get()` method is invoked.
+Last but not least, this interface can be used to destroy the contextual instance once not needed.
+
+[source, java]
+----
+public interface Handle<T> extends AutoCloseable {
+
+        T get();
+        Bean<?> getBean();
```
Ladicek, line 719: Just wondering why this is not `Bean<T>`?

Ladicek:
I would stay away from the priority comparator, if only because of the `0` value :-)

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
manovotn: You're correct, fixed.

Ladicek: I was just wondering, but if it should really be `Bean<T>`, then it should be in the API as well :-)

manovotn:
@Ladicek could you please green-stamp this? Unless there is something else you want me to change of course.

Ladicek (approved):

Ladicek:
Done. Sorry, totally forgot about this!

manovotn:
> Actually I know that I wanted something changed :-) The `Handle.getBean()` method return type.

Ah, right, good point!
Fixed that, the method now looks like this - `Bean<T> getBean();`

Ladicek (approved):

PR #536, opened 2021-10-05
extension API improvements

Ladicek:


Ladicek:
As a separate commit, I added support for records in the language model, because I thought perhaps we should stay current :-) If that is problematic for some reason, I'd be fine with removing it.

graemerocher (approved):

PR #537, opened 2021-10-08
Address TODO in inter-module injection.

### LINKED ISSUE #491: Lite/Full split: decide on "Inter-module injection in CDI Full"
The "Inter-module injection in CDI Full" chapter in the specification has a TODO for deciding whether the chapter should extend rules from CDI Lite and add few more (current state, somewhat hard to read), or completely override the CDI Lite rules (would require some duplication, but might be easier to read).

https://github.com/eclipse-ee4j/cdi/blob/864e1b4f779e4d6d5a80f147d975a510446d4987/spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc

manovotn:
Fixes #491 

I have rephrased the section so that it spell out a complete set of rules.
Furthermore, the notion of custom `Bean` implementation was moved from Lite section to Full and the corresponding TODO was erased.

### DIFF
--- a/spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
+++ b/spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
@@ -88,14 +88,16 @@ The container automatically detects inconsistent specialization and treats it as
 
 ==== Inter-module injection in {cdi_full}
 
-In addition to rules defined in <<inter_module_injection>>, the following rules apply.
+Instead of the rules in <<inter_module_injection>>, the following rules apply in {cdi_full}.
 
-A bean is also _available for injection_ in a certain module if:
+A bean is _available for injection_ in a certain module if:
 
-* the bean is not a decorator,
-* the bean is either not an alternative, or the module is a bean archive and the bean is a selected alternative of the bean archive.
+* the bean is not an interceptor or decorator,
+* the bean is enabled,
+* the bean is either not an alternative, or the module is a bean archive and the bean is a selected alternative of the bean archive, or the bean is a selected alternative of the application, and
+* the bean class is required to be accessible to classes in the module, according to the class accessibility requirements of the module architecture.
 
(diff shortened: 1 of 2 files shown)


spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
```diff
@@ -88,14 +88,17 @@ The container automatically detects inconsistent specialization and treats it as
 
 ==== Inter-module injection in {cdi_full}
 
-In addition to rules defined in <<inter_module_injection>>, the following rules apply.
+{cdi_full} extends the set of rules defined in <<inter_module_injection>>.
+Below is a complete set of rules.
```
Ladicek, line 92:
```suggestion
Instead of the rules in <<inter_module_injection>>, the following rules apply in {cdi_full}.
```

manovotn: Hmm, I had it worded similarly originally. However, that makes it sounds like the set of rules is different when it is really just a superset. I can change it to your suggestion though, I don't mind either.

spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
```diff
@@ -88,14 +88,17 @@ The container automatically detects inconsistent specialization and treats it as
...
 
-A bean is also _available for injection_ in a certain module if:
+A bean is _available for injection_ in a certain module if:
 
-* the bean is not a decorator,
-* the bean is either not an alternative, or the module is a bean archive and the bean is a selected alternative of the bean archive.
```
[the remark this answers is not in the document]
manovotn: Oops, I meant to copy those old rules indeed but somehow copied different text. Thanks for catching this!

spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
Ladicek: Or not :-) My suggestion also works :-)

PR #539, opened 2021-10-08
Explicitly state that Java SE impl of CDI has to be Full implementation.

### LINKED ISSUE #488: Lite/Full split: does CDI in Java SE environment (SeContainer etc.) have to support Full?
I added a TODO "are implementations required to support Full?" into the CDI SE chapter, because I think this should be spelled out explicitly. The answer is probably yes, but we may need to debate it a little (because I think Micronaut is trying to implement some parts of the SE API, and I don't see anything particularly wrong with that, even though they won't support Full).

Could you please confirm whether you intend to implement `SeContainer` etc., @graemerocher?

https://github.com/eclipse-ee4j/cdi/blob/864e1b4f779e4d6d5a80f147d975a510446d4987/spec/src/main/asciidoc/javase/javase.asciidoc

manovotn:
Fixes #488 

Explicitly state that Java SE impl == CDI Full variant. This is already implied by previous sentence but we wanted to underline it.


spec/src/main/asciidoc/javase/javase.asciidoc
```diff
@@ -2,7 +2,7 @@
 --
 This part of the document specifies additional rules and features when using CDI in Java SE.
 All content defined in <<part_1>> applies to this part.
-// TODO are implementations required to support Full?
+Therefore, any implementation of CDI in Java SE has to be built on top of {cdi_full}.
```
Ladicek, line 5: Well implementations of CDI Lite are also in Java SE :-) The name of this chapter is hard to use frankly.

I'd say something like:

```suggestion
In particular, implementations must provide support for the entire {cdi_full} specification.
Support for CDI in Java SE that only requires {cdi_lite} may be added in subsequent versions of CDI.
```

2nd line being optional of course :-)

PR #543, opened 2021-10-14
Introduce observable CDI events representing container state.

### LINKED ISSUE #496: add a portable event corresponding to application startup
CDI doesn't provide an event that would correspond to application startup. A commonly used substitute is the `@Initialized(ApplicationScoped.class) Object` event, but that is fired "when the application context is initialized" (quoting from the spec).

Traditionally, those two situations occur reasonably close in time, but with CDI Lite, we aim to expand the set of possible CDI implementation strategies to include those that initialize a lot of stuff (including the application context) _ahead of time_. In such case, the `@Initialized(ApplicationScoped.class) Object` event has no correlation to application startup. It may happen during application startup, or it may happen arbitrary number of hours before.

Hence, I believe CDI should provide an event type that specifically corresponds to application startup. In some ways, it would be nice to restrict it to only be fired _after_ the application fully starts up and everything is initialized, but that has prohibitive runtime costs; even the `@Initialized(ApplicationScoped.class) Object` event doesn't guarantee that (and firing the new event at the same time as this "old" event should be a valid implementation strategy). So I propose to add an event type that is fired synchronously _during application startup_.

As for the name, I originally thought it could be called `AfterStartup`, because that is symmetric to existing `BeforeShutdown`, but that reasoning is invalid. `BeforeShutdown` is a container lifecycle event that is designed for use with Portable Extensions (the spec directly says "observer methods of these events must belong to extensions"). Also, in previous paragraph, I specifically argue we shouldn't restrict the event to "after startup"; it should be "during startup". However, `DuringStartup` looks ugly.

There's also zero precedent (in CDI) for using naming conventions such as `On[Something]` or `[Something]Event`, so I'd like to avoid those. The common event naming strategy in CDI is to just use a descriptive name.

That would, in this case, be simply `Startup`. However, I'd like to reserve this name for a potential `@Startup` annotation which would, similarly to EJB's `@Startup`, force eager initialization of a bean.

Any suggestions? :-)

For symmetry, we may want to declare a corresponding "during shutdown" event. We already have the `@BeforeDestroyed(ApplicationScoped.class) Object` and `@Destroyed(ApplicationScoped.class) Object` events, and it's hard to imagine how these event types could be non-portable, so this is certainly not required. As I said, my main reason for adding such event type would be symmetry with the "during startup" event. Maybe also better discoverability (the application context events are, in my experience, not exactly obvious).

manovotn:
Fixes #496

### DIFF
--- a/spec/src/main/asciidoc/core/events.asciidoc
+++ b/spec/src/main/asciidoc/core/events.asciidoc
@@ -596,4 +596,33 @@ As specified in <<builtin_contexts>>, contexts associated with built-in normal s
 * Otherwise, if the observer method is any other kind of transactional observer method, it is called in an unspecified transaction context, but with the same lifecycle contexts as the transaction that just completed.
 * Otherwise, the observer method is called in the same transaction context and lifecycle contexts as the invocation of `Event.fire()` or `BeanManager.getEvent()`.
 
+[[observable_container_lifecycle_events]]
+
+=== Observable container lifecycle events
+
+[[startup_event]]
+
+==== Startup event
+
+Implementations are required to _synchronously_ fire an event with payload `jakarta.enterprise.event.Startup` and qualifier `jakarta.enterprise.inject.Any` during application initialization.
+This event is fired after the event with qualifier `@Initialized(ApplicationScope.class)` but before processing requests.
+
+This event can be observed by integrators and libraries to perform any kind of early initialization as well as by users as a reliable entry point for when the CDI container is ready.
+
+Observer methods for this event are encouraged to specify `@Priority` to determine ordering with lower priority numbers being recommended for platform/framework/library integration and higher numbers for user applications.
+
+Applications must never manually fire any events with payload type `jakarta.enterprise.event.Startup`.
+
+[[shutdown_event]]
+==== Shutdown event
+
+Implementations are required to _synchronously_ fire an event with payload `jakarta.enterprise.event.Shutdown` and qualifier `jakarta.enterprise.inject.Any` during application shutdown.
+This event is fired during CDI container shutdown but not later than the event with qualifier `@BeforeDestroyed(ApplicationScoped.class)`.
+
+This event can be observed by integrators and libraries to perform any kind of pre-shutdown operation as well as by users as a reliable entry point for when the CDI container is about to shut down.
+
+Observer methods for this event are encouraged to specify `@Priority` to determine ordering with lower priority numbers being recommended for user applications and higher numbers for platform/framework/library integration.
+
+Applications must never manually fire any events with payload type `jakarta.enterprise.event.Shutdown`.
+
 
--- a/api/src/main/java/jakarta/enterprise/event/Shutdown.java
+++ b/api/src/main/java/jakarta/enterprise/event/Shutdown.java
@@ -1,0 +1,27 @@
+package jakarta.enterprise.event;
+
+/**
+ * <p>
+ * A CDI event with payload of type {@link Shutdown} and qualifier {@link jakarta.enterprise.inject.Any} is
+ * <i>synchronously</i> fired by CDI container during application shutdown.
+ * Applications must never manually fire any events with {@link Shutdown} as payload.
+ * </p>
+ *
+ * <p>
+ * Implementations have to fire this event during CDI container shutdown, but not later than the event with qualifier
+ * {@code @BeforeDestroyed(ApplicationScoped.class)}.
+ * </p>
+ *
+ * <p>
+ * This event can be observed by integrators and libraries to perform any kind of pre-shutdown operation as well as by
+ * users as a reliable entry point for when the CDI container is about to shut down.
+ * </p>
+ *
+ * <p>
+ * Observers are encouraged to specify {@code @Priority} to determine ordering with lower priority numbers being
+ * recommended for user applications and higher numbers for platform/framework/library integration.
+ * </p>
+ * See also {@link jakarta.interceptor.Interceptor.Priority}
+ */
+public class Shutdown {
+}
--- a/api/src/main/java/jakarta/enterprise/event/Startup.java
+++ b/api/src/main/java/jakarta/enterprise/event/Startup.java
@@ -1,0 +1,27 @@
+package jakarta.enterprise.event;
+
+/**
+ * <p>
+ * A CDI event with payload of type {@link Startup} and qualifier {@link jakarta.enterprise.inject.Any} is
+ * <i>synchronously</i> fired by CDI container during application initialization.
+ * Applications must never manually fire any events with {@link Startup} as payload.
+ * </p>
+ *
+ * <p>
+ * Implementations have to fire this event after the event with qualifier {@code @Initialized(ApplicationScope.class)}
+ * but before processing requests.
+ * </p>
+ *
+ * <p>
+ * This event can be observed by integrators and libraries to perform any kind of early initialization as well as by
+ * users as a reliable entry point for when the CDI container is ready.
+ * </p>
+ *
+ * <p>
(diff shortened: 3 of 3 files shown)


api/src/main/java/jakarta/enterprise/util/Shutdown.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * <p>
+ * A CDI event with payload of type {@link Shutdown} and qualifier {@link jakarta.enterprise.inject.Any} is
+ * <i>synchronously</i> fired by CDI container during application shutdown.
+ * Applications must never manually fire any events with {@link jakarta.enterprise.util.Shutdown} as payload.
+ * </p>
+ *
+ * <p>
+ * Implementations have to fire this event during CDI container shutdown, but not later than
+ * {@code @BeforeDestroyed(ApplicationScoped.class)}.
```
Ladicek, line 12:
```suggestion
 * {@code @BeforeDestroyed(ApplicationScoped.class) Object}.
```

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -596,4 +596,33 @@ As specified in <<builtin_contexts>>, contexts associated with built-in normal s
...
+
+Observer methods for this event are encouraged to specify `@Priority` to determine ordering with lower priority numbers being recommended for platform/framework/library integration and higher numbers for user applications.
+
+Applications must never manually fire any events with payload type `jakarta.enterprise.util.Startup`.
+
+[[shutdown_event]]
+==== Shutdown event
+
+Implementations are required to _synchronously_ fire an event with payload `jakarta.enterprise.util.Shutdown` and qualifier `jakarta.enterprise.inject.Any` during application shutdown.
+This event is fired during CDI container shutdown but not later than `@BeforeDestroyed(ApplicationScoped.class)`.
```
Ladicek, line 620:
```suggestion
This event is fired during CDI container shutdown but not later than `@BeforeDestroyed(ApplicationScoped.class) Object`.
```

api/src/main/java/jakarta/enterprise/util/Startup.java
```diff
@@ -0,0 +1,26 @@
...
+ * </p>
+ *
+ * <p>
+ * Implementations have to fire this event after {@code @Initialized(ApplicationScope.class)} but before processing
+ * requests.
+ * </p>
+ *
+ * <p>
+ * This event can be observed by integrators and libraries to perform any kind of early initialization as well as by
+ * users as a reliable entry point for when the CDI container is ready to process requests.
```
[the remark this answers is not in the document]
manovotn: Should we just say `for when the CDI container is ready`?

api/src/main/java/jakarta/enterprise/util/Shutdown.java
Ladicek: Or perhaps better:

```suggestion
 * the event with qualifier {@code @BeforeDestroyed(ApplicationScoped.class)}.
```

Similarly to other places.

Ladicek:
Maybe move the event types to `jakarta.enterprise.event`?

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -596,4 +596,33 @@ As specified in <<builtin_contexts>>, contexts associated with built-in normal s
...
+=== Observable container lifecycle events
+
+[[startup_event]]
+
+==== Startup event
+
+Implementations are required to _synchronously_ fire an event with payload `jakarta.enterprise.util.Startup` and qualifier `jakarta.enterprise.inject.Any` during application initialization.
+This event is fired after `@Initialized(ApplicationScope.class)` but before processing requests.
+
+This event can be observed by integrators and libraries to perform any kind of early initialization as well as by users as a reliable entry point for when the CDI container is ready to process requests.
```
Ladicek, line 610:
```suggestion
This event can be observed by integrators and libraries to perform any kind of early initialization, as well as by users as a reliable entry point for when the CDI container is ready to process requests.
```

api/src/main/java/jakarta/enterprise/util/Shutdown.java
```diff
@@ -0,0 +1,26 @@
...
+ * </p>
+ *
+ * <p>
+ * This event can be observed by integrators and libraries to perform any kind of pre-shutdown operation as well as by
+ * users as a reliable entry point for when the CDI container about to shut down.
+ * </p>
+ *
+ * <p>
+ * Observers are encouraged to specify {@code @Priority} to determine ordering with lower priority numbers being
+ * recommended for user applications and higher numbers for platform/framework/library integration.
```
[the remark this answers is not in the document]
manovotn: Gosh, you're persistent :-D
Ok, I'll add a line to javadoc saying `See also {@link jakarta.interceptor.Interceptor.Priority}`
That should provide a lead for users but at the same time we won't fully rely on `Interceptor.Priority` javadoc.

manovotn:
Also CC @mkouba as an FYI because you had some comments in the original issue (#496) and we discussed this recently.

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -596,4 +596,33 @@ As specified in <<builtin_contexts>>, contexts associated with built-in normal s
...
+=== Observable container lifecycle events
+
+[[startup_event]]
+
+==== Startup event
+
+Implementations are required to _synchronously_ fire an event with payload `jakarta.enterprise.event.Startup` and qualifier `jakarta.enterprise.inject.Any` during application initialization.
+This event is fired after the event with qualifier `@Initialized(ApplicationScope.class)` but before processing requests.
+
+This event can be observed by integrators and libraries to perform any kind of early initialization as well as by users as a reliable entry point for when the CDI container is ready to process requests.
```
Ladicek, line 610:
```suggestion
This event can be observed by integrators and libraries to perform any kind of early initialization as well as by users as a reliable entry point for when the CDI container is ready.
```

We did this in the javadoc, let's do it here as well.

Ladicek:
Actually I just realized one thing: we should add a section about this to CDI Full where we specify the relationship of these events to container lifecycle events (that may only be declared on Portable Extension), notably `AfterDeploymentValidation` vs  `Startup` and `BeforeShutdown` vs. `Shutdown`.

Now I'm not sure what's the relationship between `@Initialized(ApplicationScoped.class)` and `AfterDeploymentValidation`, but I think `Startup` should probably be _after_ `AfterDeploymentValidation`. Similarly, I'm not sure what's the relationship between `@BeforeDestroyed(ApplicationScoped.class)` and `BeforeShutdown`, but I _think_ that `Shutdown` should be _before_ `BeforeShutdown`. But that's a naive interpretation and probably needs more thinking.

spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -596,4 +596,33 @@ As specified in <<builtin_contexts>>, contexts associated with built-in normal s
 * Otherwise, if the observer method is any other kind of transactional observer method, it is called in an unspecified transaction context, but with the same lifecycle contexts as the transaction that just completed.
 * Otherwise, the observer method is called in the same transaction context and lifecycle contexts as the invocation of `Event.fire()` or `BeanManager.getEvent()`.
 
+[[observable_container_lifecycle_events]]
+
+=== Observable container lifecycle events
```
Ladicek, line 601: Since "Container lifecycle events" belong to Portable Extensions, should we call these differently? Maybe "Application lifecycle events"? The Packaging and Deployment chapter has sections "Application initialization lifecycle" and "Application shutdown lifecycle", which is a nice match. (These sections don't have to refer to these new event types, as you described.)

Ladicek (approved):

PR #544, opened 2021-10-18
improve extension API and resolve all remaining open questions

Ladicek:
- settled on the name "Build Compatible Extensions"
- renamed `@Processing` to `@Registration`, because if we ever want
  to introduce an extension phase for modifying beans before they
  are registered, _that_ should be called `@Processing`
- moved the synthetic bean creation/destruction functions from
  the `CreationalContext` style to the `Instance` style
- specified precisely how `ClassInfo.methods` and `fields` work,
  specified that `ClassConfig` behaves identically, and specified
  that an extension methods with parameters of type `MethodInfo`
  or `FieldInfo` also behave identically
- removed the `@ExactType` and `@SubtypesOf` annotations and moved
  type queries directly to the `@Enhancement` and `@Registration`
  annotations
- moved the `MetaAnnotations` API from accepting a callback
  to directly returning `ClassConfig`
- specified lifecycle of `SyntheticBean{Creator,Disposer}` and
  `SyntheticObserver`
- specified that `Messages.error` is treated as a deployment problem


Ladicek:
@graemerocher There are 2 changes here that are worth highlighting.

The `MetaAnnotations` API now returns `ClassConfig` instead of accepting a `Consumer<ClassConfig>`. I investigated if that can be implemented in Quarkus and it can -- but if that would cause trouble for you, we can move back to the callback style.

The other thing is that `SyntheticBean{Creator,Disposer}` now take an `Instance<Object>` instead of a `CreationalContext`. As such, they are much nicer to use. If you already have the implementation using `CreationalContext`, then the `Instance`-based API can be implemented on top of it like this:

1. The `Instance` is very similar to what `BeanContainer.createInstance()` returns, except:
    1. it must not have its own `CreationalContext`, it must use the `CreationalContext` for the synthetic bean;
    2. it must not have its own "current `InjectionPoint`", it must use the "current `InjectionPoint`" from when the function is called;
    3. if the synthetic bean is not `@Dependent`, trying to lookup `InjectionPoint` should throw.
2. The creator and disposer function share the same `CreationalContext`, so it's fine to create a separate `Instance` for each invocation.

Ladicek:
BTW, this completes the extension API (at least from my side). I'll write the accompanying specification text next.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Enhancement.java
```diff
@@ -1,36 +1,77 @@
...
+
+    /**
+     * Narrows down the set of <em>expected types</em>, defined by {@link #types() types}
+     * and {@link #withSubtypes() withSubtypes}, to types that use given annotations.
+     * The annotations can appear on the type, or on any member of the type,
+     * or any parameter of any member of the type.
+     *
+     * @return types of annotations that must be present on the <em>expected types</em>
+     */
+    Class<? extends Annotation>[] withAnnotations() default Annotation.class;
```
graemerocher, line 76: I'm still concerned this defaults to any annotation. It should maybe be clarified that if not specified then only types with at least one bean defining annotation will be visited?

Ladicek: Also, I just noticed that Portable Extensions' `@WithAnnotations` also treats a type annotated with _given annotation_ if the type is annotated with some _other_ annotation that is, in turn, meta-annotated with _given annotation_. Do we want to add that too?

graemerocher: yes stereotypes are fine

api/src/main/java/jakarta/enterprise/inject/build/spi/AnnotationBuilder.java
```diff
@@ -1,4 +1,4 @@
-package jakarta.enterprise.inject.build.compatible.spi;
+package jakarta.enterprise.inject.build.spi;
```
Emily-Jiang, line 1: As mentioned on today's call, after the rename, it sends the impression the package is build SPIs, which are different from the attended purpose (works under both runtime and build time). If lite is not suitable to be used in the package name, how about `core` instead?

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildCompatibleExtension.java
```diff
@@ -3,34 +3,34 @@
...
- * All the parameters will be provided by the container when the extension is invoked.
+ * Extension methods may declare arbitrary number of parameters. Which parameters may be declared depends
+ * on the particular execution phase and is documented in the corresponding extension annotation.
+ * All the parameters will be provided by the container when the extension method is executed.
  * <p>
- * Extension methods can be assigned a priority using {@link jakarta.annotation.Priority @Priority}.
+ * Extension methods may be assigned a priority using {@link jakarta.annotation.Priority @Priority}.
  * Extension methods with smaller priority values are invoked first. Extension methods without specified priority
  * have a default priority of {@link jakarta.interceptor.Interceptor.Priority#APPLICATION Priority.APPLICATION} + 500.
  * If two extension methods have equal priority, the ordering among them is undefined. Note that priority
```
Emily-Jiang, line 28:
```suggestion
 * If two extension methods have equal priority, the ordering between them is undefined. Note that priority
```

Ladicek: Fixed.

PR #547, opened 2021-10-21
Change BeanManager treatment in CDI Lite

Ladicek:
Previously, the specification downright disallowed obtaining
a `BeanManager` in CDI Lite. This makes portability from CDI Full
to CDI Lite harder, which also manifests in the CDI TCK.

Wit this commit, CDI Lite applications may obtain a `BeanManager`,
but may only call methods inherited from `BeanContainer`.


graemerocher (approved):

manovotn (approved):
Awesome! This is going to save us some headaches in TCKs.

PR #560, opened 2021-11-22
Language model and build compatible extensions API clarifications

Ladicek:
- renamed `BeanInfo.getName()` to `name()` for consistency
- renamed `ObserverInfo.observedType()` to `eventType()` for clarity
- added `SyntheticComponents.addObserver(Type)` to allow registering
  synthetic observers of parameterized types
- changed `WildcardType` representation to always include an upper
  bound; the `? extends Object` type is no longer represented as `?`
  because the upper bound may be annotated
- improved specification of `AnnotationTarget`, especially around
  repeatable annotation types
- improved specification of `ClassInfo` and `MethodInfo`, especially
  around implicit declarations and `abstract` classes/methods

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
@@ -127,7 +127,7 @@ public interface BeanInfo {
      *
      * @return the bean name, or {@code null} if the bean does not have a name
      */
-    String getName();
+    String name();
 
     /**
      * Returns the {@linkplain DisposerInfo disposer} method of this producer-based bean.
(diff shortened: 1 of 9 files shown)


Ladicek:
These changes are motivated by my TCK work. It should be the last batch of changes, at least to the language model, whose TCK should now be complete (https://github.com/eclipse-ee4j/cdi-tck/pull/310). I'll continue working on the build compatible extensions tests, which may motivate some changes in the extension API, but hopefully minimal.

graemerocher (approved):

Ladicek:
Actually I just realized that the commit message is wrong, the `WildcardType` does not always include an _upper_ bound. It's that it always includes _some_ bound -- either lower or upper. The unbounded wildcard is no longer represented as without bounds.

PR #565, opened 2021-11-28
Add module-info to the api artifacts, update to Java SE 11

starksm64:
This relies on the staged 2.0.1-RC1 jakarta.interceptor:jakarta.interceptor-api artifact


api/src/main/java/module-info.java
```diff
@@ -0,0 +1,26 @@
...
+    requires transitive jakarta.interceptor;
+    requires transitive jakarta.cdi.model;
+    requires transitive jakarta.inject;
+    requires static jakarta.el;
+    // For javadoc
+    requires static java.naming;
+    //TODO: requires static jakarta.transation;
+
+    uses jakarta.enterprise.inject.se.SeContainerInitializer;
+    uses jakarta.enterprise.inject.spi.CDIProvider;
```
Ladicek, line 25: Maybe also `jakarta.enterprise.inject.build.compatible.spi.BuildServices`? I admit I don't fully understand the service provider stuff in JPMS.

lang-model/src/main/java/module-info.java
```diff
@@ -0,0 +1,5 @@
+module jakarta.cdi.model {
```
Ladicek, line 1: I'd probably call it `jakarta.cdi.lang.model`.

Ladicek:
IIUC, this also makes `javac` emit Java 11 bytecode?

Not sure if that's a good thing or a bad thing, but it should be a conscious choice.

starksm64:
> IIUC, this also makes `javac` emit Java 11 bytecode?
> 
> Not sure if that's a good thing or a bad thing, but it should be a conscious choice.

It is essentially required as we depend on artifacts that are being built with Java 11 (annotations 2.1), and it is the baseline Java version for all major specification versions going into EE10.

Ladicek:
I mean, I'm fine with making it 11+ only, especially if we don't have a choice :-) But it would be good if it was highlighted in the PR title / commit message.

api/src/main/java/module-info.java
starksm64: You need a uses statement for any services loaded via a ServiceLoader, so yes, we should have BuildServices uses

lang-model/src/main/java/module-info.java
starksm64: Ok, updated.

starksm64:
Ok, added updating to Java SE 11 in PR title.

Ladicek (approved):
Looks fine to me, though my understanding of JPMS is virtually zero :-)

Ladicek:
I'm keeping this open for more review.

I originally thought maybe this should be a draft because it uses an RC version of Interceptors, but we already have a non-final version of Annotations, so that's fine.

Also looking at the list of commits, I think I'd merge this using "Squash and merge".

manovotn (approved):
LGTM  although my understanding of JPMS is not the deepest.

starksm64:
There still an outstanding question about moving to the Java SE 11 base, but we will go with it for the RC so that we can get feedback on potential problems.

PR #566, opened 2021-11-29
Allow @Enhancement to not restrict the types based on annotations

### LINKED ISSUE #564: @Enhancement(withAnnotations = ...) restricts usability of non-annotated classes added via @Discovery
We (actually @manovotn) just found that the `@Enhancement` annotation's `withAnnotations` member restricts how non-annotated classes may be used. The way `withAnnotations` is currently specified, it expects that all classes processed via `@Enhancement` have _some_ annotation (preferrably a bean defining annotation, but `withAnnotations = Annotation.class` means _any_ annotation is enough). That constraint exists for annotation processors.

That seems entirely reasonable for classes that are subject to normal bean discovery, because CDI Lite only has annotated discovery. However, for classes that are added to bean discovery through `@Discovery`, maybe that is too strict? What if I want to add a non-annotated class through `@Discovery` and then give it a bean defining annotation in `@Enhancement`?

@graemerocher do you think that would be possible in an annotation processor?

If so, I can see several options:

- add yet another marker annotation similar to `Enhancement.BeanDefiningAnnotations`, something like `Enhancement.NoneRequired` (this is still constrained to _discovered_ classes, as currently defined, no change on that front);
- make `withAnnotations = {}` mean "any classes, even classes without annotations" (this is still constrained to _discovered_ classes, as currently defined, no change on that front);
- some other way to express this?

We may also want to change the default. Currently, `withAnnotations` defaults to `BeanDefiningAnnotations.class`, which makes sense in my opinion, but diverges from what Portable Extensions do (there, there's no annotation constraint by default).

If not, then ... well ... maybe we could make `ScannedClasses.add()` return a `ClassConfig` to be able to add annotations in the `@Discovery` phase, before `@Enhancement`? (I'd prefer not, but if it would be the only way, then it's better than nothing.)

If not even that, then I don't know and we have a problem :-)

Ladicek:
Fixes #564

I intentionally split this into 2 commits:

1. Make `@Enhancement(withAnnotations = {})` mean "no restriction on annotations". That is, types without annotations added through `@Discovery` may be subject to `@Enhancement`. The default is still `BeanDefiningAnnotations.class`.
2. Chagne the default value to an empty array, to align with Portable Extensions.

That's because I'm pretty sure we'd have a consensus on item 1, but not necessarily on item 2.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Enhancement.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Enhancement.java
@@ -13,7 +13,7 @@ import java.lang.annotation.Target;
  * In the following text, the term <em>expected types</em> denotes the set of types defined by
  * the {@link #types() types}, {@link #withSubtypes() withSubtypes} and {@link #withAnnotations() withAnnotations}
  * members of the {@code @Enhancement} annotation. The term <em>discovered types</em> denotes
- * the subset of <em>expected types</em> that were discovered during bean discovery.
+ * the subset of <em>expected types</em> that were discovered during type discovery.
  * <p>
  * Methods annotated {@code @Enhancement} must declare exactly one parameter of one of these types:
  * <ul>
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Enhancement.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Enhancement.java
@@ -72,20 +72,24 @@ public @interface Enhancement {
      * parameter of any member of the type, or as a meta-annotation on any annotation
      * that is considered by these rules.
      * <p>
-     * Defaults to the {@linkplain BeanDefiningAnnotations set of bean defining annotations}.
+     * If empty, the set of <em>expected types</em> is not narrowed down in any way.
+     * If {@code java.lang.Annotation} is present, the set of <em>expected types</em>
+     * is narrowed down to types that use any annotation.
+     * The {@link BeanDefiningAnnotations @BeanDefiningAnnotations} marker type may
+     * be used to narrow down the set of <em>expected types</em> to types that use
+     * any bean defining annotation.
      * <p>
-     * If empty, or if {@code java.lang.Annotation} is present, all annotations are used.
-     * That is, the set of <em>expected types</em> is narrowed down to the set of types
-     * that use any annotation.
+     * Defaults to the {@linkplain BeanDefiningAnnotations set of bean defining annotations}.
      *
      * @return types of annotations that must be present on the <em>expected types</em>
      */
     Class<? extends Annotation>[] withAnnotations() default BeanDefiningAnnotations.class;
 
     /**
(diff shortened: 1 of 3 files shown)


graemerocher:
It should be noted somewhere that if a bean defining annotation is not specified then then enhancement applies to all types discovered via annotations or added via the discovery phase, but not every type in the application

Ladicek:
I think that directly follows from the `@Enhancement` javadoc/specification. It's probably not the best text I ever wrote, but it should be very clear in that `@Enhancement` only ever applies to types discovered during bean discovery.

Ladicek:
Though currently the text says "that were discovered during bean discovery", which is something I discussed with @manovotn  and @mkouba at length on Friday. I'll change that to say "during type discovery".

Ladicek:
Added a commit with clarifications of type discovery / bean discovery.

graemerocher (approved):

manovotn (approved):
Looks good, thanks @Ladicek!

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Enhancement.java
```diff
@@ -72,20 +72,25 @@
...
+    Class<? extends Annotation>[] withAnnotations() default {};
 
     /**
-     * Marker annotation type that represents set of bean defining annotations after
-     * the {@link Discovery @Discovery} phase is finished. That is, it includes custom
-     * normal scope annotations as well as custom stereotypes.
+     * Marker annotation type that, for the purpose of {@link Enhancement#withAnnotations()},
+     * represents set of bean defining annotations after the {@link Discovery @Discovery}
+     * phase is finished. That is, it includes custom normal scope annotations as well as
+     * custom stereotypes.
```
manovotn, line 93: @Ladicek I have been thinking about this some more and I am not sure if `BeanDefiningAnnotations`  makes much sense with latest change (i.e. if it is justifiable to have that coded into spec).
Setting aside that I am not sure how well can we filter out classes with just custom stereotypes (which I will try to figure tomorrow), in what case would you want to filter for these classes? With annotated mode, this basically translates to "all classes". The only ones which are potentially omitted are those added by build compatible extensions that that don't have any annotation on them. That, to me, seems like a weird filter but maybe I am just not seeing the use case.

graemerocher: The idea was to indicate document that in annotated mode only classes with at least a bean definition annotation are processed. Having said that there is no technical reason for needing the annotation itself and maybe it is just documentation.

Ladicek: I had to think about this for a while, but @manovotn is right. During type discovery, we discovery classes with a BDA and classes added via extension in `@Discovery`. Classes added via extension automatically get `@Dependent` if they don't have a BDA already. So saying "only process classes that were discovered during type discovery and have a BDA" is basically equivalent to "only process classes that were discovered during type discovery".

PR #573, opened 2021-12-03
Improve ambiguous dependency resolution spec working

scottkurz:
Signed-off-by: Scott Kurz <skurz@us.ibm.com>

I find this wording confusing and am at least trying to understand if not clarify the intent.
This might seem pedantic but maybe this is just a beginner's POV.

Does my new text capture the correct meaning?  If so, I think it is a better wording.

If we say "A" is the set of  "eligible beans that are not alternatives selected for the application" and "B" is the set of " producer methods and fields of beans that are alternatives."... then I think the spec is saying the CDI impl eliminates the set A and takes the set B.  

But you can read this a certain way to say that the CDI impl eliminates "A - B", i.e. that B is some subset of A that we do NOT eliminate.    And maybe you're thinking that makes no sense.. but I'm just saying it could be clearer.

Of possibly I am completely misunderstanding.

### DIFF
--- a/spec/src/main/asciidoc/core/injectionandresolution.asciidoc
+++ b/spec/src/main/asciidoc/core/injectionandresolution.asciidoc
@@ -106,7 +106,7 @@ An _unsatisfied dependency_ exists at an injection point when no bean is eligibl
 An _ambiguous dependency_ exists at an injection point when multiple beans are eligible for injection to the injection point.
 
 When an ambiguous dependency exists, the container attempts to resolve the ambiguity.
-The container eliminates all eligible beans that are not alternatives selected for the application, except for producer methods and fields of beans that are alternatives.
+The container eliminates all eligible beans that are not alternatives selected for the application, and selects only those producer methods and fields of beans that are alternatives.
 If:
 
 * there is exactly one bean remaining, the container will select this bean, and the ambiguous dependency is called resolvable.
--- a/spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
+++ b/spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
@@ -149,7 +149,7 @@ A special set of rules, defined in <<delegate_assignable_parameters>>, apply if 
 In addition to rules defined in <<ambig_names>>, the following rules apply.
 
 When an ambiguous name exists, the container attempts to resolve the ambiguity.
-The container eliminates all eligible beans that are not alternatives selected for the bean archive or selected for the application, except for producer methods and fields of beans that are alternatives.
+The container eliminates all eligible beans that are not alternatives selected for the application, and selects only those producer methods and fields of beans that are alternatives.
 
 [[client_proxies_full]]
 


spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -106,7 +106,7 @@ An _unsatisfied dependency_ exists at an injection point when no bean is eligibl
 An _ambiguous dependency_ exists at an injection point when multiple beans are eligible for injection to the injection point.
 
 When an ambiguous dependency exists, the container attempts to resolve the ambiguity.
-The container eliminates all eligible beans that are not alternatives selected for the application, except for producer methods and fields of beans that are alternatives.
+The container eliminates all eligible beans that are not alternatives selected for the application, and selects only those producer methods and fields of beans that are alternatives.
```
Ladicek, line 109: I don't think this is an equivalent change. Say I have the following set of beans that are eligible for injection to one injection point:

* `A`: class-based bean, not alternative
* `B`: class-based bean, alternative, but not selected (doesn't have `@Priority`)
* `C`: class-based bean, alternative, selected for the application (has `@Priority`)
* `D`: producer-based bean, not alternative
* `E`: producer-based bean, alternative, but not selected
* `F`: producer-based bean, alternative, selected for the application

Then, in this first step, the container is supposed to eliminate `A` (not an alternative), `B` (not selected), and `D` (producer, but not an alternative). Beans `{C, E, F}` remain. Under your proposal, `C` would be eliminated as well.

Ladicek: (I'm actually not exactly sure what should happen to `E`, or if `E` may even exist, considering the below expectation that the remaining beans should have a priority. But that's orthogonal to the present question :-) )

scottkurz: Thank you for explaining @Ladicek .   

I wasn't reading that correctly and wasn't getting differentiating between "class" and "producer-based" beans, (lacking the context to appreciate that this was a dimension where things might differ).     

Well, I like the way you explained it to me better ;) but I see there isn't already a "class-based" vs. "producer-based" breakdown already in the spec... and maybe I'm the only one confused, so  thanks for the quick response just for my personal education here.

Ladicek: The terms "class-based bean" or "producer-based bean" are colloquial. The spec calls "class-based beans" _managed beans_, and "producer-based" are often called just _producer methods_ and _producer fields_.

Ladicek: Also I'd like to point out that the phrase "producer methods and fields of beans that are alternatives" may be interpreted in 2 ways:

1. producer methods/fields that define beans that are, themselves, alternatives;
2. producer methods/fields that are declared by a bean that is an alternative

My explanation above implies the 1st interpretation, though I'm not exactly sure that's correct :-) In any case, the explanation still holds.

PR #574, opened 2021-12-07
Move getInjectableReference method from BeanContainer to BeanManager

### LINKED ISSUE #568: BeanContainer#getInjectableReference should probably be on BeanManager instead
`BeanContainer` API currently houses `getInjectableReference(InjectionPoint ij, CreationalContext<?> ctx)` which was originally on `BeanManager`.

After some discussion with @Ladicek we arrived at the conclusion that this doesn't make much sense.
In Lite, you don't have access (via `BeanContainer`) to programmatic creation of `InjectionPoint` because that's based off the former metadata model (`AnnotatedType` and friend). You cannot store them from extensions either so your means of getting hold of them are very limited. Possibly only to creating a class directly implementing an IP which, again, requires usage of the `Annotated*` metamodel.

Therefore, we should move this method back to `BeanManager`.
Note that recently we changed the wording so that implementations in Lite are allowed to support `BeanManager` in some limited form as well meaning that if this functionality turns out to be needed, it can still be implemented without breaking spec compatibility.

manovotn:
Fixes #568

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
@@ -10,7 +10,6 @@ import jakarta.enterprise.inject.AmbiguousResolutionException;
 import jakarta.enterprise.inject.Default;
 import jakarta.enterprise.inject.Instance;
 import jakarta.enterprise.inject.Stereotype;
-import jakarta.enterprise.inject.UnsatisfiedResolutionException;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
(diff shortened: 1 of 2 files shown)


Ladicek (approved):

Emily-Jiang:
Can you explain why the method needs to be moved to BeanManager?

manovotn:
@Emily-Jiang the linked issue explains it - in short it is very hard (in Lite) to get hold of `InjectionPoint` which is a parameter needed for this method. Therefore, you are highly unlikely to invoke this method in purely CDI Lite.
Note that Lite impls can still implement it if they wish to since we allow them to use `BeanManager` which will retain this method anyway.

Emily-Jiang:
okay, thanks @manovotn ! For some strange reason, I did not see the linked issue on this PR when I looked. I might have missed it. Your explanation makes sense.

PR #577, opened 2021-12-07
Move the session scope to CDI Full

Ladicek:
When I was reviewing the remaining CDI 4.0 work, I realized we kept the session scope in CDI Lite. I believe it should be moved to CDI Full, because:

- the session scope is not actually defined at all in Core CDI
- the definition in CDI EE heavily relies on the Servlet API, which isn't even proposed to be part of the Jakarta EE 10 Core Profile (per https://eclipse-ee4j.github.io/jakartaee-platform/jakartaee10/JakartaEE10CoreProfileRelasePlan)
- modern application architectures often don't rely on in-memory HTTP sessions at all (and we've deliberately left passivation out of Lite)

Hence, I'm proposing this PR to remove the session scope from Lite. It stays in Full.

### DIFF
--- a/spec/src/main/asciidoc/core/definition.asciidoc
+++ b/spec/src/main/asciidoc/core/definition.asciidoc
@@ -482,7 +482,7 @@ public class Order { ... }
 
 [NOTE]
 ====
-Conversations are only available in {cdi_full}.
+Session scope and conversation scope are only available in {cdi_full}.
 ====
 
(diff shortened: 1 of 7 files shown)


Emily-Jiang (approved):

manovotn (approved):
Looks good!

manovotn:
Going to merge this, if @graemerocher has any comments/remarks, feel free to leave them here and we can reopen the discussion.

PR #578, opened 2021-12-07
Add ObserverMethod#getDeclaringBean method.

### LINKED ISSUE #563: Portable Extension's `ObserverMethod` doesn't expose the declaring `Bean`
The Build Compatible Extension API has a method `ObserverInfo.bean()` that provides access to the _bean_ that declares given observer. The Portable Extension API only has `ObserverMethod.getBeanClass()` (providing access to the _class_ that declares given observer).

To keep BCExtensions a strict subset of PExtensions, either we have to remove `ObserverInfo.bean()`, or add `ObserverMethod.getDeclaringBean()` (copying the name from Weld's `ObserverMethodImpl`, which does expose the declaring bean).

manovotn:
Fixes #563 

I am deliberately keeping the return value for synth observers undefined because I understand some implementations can choose to return `null` but in Weld this is apparently never the case and instead synth. OMs belong to the extension declaring them.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/spi/ObserverMethod.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/ObserverMethod.java
@@ -56,6 +56,16 @@ public interface ObserverMethod<T> extends Prioritized {
     public Class<?> getBeanClass();
 
     /**
+     * <p>
+     * Obtains the {@linkplain Bean bean} that declares the observer method.
+     * For synthetic observers, the return value is undefined.
+     * </p>
+     *
+     * @return the declaring {@linkplain Bean bean}
+     */
+    public Bean<?> getDeclaringBean();
+
+    /**
      * Obtains the {@linkplain jakarta.enterprise.event observed event type}.
      * 
      * @return the observed event {@linkplain Type type}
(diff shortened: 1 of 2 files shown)


Ladicek:
The Build Compatible Extension API demands that for synthetic observers, `ObserverInfo.bean()` returns `null`. Can that be implemented on top of this?

manovotn:
> The Build Compatible Extension API demands that for synthetic observers, `ObserverInfo.bean()` returns `null`. Can that be implemented on top of this?

@Ladicek that sounds like a mistake we made. I think the declaring class and the declaring bean are both (in case of Portable Extensions) derived from https://github.com/eclipse-ee4j/cdi/blob/master/api/src/main/java/jakarta/enterprise/inject/spi/configurator/ObserverMethodConfigurator.java#L76-L82
According to spec, it'll always result in a non-null value. On top of that, for your StillDI impl, I expect this to default to the class of the Portable Extension via which we implement build compatible extensions.

So the way I see it, we have two options:
* State in Lite it is undefined return value for synth observers
* State in Lite that the return value is the extension defining it
  * I don't think this is possible because in Lite, compatible extensions are *not* beans

Ladicek:
Actually I'm looking at it again and it seems even my `ObserverInfo.isSynthetic()` method implementation is wrong (because `ProcessSyntheticObserverMethod.getAnnotatedMethod()` is non-portable). Wow.

Ladicek:
At the same time, one can distinguish synthetic observers from non-synthetic by the event type, and so it is not possible to invoke any of these non-portable methods. So it's all good.

Ladicek (approved):

Ladicek:
With this change https://github.com/weld/core/pull/2605, all should be fine and this can be merged.

manovotn:
> With this change [weld/core#2605](https://github.com/weld/core/pull/2605), all should be fine and this can be merged.

Thanks a lot for that!

PR #580, opened 2021-12-08
Add explicit notes to APIs that are not required to be present in CDI Lite

Ladicek:


manovotn (approved):
Good thinking, this makes the limitations of CDI API in Lite more visible from plain javadoc!

Ladicek:
Yup, this is just to [slightly] improve usability of the API source code.

PR #581, opened 2021-12-08
Invalid parameter combinations in @Enhancement and @Registration meth…

manovotn:
…ods should be treated as definition errors.

As I pointed out in https://github.com/eclipse-ee4j/cdi-tck/pull/329, I think we should handle these cases as definition errors.


Ladicek (approved):
That's fine with me, though the difference between definition errors and deployment problems is kinda blurry in case of extensions (and especially Build Compatible Extensions, where there's no way for the extension to produce a definition error, unlike Portable Extensions).

manovotn:
`DefinitionException` (thrown in case of definition error) indicated that it's developer's error which is IMO the case here.
See https://github.com/eclipse-ee4j/cdi/blob/master/api/src/main/java/jakarta/enterprise/inject/spi/DefinitionException.java#L25-L29

> (and especially Build Compatible Extensions, where there's no way for the extension to produce a definition error ....

Not sure I follow, there is little difference in producing definition versus deployment error for Build Compatible Extensions. Or am I missing something?

Ladicek:
No, you're not. Sorry for confusion, the remark was only tangentially related.

PR #582, opened 2021-12-11
fixes #567 split CDI-lite into own module

### LINKED ISSUE #567: Move jakarta.enterprise.inject.build.compatible.spi to another spec (jakarta.build.inject?)
Hi all,

jakarta.enterprise.inject.build.compatible.spi has no real link with CDI except a few annotation overlapping so it is really worth extracting it in a "build IoC" dedicated spec to benefit to the whole platform.

The drawbacks are obvious:

1. No way to use "CDI" but you can only use subpart os CDI (build is incompatible with runtime part and the opposite by design and API)
2. No way to use CDI lite with EE - EE design is runtime which makes it strength and enterprise readiness by enabling to work on already compiled code (frameworks, tooling, transversal concerns of companies .... Indeed, EE can become more build oriented but it is a totally different design as shown by CDI lite work and requires a rethinking of the framework so better to not mess up everything and keep API straight forward (and avoid the "oh, even if I put code it is ignored??" on user land)
3. User inconsistency: it is unlikely an user finds it natural until he checks the implementation(s) because the API is exposed as a while but is not used as such as explained
4. "Framework integration with CDI" means nothing since it is either at build time or at runtime (or 2 integrations) so will create a lot of noise
5. There is no technical limitation as of today to use CDI in lite mode without all the new annotations, it got proven by multiple impementations and it can even be *already* reflectionless technically if desired (not sure the gain is worth it but technically we know how to do it)

Indeed the advantage of this package is obvious: design closer to graalvm (which is a good thing)

So overall this task is about not polluting CDI which is a strong, core and widely used basis in the industry and enable users to start PoC-ing build time EE code through an incubating new specification which could connect to CDI by an appendix (proposal).
After 1-2 years we can review what was there and check it is relevant or not to include in CDI but as of today there is no real strong argument to do so but several to not do.

struberg:
Technically there is no dependency from the 'Full' api to the lite api.
CDI-Lite is only an additional mechanism to support build time generation of proxies etc.
It is a kind of build-time replacement of the classic CDI Extension mechanism.
Servers should be free to support Full or Lite, but not forced to implement both.


graemerocher:
@Ladicek this is actually not a bad idea either way since technically these classes are not needed at runtime (except for the synthetic bean stuff maybe that could be a cdi-lite-runtime module?)

manovotn (changes requested):
-1 

> Servers should be free to support Full or Lite, but not forced to implement both.

If this is the motivation for the PR, then I am afraid you misunderstood the concept of Lite.
CDI Lite is designed to be a subset of CDI Full and for that reason any CDI Full implementation needs to support CDI Lite as well.
Note that you don't need to do much to meet that requirement - CDI Lite extension model can be executed/implemented via Portable Extensions. See https://github.com/Ladicek/StillDI which is a PoC to achieve just that. The rest of CDI Lite is a functionality that any existing CDI implementation already has (or should have for that matter).

Apart from the above, there were numerous discussions on how to deliver CDI API jar WRT to Lite addition and if I remember correctly, we agreed that for the first version, we will just keep it in one API jar with the possibility to split it up in a more breaking manner later.
Note that moving just these few classes doesn't solve much because CDI Lite still needs most of the other CDI-related parts.

Furthermore, these classes are actually needed during runtime for any runtime implementation of CDI Lite (be it standalone Lite impl or perhaps CDI Full impl). CDI Lite isn't necessarily always build-time oriented but is designed with that in mind.

Last but not least, @starksm64 is working on a CR release of CDI API (to meet required deadline) and a change like this sounds too invasive at this stage of development.

rmannibucau:
It is a good start to move lite away from the normal api but since it is by design another spec it should move to a build spec since it breaks the full spec and pkatform and not stay in cdi.

Anyway +1000 to *NOT* include it in any new release of cdi since it is a massive shout in the foot of cdi spec and jakarta platform (basically it says dont use us anymore since you cant rely on any investment done on us)

struberg:
> CDI Lite is designed to be a subset of CDI Full and for that reason any CDI Full implementation needs to support CDI Lite as well.

This seems not to be correct. Lite technically needs CDI Full, but not the other way around as I did prove with the PR. The dependencies clearly shows that.

> Note that moving just these few classes doesn't solve much because CDI Lite still needs most of the other CDI-related parts.

Yes, but not the other way around. If someone doesn't run on Quarkus then one doesn't need those APIs. In other words: CDI lite is actually not lighter in terms of code but adds *additional* support for build time generators. That's fine, but the name indicates that 'lite' is a subset whereas in reality it's the other way around.

> Note that moving just these few classes 
Those 'few' classes add about 30%.

> Furthermore, these classes are actually needed during runtime for any runtime implementation of CDI Lite 

Yes, but not for Full. That's the whole point.

> Last but not least, @starksm64 is working on a CR release of CDI API

Is this really an argument for a technical question? What is that for a deadline? It's not a final release yet. Do ship your preview, but we can still do this properly afterwards.

rmannibucau:
It is actually worse: CDI "full" and lite are not interoperable which means both have a clear overlap but none of both can work together at any time so they are really  2 distinct specs (I assume we could extract a cdi-common like spec to share what is shareable but not sure it would help and wouldnt mess more what is in user minds).

Also note that it is not true that the CDI implementation runtime needs these classes, most are only build time classes so very worse case they belong to another jar (just speaking in terms of modularity but the point is that it is not the same spec since it is not compatible with the existing and other package flavor so shouldnt be there at all).

struberg:
I've change the module name from cdi-light to cdi-lite.

manovotn:
> This seems not to be correct. Lite technically needs CDI Full, but not the other way around as I did prove with the PR.

This is simply untrue statement. Lite was deliberately designed to be a *functional* subset of Full and works as such as is captured in API, TCK and compatible implementation(s).
The dependencies you refer to are simply a matter of agreement that we do not want to split CDI JAR in this release as that would be breaking everything. But if we were really going to break it, than we should split all CDI Full parts into separate JARs (and therefore even different packages) and make CDI Full depend on CDI Lite which would then contain the core functionality.
That is how it was designed and intended as I am sure you read in the numerous documents, email conversations, issues,...

> Yes, but not the other way around. If someone doesn't run on Quarkus then one doesn't need those APIs. In other words: CDI lite is actually not lighter in terms of code but adds additional support for build time generators. That's fine, but the name indicates that 'lite' is a subset whereas in reality it's the other way around.

Quarkus is not the only implementation, Micronaut folks started and actively work on OpenDI. See https://projects.eclipse.org/projects/ee4j.odi
And once again, Lite is a functional subset. Build compatible extension are designed to be supported in both environments and can be easily implemented in Full via portable extensions as can be seen in compatible implementation (and the StillDI PoC that initially showed that). Therefore, they are also part of Full, albeit as a new feature compared to previous CDI major version.

> Yes, but not for Full. That's the whole point.

Yes, for Full. That was the whole point ever since the work on Build Compatible Extensions started - for them to be available and usable in both approaches, if the user wishes to go that way.

> Is this really an argument for a technical question? What is that for a deadline? It's not a final release yet. Do ship your preview, but we can still do this properly afterwards.

No, it's not. It is a note saying that the design, discussion and development across CDI, TCK and at least three implementations is ongoing for a long period of time and has reached a point where we have it working and in CR state for Jakarta EE 10. Simple as that.

> It is actually worse: CDI "full" and lite are not interoperable which means both have a clear overlap...

As stated above, CDI Full contains CDI Lite and Weld 5 implements both, so I am not sure what is the "not interoperable" part supposed to mean?

> Also note that it is not true that the CDI implementation runtime needs these classes, most are only build time classes so very worse case they belong to another jar

Your statement is based on false presumption that CDI Full doesn't support that. As stated above, it does, and as such it needs those classes at runtime. Again, Weld 5 is a proof of that.

> Tempted to answer that you got warned this design was not working month ago but didn't listen...

The design works just fine - we have APIs, TCKs and implementations. And we had them, or at least PoC of them, even that month ago so pardon me for now listening but I am still not seeing that grave design flaw you are talking about.

rmannibucau:
How do you reconcile runtime and build time extensions? How do you make discovery unified? How to you make build-less (what jakarta platform targets for next release and tck) working for a build only api?
How do you provide an unified api at build time compatible with annot proc round trips to integrate with java instead of reinventing another partial build time api? How do you make build time extensions running properly with existing cdi extensions frameworks made? How do you integrate it with the platform (jaxrs, jbatch, jsf, ...)? How do you respect the type safety of cdi - its root core - if you have 2 envs to unify with no guarantee between both? How can you claim to be CDI compatible if using CDI makes you no more compatible with CDI (indeed add full and lite but it is the mess you create for users). And so on...

Last, which use case do you enable? Configure at buld time CDI? It is already doable and done by vendors. Becoming native? It is already done. Anything else you didnt mention and worth duplicating so much API?

Please dont mix "we dont do" and "we cant do". For ex quarkus does not support decorators but it can to show a trivial example.

struberg:
> Lite was deliberately designed to be a _functional_ subset of Full

No, it is a superset of current CDI. It is NOT lighter than current CDI. It adds 30% more complexity. I personally don't get the gain because already now one can easily do build-time generation and then let this run in a standard CDI-2.0 container. And if one says he's fine with not using Extensions, then he can even run pretty much free of reflection. This is now Arthur, Helidon etc do work.

> That is how it was designed and intended as I am sure you read in the numerous documents, email conversations, issues,...

I read it and spoke out against it many times. Like many others did. You did choose to ignore the opinion of all those people. I totally understand and support to explore new grounds and play around what can be done. But when it comes to the final spec we really must review the findings and make honest decisions. "We spent so many hours" is simply not an argument. There are many implementations which did easily manage to run on GraalVM without those APIs. Fine to keep it as optional API nonetheless if some vendors want it, but it should not be forced on Full CDI imo.

Ladicek:
Here's a short historical digression. The original idea was that CDI Lite would be just a subset of the CDI API, without extensions. People complained that they need extensions, so a new extension API was born (because no matter how loud you say that Portable Extensions can easily be implemented at build time, it just won't become true). The original idea of that API was that it would only be part of CDI Lite and CDI Full would not have to implement that. People complained that Lite would no longer be a subset of Full, so the API was reshaped a little to make sure it can be implemented on top of Portable Extensions. People should make up their minds :-)

One thing that deserves extra clarification:

> How do you reconcile runtime and build time extensions? [...]

You don't. In Weld, Build Compatible Extensions are executed together with Portable Extensions. (In fact, they are executed _by_ a Portable Extension.) Build Compatible Extensions are build _compatible_, not build _only_.

Hence, users can have Portable Extensions and Build Compatible Extensions on their classpath (hand-waving a little here), and they will all be processed at the same time. There's also an annotation to skip a Build Compatible Extension if a corresponding Portable Extension is present, allowing libraries to ship both.

> > Lite was deliberately designed to be a _functional_ subset of Full
> 
> No, it is a superset of current CDI.

Those two sentences are not in contradiction. Core CDI in version 4.0 is obviously a superset of Core CDI in 3.0 (or 2.0). A subset of _that_ is called CDI Lite.

rmannibucau:
@Ladicek well, you well explained why it can't sit in CDI, because 1. it breaks EE "runtime" design - which is important in enterprises, 2. it breaks CDI itself since you never reconcile both worlds as expected and you can't use build time api at runtime so it *must* not be there in cdi-api *by design* so it is an optional CDI build time API (likely a vendor one since it is at build by design in EE)

So it must be at least extracted as an optional layer integrating with CDI like JAX-RS does for example (in terms of integration) and not be in cdi-api at all as you explained and by CDI design.

Ladicek:
I have explained that the build _compatible_ API can be used at runtime, is supposed to be used at runtime by runtime implementations of CDI, and is actually used at runtime by Weld. There is no expectation that build compatible extensions are always executed during build, only to be reconciled later with whatever portable extensions do. If an implementation decides to do it anyway, it's their job to define, implement and document their semantics and limitations -- but, again, there's nothing in CDI that would suggest that this is required or even recommended. All in all, it 1. doesn't break any runtime design, because there are no build time requirements, 2. doesn't break CDI, because there are no reconciliation requirements, 3. belongs to CDI just fine.

rmannibucau:
The reconcialiation requirement is in the cdi "current" api with extension design. Lite breaks that by design. Look quarkus for an example. Another thing is the goal of graalvm compat is not reached as you explains cause you need the runtime. So it is just noise in the api and no new feature is added so no need to x2 the api for no user gain so please drop this useless package from what is called "cdi" and move it to another spec.

struberg:
+1 It's literally optional! @Ladicek as you said yourself, it even can be done in a portable extension.

Ask yourself the question: is ANY of those classes neeed at runtime by the user code? Clearly not!

manovotn:
This PR is no longer relevant because CDI 4.0 has been released some time ago.
Closing it.

PR #586, opened 2022-01-05
Change parameter type of SkipIfPortableExtensionPresent from String to Class

### LINKED ISSUE #585: The parameter of @SkipIfPortableExtensionPresent should have type Class
I am referring to [this annotation](https://github.com/eclipse-ee4j/cdi/blob/4.0.0-RC1/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SkipIfPortableExtensionPresent.java).

Currently, it has a parameter of type `String` which makes it rather inconvenient to use as you need to spell the whole FQCN. The ability to use `Class` instead would make it way easier.

Now, I am not sure if we chose `String` for some specific purpose so please do jump in and tell me if you recall that. But thinking on it some more, this annotation is only ever used in CDI Full where portable extension and build compatible extension co-exist so it shouldn't be blocking any build oriented approach when we use `Class` parameter.

manovotn:
Fixes #585


Ladicek (approved):
I originally thought it would be just `Class<?>`, but this is even better, thanks!

PR #590, opened 2022-01-31
Update package version

Emily-Jiang:
The package versions should be updated accordingly for this new release. At the moment, they were still on the same version as the CDI 3.0 release. Since this spec does not use semantic versioning, I have simply updated the major versions.


Ladicek (approved):
Makes sense to me.

manovotn (approved):

PR #601, opened 2022-04-07
Clarify handling exceptions during async events

### LINKED ISSUE #599: Clarify handling exceptions during async events
The specification [has a chapter on this](https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#async_exception) but it is IMO not entirely clear because it only covers the case where more than one exception is thrown.

The TCK likewise only covers such scenario, see https://github.com/eclipse-ee4j/cdi-tck/tree/master/impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/async/handlingExceptions

We should change the chapter title to something like `Handling exceptions thrown during an asynchronous event` and in the text we need to either explicitly state that the cause is always `CompletionException` or just mention what happens in case there is only one exception.

Afterwards, we should add a TCK test for it.

Note that this was originally brought up in Quarkus issue - https://github.com/quarkusio/quarkus/issues/24646

manovotn:
Fixes #599 

I've reworded the chapter to make it clearer that `CompletionException` is always expected.

FYI @mkouba


spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -536,9 +536,9 @@ If the exception is a checked exception, it is wrapped and rethrown as an (unche
 
 [[async_exception]]
 
-==== Handling multiple exceptions thrown during an asynchronous event
+==== Handling exceptions thrown during an asynchronous event
 
-If an event is asynchronous, and an exception is thrown by any of its notified observers, the `CompletionStage` returned by `fireAsync` will complete exceptionally with `java.util.concurrent.CompletionException` exception.
+If an event is asynchronous, and an exception is thrown by one or more of its notified observers, the `CompletionStage` returned by `fireAsync` will complete exceptionally with `java.util.concurrent.CompletionException` exception as its cause.
```
Ladicek, line 541: What does the "as its cause" at the end mean?

manovotn: Fair enough, changed the wording

Ladicek (approved):

PR #609, opened 2022-05-16
Remove dependency on JTA API

### LINKED ISSUE #608: Dependency on outdated JTA API
The latest release of CDI 4 has a dependency on `jakarta.transaction-api` in version `2.0.0`.
That artifact is actually not Jakarta EE 10 (there is no such artifact as of now it seems?) and it in turn depends on CDI 3.0, see [this POM](https://github.com/jakartaee/transactions/blob/2.0.0/api/pom.xml#L280-L284).
The dependency is not even `optional`.

This causes a cyclic dependency that has been mentioned in the jakarta-platform mailing list - [link to the archive](https://www.eclipse.org/lists/jakartaee-platform-dev/msg03409.html). This is nothing new and the dep has existed prior to our release.

As far as I can tell, we could completely remove the JTA API dependency as it is only used in javadoc inside `TransactionPhase` enum.
Granted, the javadoc won't be as nice but we'll get rid of the dep altogether. OTOH, JTA certainly cannot remove dep on CDI API, so we are the only ones who can break that cycle.

So we have two choices:
* Remove the dependency
* Upgrade the dependency (once there is EE 10 version)

Personally, I think we can easily remove the dependency.

Either way, we should then have an SP release and I am not sure what would that take- cc @starksm64.

manovotn:
Fixes #608 

This is one of the ways to address it as we don't really need the dependency on JTA apart from javadoc of a single class.

(FYI @arjantijms)

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -135,7 +135,6 @@
 		<interceptor.api.version>2.1.0</interceptor.api.version>
 		<!-- These are only used in javadoc links -->
 		<ejb.api.version>4.0.0</ejb.api.version>
-		<transaction.api.version>2.0.0</transaction.api.version>
 
 		<maven.compiler.release>11</maven.compiler.release>
 		<maven-bundle-plugin.version>5.1.2</maven-bundle-plugin.version>
@@ -183,13 +182,6 @@
 				<artifactId>jakarta.interceptor-api</artifactId>
 				<version>${interceptor.api.version}</version>
 			</dependency>
-
-			<!-- Only for javadoc references -->
-			<dependency>
-				<groupId>jakarta.transaction</groupId>
-				<artifactId>jakarta.transaction-api</artifactId>
-				<version>${transaction.api.version}</version>
-			</dependency>
 		</dependencies>
 	</dependencyManagement>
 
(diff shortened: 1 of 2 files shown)


manovotn:
I have also noticed that we are declaring EJB API dependency directly inside maven javadoc plugin.
Maybe that would be a better option? We'd still be keeping the cyclic dep and would, by definiton, have an outdated version there on release.

Ladicek:
Sounds like a good idea, but I wouldn't mind if we got rid of the EJB dependency as well :-)

Ladicek (approved):

Ladicek:
Not merging yet, as I assume we'll wanna discuss this on the CDI call tomorrow.

manovotn:
> Sounds like a good idea, but I wouldn't mind if we got rid of the EJB dependency as well :-)

Yeah, that's exactly what I wanted to bring up in the meeting tomorrow

arjantijms (approved):

arjantijms:
Thanks!

Emily-Jiang (approved):

starksm64 (approved):

PR #612, opened 2022-05-18
Remove javadoc dependency on EJB

manovotn:
Second part of #608


api/src/main/java/jakarta/enterprise/inject/spi/SessionBeanType.java
```diff
@@ -24,15 +24,15 @@
  */
 public enum SessionBeanType {
     /**
-     * A {@linkplain jakarta.ejb.Stateless stateless} session bean
+     * A {@code jakarta.ejb.Stateless stateless} session bean
```
Ladicek, line 27: Ugh, how does this render?

> A `jakarta.ejb.Stateless stateless` session bean

is probably not what we want :-)

Ladicek: I'd say we should have something like

> A stateless session bean

or

> A stateless (`jakarta.ejb.Stateless`) session bean

or

> A `@Stateless` session bean

or

> A `@jakarta.ejb.Stateless` session bean

manovotn: Oops, should have finished my coffee before commiting. Fixed.

Ladicek (approved):

PR #617, opened 2022-05-24
add BuildServicesResolver.setBuildServices()

Ladicek:


api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildServicesResolver.java
```diff
@@ -12,12 +12,13 @@
...
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.Objects;
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.TreeSet;
 
-final class BuildServicesResolver {
+public final class BuildServicesResolver {
```
graemerocher, line 21: perhaps now that this is public it needs type level javadoc

Ladicek: Good point.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildServicesResolver.java
```diff
@@ -61,4 +62,12 @@ private static void discoverFactories() {
 
         BuildServicesResolver.discoveredBuildServices = Collections.unmodifiableSet(factories);
     }
+
+    /**
+     * This method should <em>not</em> be used by applications. It is only exposed for integrators
+     * with complex classloading architectures, where service loader lookup doesn't work out of the box.
```
arjantijms, line 68: As a public method, also needs the documentation for the parameter. Javadoc lint will likely also complain about this.

manovotn (commented):
@Ladicek should we also include the policy to only be able to set the impl only once?
We do the same thing for [CDI providers](https://github.com/jakartaee/cdi/blob/master/api/src/main/java/jakarta/enterprise/inject/spi/CDI.java#L120).

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildServicesResolver.java
Ladicek: Good point too.

Ladicek:
> @Ladicek should we also include the policy to only be able to set the impl only once? We do the same thing for [CDI providers](https://github.com/jakartaee/cdi/blob/master/api/src/main/java/jakarta/enterprise/inject/spi/CDI.java#L120).

I considered that, but contrary to your statement (and contrary to the javadoc you link to), I didn't find `CDI.setCDIProvider` to check this.

manovotn:
> > @Ladicek should we also include the policy to only be able to set the impl only once? We do the same thing for [CDI providers](https://github.com/jakartaee/cdi/blob/master/api/src/main/java/jakarta/enterprise/inject/spi/CDI.java#L120).
> 
> I considered that, but contrary to your statement (and contrary to the javadoc you link to), I didn't find `CDI.setCDIProvider` to check this.

Indeed, you're correct. I only read the javadoc and assumed that.
It seems like a bug TBF

Emily-Jiang (approved):

manovotn (approved):

manovotn:
This change was discussed during the meeting and agreed on as a way forward.
We want to keep it as similar to how setting CDI provider works as possible (including the current CL used for loading the services).

@arjantijms do you need a service release with this fix or are you going to use the workaround mentioned in the issue?

arjantijms:
> do you need a service release with this fix or are you going to use the workaround mentioned in the issue?

I will use the workaround mentioned in the issue for now, so a service release is not strictly needed from my part.

PR #619, opened 2022-05-25
Make sure CDI and BuildServicesResolves both guard against repeated i…

### LINKED ISSUE #618: CDI.setCDIProvider has different behavior than the javadoc states
`CDI.setCDIProvider` states that you can only set the provider once but the actual code doesn't check this at all.
Code link - https://github.com/jakartaee/cdi/blob/master/api/src/main/java/jakarta/enterprise/inject/spi/CDI.java#L120

The javadoc makes sense and the code should IMO adhere to it - there is no reason to change CDI provider during runtime.
Since this is behavioral change, we should fix it for next major.

This is also related to https://github.com/jakartaee/cdi/pull/617 since both APIs should work the same way so we should change both.

manovotn:
…nvocations of set method

Fixes #618 

Both APIs now behave equally and guard firstly against `null` value and then against attempts to re-set previously configured value.

This is potentially breaking change and as such should only be part of next bigger release.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildServicesResolver.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildServicesResolver.java
@@ -12,7 +12,6 @@ package jakarta.enterprise.inject.build.compatible.spi;
 
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.Objects;
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


Ladicek (approved):
LGTM

starksm64:
So the question is do we want to plan for a 4.1 release, or go to 5.0. Both require a creation review plan, so it is really just a question of what the scope of planned changes might be.

manovotn:
> So the question is do we want to plan for a 4.1 release, or go to 5.0. Both require a creation review plan, so it is really just a question of what the scope of planned changes might be.

TBF, I don't know.
I'd say we only need a major bump if we were planning some big breaking change. The current list of issues consists mostly of smaller items.
That being said, it might be worth asking people during the meeting (or two) and/or through mailing list for their ideas on next version and capture it in dedicated tracking issue. From there we might get a hang of what the list of possible changes looks like and decide. WDYT @starksm64 @Ladicek?

Ladicek:
I agree we should first put together a plan for the next version, and its version number will naturally follow from the content of that plan.

manovotn:
I am going to merge this.
In case we need a service release, we can always make a dedicated branch and cut off required commits.

PR #620, opened 2022-05-26
Add an aggregate javadoc generation to combine api and lang-model jav…

starksm64:
…adoc for spec site

Signed-off-by: Scott M Stark <starksm64@gmail.com>


manovotn (approved):

Ladicek:
This is only meant for manual usage, right? Because I _think_ if you just do `mvn clean install`, the root module will be built first and hence won't produce anything, because the `api` and `lang-model` modules aren't built yet.

starksm64:
Yes, it is just used to build the aggregate javadoc that go into the apidocs found on the spec page:
https://jakarta.ee/specifications/cdi/4.0/apidocs/

I don't really like how this works, but it took hours to come up with something that was generating a descent view of the combined javadoc. I must not be understanding something, but for now there is a way to generate the aggregate javadoc.

manovotn:
@starksm64 do we still need this? I don't see any such functionality in the repo but I am not sure since this has been hanging here for a while. Feel free to merge if you need, the approval is in place.

PR #627, opened 2022-06-29
Clarify ambiguity resolution during dynamic lookup

Ladicek:
This attempts to clarify the current specification of `Instance.iterator()`,
`Instance.stream()` and `Instance.handles()` in presence of dependency
ambiguity. My understanding is that this modification is what has always
been intended and how current implementations already work.

### DIFF
--- a/spec/src/main/asciidoc/core/injectionandresolution.asciidoc
+++ b/spec/src/main/asciidoc/core/injectionandresolution.asciidoc
@@ -649,12 +649,18 @@ The `get()` method must:
 The `iterator()` method must:
 
 * Identify the set of beans that have the required type and required qualifiers and are eligible for injection into the class into which the parent `Instance` was injected, according to the rules of typesafe resolution, as defined in <<performing_typesafe_resolution>>, resolving ambiguities according to <<unsatisfied_and_ambig_dependencies>>.
-* Return an `Iterator`, that iterates over the set of contextual references for the resulting beans and required type, as defined in <<contextual_reference>>.
+* If typesafe resolution results in an unsatisfied dependency, the set of resulting beans is empty.
+  If typesafe resolution results in an ambiguous dependency and the set of candidate beans contains at least one alternative, the set of resulting beans contains all beans that were not eliminated during ambiguity resolution.
+  If typesafe resolution results in an ambiguous dependency and the set of candidate beans contains no alternative, the set of resulting beans contains all candidate beans.
+* Return an `Iterator`, that iterates over the set of contextual references for the resulting beans, as defined in <<contextual_reference>>.
 
 The `stream()` method must:
 
 * Identify the set of beans that have the required type and required qualifiers and are eligible for injection into the class into which the parent `Instance` was injected, according to the rules of typesafe resolution, as defined in <<performing_typesafe_resolution>>, resolving ambiguities according to <<unsatisfied_and_ambig_dependencies>>.
-* Return a `Stream`, that can stream over the set of contextual references for the resulting beans and required type, as defined in <<contextual_reference>>.
+* If typesafe resolution results in an unsatisfied dependency, the set of resulting beans is empty.
+  If typesafe resolution results in an ambiguous dependency and the set of candidate beans contains at least one alternative, the set of resulting beans contains all beans that were not eliminated during ambiguity resolution.
+  If typesafe resolution results in an ambiguous dependency and the set of candidate beans contains no alternative, the set of resulting beans contains all candidate beans.
+* Return a `Stream`, that can stream over the set of contextual references for the resulting beans, as defined in <<contextual_reference>>.
 
 The methods `isUnsatisfied()`, `isAmbiguous()` and `isResolvable()` must:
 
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -671,8 +673,8 @@ An `UnsupportedOperationException` is thrown if the active context object for th
 The `getHandle()` method must:
 
 * Return an initialized contextual reference `Handle<T>` for a bean that has the required type and qualifiers and is eligible for injection. The contextual reference must be resolved lazily, i.e. when first needed. <<handle_interface>> is described in a separate paragraph.
-* Throw `UnsatisfiedResolutionException` if there is no bean with given type and qualifiers
-* Throw `AmbiguousResolutionException` if there is more than one bean given type and qualifiers
+* Throw `UnsatisfiedResolutionException` if there is no bean with given type and qualifiers.
+* Throw `AmbiguousResolutionException` if there is more than one bean given type and qualifiers.
 
 The `handles()` method must:
```
Ladicek, line 679: We should probably add a similar change here?

Ladicek:
Also the ambiguity resolution process is described in a way that it _always_, unconditionally, eliminates all non-alternative beans -- which would technically mean that dynamic lookup of an ambiguous dependency for which no alternative exists should always end up empty. Which is clearly not what happens in existing implementations :-)

mkouba:
> Also the ambiguity resolution process is described in a way that it _always_, unconditionally, eliminates all non-alternative beans -- which would technically mean that dynamic lookup of an ambiguous dependency for which no alternative exists should always end up empty. Which is clearly not what happens in existing implementations :-)

I tend to disagree. The spec is clear that _"**When an ambiguous dependency exists**, the container attempts to resolve the ambiguity."_

manovotn:
> Also the ambiguity resolution process is described in a way that it always, unconditionally, eliminates all non-alternative beans -- which would technically mean that dynamic lookup of an ambiguous dependency for which no alternative exists should always end up empty.

I think you are reading too much into it :)
The container attempts to resolve it but if you were to observe the state after attempted ambiguity removal which failed, you wouldn't be able to tell the difference between ambig. and unsatis. dependency since the set might just be empty.
If the resolution fails, you simply get to view all candidate beans that are valid picks for given type/qualifier combination

Ladicek:
> > Also the ambiguity resolution process is described in a way that it _always_, unconditionally, eliminates all non-alternative beans -- which would technically mean that dynamic lookup of an ambiguous dependency for which no alternative exists should always end up empty. Which is clearly not what happens in existing implementations :-)
> 
> I tend to disagree. The spec is clear that _"**When an ambiguous dependency exists**, the container attempts to resolve the ambiguity."_

Exactly! And the first step of the ambiguity resolution process is elimination of non-alternative beans. If ambiguity exists but no eligible bean is an alternative, the ambiguity resolution process ends with no candidate.

I'm trying to describe `Instance.iterator()` in terms of the ambiguity resolution process, but its wording isn't exactly friendly :-)

> I think you are reading too much into it :) The container attempts to resolve it but if you were to observe the state after attempted ambiguity removal which failed, you wouldn't be able to tell the difference between ambig. and unsatis. dependency since the set might just be empty.

Right. As a spec lawyer wannabe, I still find that deeply disturbing :laughing:

> If the resolution fails, you simply get to view all candidate beans that are valid picks for given type/qualifier combination

No, you get all the eligible beans after the elimination process, that's what we investigated yesterday. (Unless all beans are eliminated :-) )

mkouba:
> If ambiguity exists but no eligible bean is an alternative, the ambiguity resolution process ends with no candidate.

Not really, the ambiguity resolution process ends with an _unresolvable dependency_ ;-). But yes, the behavior is undefined.

manovotn:
> No, you get all the eligible beans after the elimination process, that's what we investigated yesterday. (Unless all beans are eliminated :-) )

Yes, that's what I tried to describe but worded it imprecisely.

> Right. As a spec lawyer wannabe, I still find that deeply disturbing

Uff, you'd have to then make the spec even harder to read and expand on what's eligible bean while talking resolution process and then go in depth on describing the various outcomes you can have when interacting with `Instance`. Which might just bring more confusion than clarity in the end :)

Ladicek:
> > No, you get all the eligible beans after the elimination process, that's what we investigated yesterday. (Unless all beans are eliminated :-) )
> 
> Yes, that's what I tried to describe but worded it imprecisely.

This is important though ;-) For example, the specification of `Instance.handles()` currently says that it must "allow iterating over contextual reference handles for all beans that have the required type and required qualifiers and are eligible for injection." Though I believe `handles()` should really hand out handles to the same set of beans as `iterator()` or `stream()`, right?

> Uff, you'd have to then make the spec even harder to read and expand on what's eligible bean while talking resolution process and then go in depth on describing the various outcomes you can have when interacting with `Instance`. Which might just bring more confusion than clarity in the end :)

I actually think I can fix that by adding one more sentence to the description of `Instance.iterator()`, I'll update this PR in a bit.

manovotn (approved):
Had some off-list discussion with Ladislav and I think this state of PR is fine.

Just for the sake of clarity, this doesn't change any of current behaviors, it is purely a clarification of existing state.

Ladicek:
Thanks! I'll check the TCK and possibly submit a PR there, then.

PR #629, opened 2022-07-18
Introduce BeanContainer.getContexts() to be able to get hold of all context objects, even if inactive

### LINKED ISSUE #628: Accessing custom context objects in CDI Lite
If a build compatible extension adds a custom scope via `MetaAnnotations#addScope`, there will be exactly one context object of the given context class in the container if I read the JavaDoc correctly. However, there seems to be no way to extract this unique instance from the container once its running, except via `BeanContainer#getContext` and that only works if the context object I want is immediately active after instantiation and also happens to be the unique active context object for its scope.

This makes it surprisingly hard to define a new scope and context that can be programmatically activated and deactivated, like the RequestScope can.

In CDI full, this would be easy with a portable extension, because then I actually add the context objects themselves, not just a context class, and I could just store them in some field in the Extension and extract them later by injecting the Extension somewhere. In CDI Lite however, I would have to do a non-zero amount of shenanigans, e.g. 

- I could make the constructor store a reference to the new instance in some static variable somewhere and hope nobody except the container will ever ever ever call that constructor. This introduces fragile global state and makes it impossible to have multiple containers running in parallel (for unit testing say).
- Or I could have the context active immediately after instantiation, call `getContext(MyScope.class)` and store the object in some `@Singleton` bean by listening to a start up event and then immediately deactivate the context until it is truly needed. It is probably semantically wrong to have to scope start out as active if the scope should be programmatically activated and deactivated. Also, this is simply impossible if I ever need to context classes for the same scope. Deactivating the context may accidentally trigger `@MyScope` scoped beans that listen to a `@BeforeDestroyed` event.

It's probably possible to find better workarounds with even more shenanigans, but that seems unnecessarily hard. Has this simply been overlooked or is it by design not (easily) possible in CDI Lite?

manovotn:
Fixes #628 

I will also create tracking TCK issue.
Also, I've checked in Weld that this is indeed trivial to implement as all `BeanManager` instances hold references to all contexts anyway.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
@@ -13,6 +13,7 @@ import jakarta.enterprise.inject.Stereotype;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import java.util.Collection;
 import java.util.List;
 import java.util.Set;
 
@@ -211,6 +212,15 @@ public interface BeanContainer {
      * @throws IllegalArgumentException  if there is more than one active context object for the given scope
      */
     Context getContext(Class<? extends Annotation> scopeType);
+
+    /**
+     * Obtains all {@linkplain Context context objects}, active and inactive, for the given
+     * {@linkplain jakarta.enterprise.context scope}.
+     *
+     * @param scopeType the {@linkplain jakarta.enterprise.context scope}; must not be {@code null}
+     * @return immutable collection of {@linkplain Context context objects}; never {@code null}, but may be empty
+     */
+    Collection<Context> getContexts(Class<? extends Annotation> scopeType);
 
     /**
      * Returns an instance of Event with specified type <code>java.lang.Object</code> and specified qualifier <code>@Default</code>


Ladicek (approved):

api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
```diff
@@ -212,6 +213,15 @@ public interface BeanContainer {
      */
     Context getContext(Class<? extends Annotation> scopeType);
 
+    /**
+     * Obtains all {@linkplain Context context objects}, active and inactive, for the given
+     * {@linkplain jakarta.enterprise.context scope}.
+     *
+     * @param scopeType the {@linkplain jakarta.enterprise.context scope}; must not be {@code null}
+     * @return immutable collection of {@linkplain Context context objects}; never {@code null}, but may be empty
```
Emily-Jiang, line 221:
```suggestion
     * @return an immutable collection of {@linkplain Context context objects}; never {@code null}, but may be empty
```

Ladicek:
Need to file a TCK tracking issue now.

PR #639, opened 2022-12-02
Method invokers

Ladicek:
This PR has 2 commits with detailed commit messages. I'm copying the commit messages here for easier reference.

This PR doesn't include changes to the specification text, but the API javadocs are intentionally a specification-level material. There are several TODOs and everything is up for debate anyway, but I still wanted to provide something that is pretty well defined.

## Initial proposal for method invokers: introduce core concepts

_`Invoker`_ is an indirect way of invoking a bean method. It is a simple
functional interface:

```java
interface Invoker<T, R> {
    R invoke(T instance, Object[] arguments);
}
```

Invokers are built using an _`InvokerBuilder`_. They must be thread-safe,
so frameworks can build an invoker for each relevant method once and then
reuse it to invoke that method whenever necessary.

An invoker may be built for a method that:

1. is not `private`;
2. is not a constructor;
3. is declared on a bean class of a managed bean (class-based bean),
   with the exception of interceptors and decorators, or inherited
   from its supertypes.

If an invocation of a method is intercepted, an equivalent invocation
through an invoker is also intercepted.

Portable extensions may obtain an `InvokerBuilder` in an observer
of the `ProcessManagedBean` type. The `InvokerBuilder` in this case
produces `Invoker` objects directly.

Build compatible extensions may obtain an `InvokerBuilder` during
`@Registration` from the `BeanInfo` object. The `InvokerBuilder` in
this case produces `InvokerInfo` objects. `InvokerInfo` is an opaque
token that can only be used to materialize an `Invoker` in synthetic
components.

## Initial proposal for method invokers: invoker builder features

This commit introduces actual interesting features to the `InvokerBuilder`:

- automatic lookup of the target instance and arguments
- transformation of the target instance and arguments before invocation
- transformation of the return value and thrown exception after invocation
- wrapping the invoker into a custom piece of code for maximum flexibility

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticObserverBuilder.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticObserverBuilder.java
@@ -335,6 +335,34 @@ public interface SyntheticObserverBuilder<T> {
     SyntheticObserverBuilder<T> withParam(String key, Annotation[] value);
 
     /**
+     * Adds an invoker-valued parameter to the parameter map. The parameter map is passed
+     * to the {@linkplain SyntheticObserver event notification} function when the event is fired.
+     * <p>
+     * When looked up from the parameter map in the event notification function, the value will be
+     * an instance of {@link jakarta.enterprise.invoke.Invoker Invoker}, <em>not</em> an {@code InvokerInfo}.
+     *
+     * @param key the parameter key, must not be {@code null}
+     * @param value the parameter value
+     * @return this {@code SyntheticBeanBuilder}
+     * @since 4.1
+     */
+    SyntheticObserverBuilder<T> withParam(String key, InvokerInfo value);
+
+    /**
+     * Adds an invoker array-valued parameter to the parameter map. The parameter map is passed
+     * to the {@linkplain SyntheticObserver event notification} function when the event is fired.
+     * <p>
+     * When looked up from the parameter map in the event notification function, the values will be
+     * instances of {@link jakarta.enterprise.invoke.Invoker Invoker}, <em>not</em> {@code InvokerInfo}.
+     *
+     * @param key the parameter key, must not be {@code null}
+     * @param value the parameter value
+     * @return this {@code SyntheticBeanBuilder}
+     * @since 4.1
+     */
+    SyntheticObserverBuilder<T> withParam(String key, InvokerInfo[] value);
+
+    /**
      * Sets the class of the synthetic observer {@linkplain SyntheticObserver event notification} function.
      * CDI container will create an instance of the event notification function every time when it needs
      * to notify the synthetic observer. The class must be {@code public} and have a {@code public}
--- a/api/src/main/java/jakarta/enterprise/inject/spi/ProcessManagedBean.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/ProcessManagedBean.java
@@ -15,6 +15,9 @@
  */
 package jakarta.enterprise.inject.spi;
 
+import jakarta.enterprise.invoke.Invoker;
+import jakarta.enterprise.invoke.InvokerBuilder;
+
 /**
  * <p>
  * The container fires an event of this type for each enabled managed bean, before registering the
@@ -38,4 +41,17 @@ public interface ProcessManagedBean<X> extends ProcessBean<X> {
      * @throws IllegalStateException if called outside of the observer method invocation
      */
     public AnnotatedType<X> getAnnotatedBeanClass();
+
+    /**
+     * Returns a new {@link InvokerBuilder} for given method. The builder eventually produces an invoker
+     * for the given method.
+     * <p>
+     * The {@code method} must be declared on the bean class or inherited from a supertype
+     * of the bean class of the bean being registered, otherwise an exception is thrown.
+     *
+     * @param method method of the bean being registered, must not be {@code null}
+     * @return the invoker builder, never {@code null}
+     * @since 4.1
+     */
+    public InvokerBuilder<Invoker<X, ?>> createInvoker(AnnotatedMethod<? super X> method);
 }
--- a/api/src/main/java/jakarta/enterprise/invoke/Invoker.java
+++ b/api/src/main/java/jakarta/enterprise/invoke/Invoker.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) 2022 Red Hat and others
+ *
+ * This program and the accompanying materials are made available under the
+ * Apache Software License 2.0 which is available at:
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package jakarta.enterprise.invoke;
+
+/**
+ * Allows indirectly invoking a method that belongs to a managed bean (the <em>target method</em>).
+ * To invoke the method, the caller must provide all the arguments that the target method accepts,
+ * as well as the instance on which the target method is to be invoked, if it is not {@code static}.
+ * <p>
+ * Whenever a direct invocation of a method is a business method invocation, an indirect invocation
+ * of that method through an invoker is also a business method invocation.
+ * <p>
+ * Invoker implementations must be thread-safe. It is possible to use a single invoker instance
+ * to perform multiple independent invocations of the target method, possibly on different instances
+ * and with different arguments.
+ *
+ * <h2>Obtaining an invoker</h2>
+ *
+ * The CDI container allows {@linkplain InvokerBuilder building} an invoker for non-private
+ * methods declared on a managed bean class or inherited from a supertype. Attempting to build
+ * an invoker for a private method or a constructor of a managed bean class leads to a deployment
+ * problem. Attempting to build an invoker for a method of a class that is not a managed bean class
+ * or that is an interceptor or decorator class leads to a deployment problem.
+ * <p>
+ * Multiple managed beans may inherit a method from a common supertype. In that case, each bean
+ * conceptually has its own method and an invoker obtained for one bean may not be used to invoke
+ * the method on the other bean.
+ * <p>
+ * Using the {@link InvokerBuilder} is the only way to obtain an invoker. An {@code InvokerBuilder}
+ * can only be obtained in CDI portable extensions and build compatible extensions.
+ *
+ * <h2>Example</h2>
+ *
+ * To illustrate how invokers work, let's take a look at an example. Say that the following bean
+ * exists and has a method that you want to invoke indirectly:
+ *
+ * <pre>
+ * &#64;Dependent
+ * class MyService {
+ *     String hello(String name) {
+ *         return "Hello " + name + "!";
+ *     }
+ * }
+ * </pre>
+ *
+ * When you obtain an {@code InvokerBuilder} for the {@code hello()} method, you can
+ * immediately build a direct invoker. In a portable extension, this results in an invoker:
+ *
+ * <pre>
+ * InvokerBuilder&lt;Invoker&lt;MyService, String&gt;&gt; builder = ...;
+ * Invoker&lt;MyService, String&gt; invoker = builder.build();
+ * </pre>
+ *
+ * In a build compatible extension, this results in an opaque token that later
+ * materializes as an invoker:
+ *
+ * <pre>
+ * InvokerBuilder&lt;InvokerInfo&gt; builder = ...;
+ * InvokerInfo invoker = builder.build();
+ * </pre>
+ *
+ * To call the {@code hello()} method through this invoker, call
+ * {@code invoker.invoke(myService, new Object[] {"world"})}.
+ * The return value is {@code "Hello world!"}.
+ * <p>
+ * An implementation of the direct invoker above is equivalent to the following class:
+ *
+ * <pre>
+ * class TheInvoker implements Invoker&lt;MyService, String&gt; {
+ *     String invoke(MyService instance, Object[] arguments) {
+ *         return instance.hello((String) arguments[0]);
+ *     }
+ * }
+ * </pre>
+ *
+ * @param <T> type of the target instance
+ * @param <R> return type of the method
+ * @since 4.1
+ * @see #invoke(Object, Object[])
+ */
+public interface Invoker<T, R> {
+    /**
+     * Invokes the target method of this invoker on given {@code instance}, passing given
+     * {@code arguments}. If the target method is {@code static}, the {@code instance} is ignored;
+     * by convention, it should be {@code null}. If the target method returns normally, this
+     * method returns its return value, unless the target method is declared {@code void},
+     * in which case this method returns {@code null}. If the target method throws an exception,
+     * this method rethrows it directly.
+     * <p>
+     * If some parameter of the target method declares a primitive type, the corresponding element of
+     * the {@code arguments} array must be of the corresponding wrapper type. No type conversions are
+     * performed, so if the parameter is declared {@code int}, the argument must be an {@code Integer}
+     * and may not be {@code Short} or {@code Long}. If the argument is {@code null}, the default value
+     * of the primitive type is used. Note that this does not apply to arrays of primitive types;
+     * if a parameter is declared {@code int[]}, the argument must be {@code int[]} and may not be
+     * {@code Integer[]}.
+     * <p>
+     * If the target method is not {@code static} and {@code instance} is {@code null},
+     * a {@link NullPointerException} is thrown. If the target method is not {@code static} and
+     * the {@code instance} is not assignable to the class of the bean to which the method belongs,
+     * a {@link ClassCastException} is thrown.
+     * <p>
+     * If the target method declares no parameter, {@code arguments} are ignored. If the target method
+     * declares any parameter and {@code arguments} is {@code null}, {@link NullPointerException} is
+     * thrown. If the {@code arguments} array has fewer elements than the number of parameters of
+     * the target method, {@link ArrayIndexOutOfBoundsException} is thrown. If the {@code arguments}
+     * array has more elements than the number of parameters of the target method, the excess elements
+     * are ignored. If some of the {@code arguments} is not assignable to the declared type of
+     * the corresponding parameter of the target method, {@link ClassCastException} is thrown.
+     *
+     * TODO the previous 2 paragraphs refer to "assignability", which needs to be defined somewhere!
+     *
+     * @param instance the instance on which the target method is to be invoked, may only be {@code null}
+     *                 if the method is {@code static}
+     * @param arguments arguments to be supplied to the target method, may only be {@code null}
+     *                  if the method declares no parameter
+     * @return return value of the target method, or {@code null} if the method is declared {@code void}
+     */
+    R invoke(T instance, Object[] arguments); // TODO throws Exception ?
+}
--- a/api/src/main/java/jakarta/enterprise/invoke/Invoker.java
+++ b/api/src/main/java/jakarta/enterprise/invoke/Invoker.java
@@ -118,6 +118,10 @@ public interface Invoker<T, R> {
      *
      * TODO the previous 2 paragraphs refer to "assignability", which needs to be defined somewhere!
      *
+     * TODO when the `InvokerBuilder` applies transformations, some of the requirements above
+     *  are no longer strictly necessary, should reflect that in this text somehow (it is already
+     *  mentioned in `InvokerBuilder`, but that likely isn't enough)
+     *
      * @param instance the instance on which the target method is to be invoked, may only be {@code null}
      *                 if the method is {@code static}
      * @param arguments arguments to be supplied to the target method, may only be {@code null}
--- a/api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
+++ b/api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
@@ -11,16 +11,357 @@
 package jakarta.enterprise.invoke;
 
 /**
- * Builder of {@link Invoker}s.
+ * Builder of {@link Invoker}s that allows configuring input lookups, input and output
+ * transformations, and invoker wrapping. The method for which the invoker is built is
+ * called the <em>target method</em>. If a lookup is configured, the corresponding input
+ * of the invoker is ignored and an instance is looked up from the CDI container before
+ * the target method is invoked. If a transformation is configured, the corresponding input
+ * or output of the invoker is modified in certain way before or after the target method
+ * is invoked. If a wrapper is configured, the invoker is passed to custom code for getting
+ * invoked. As a result, the built {@code Invoker} instance may have more complex behavior
+ * than just directly calling the target method.
+ * <p>
+ * Transformations and wrapping are expressed by ordinary methods that must have
+ * a pre-defined signature, as described below. Such methods are called
+ * <em>transformers</em> and <em>wrappers</em>.
+ * <p>
+ * Invokers may only be built during deployment. It is not possible to build new invokers
+ * at application runtime.
+ *
+ * <h2>Example</h2>
+ *
+ * Before describing in detail how lookups, transformers and wrappers work, let's take
+ * a look at an example. Say we have the following bean with a method:
+ *
+ * <pre>
+ * class MyService {
+ *     String hello(String name) {
+ *         return "Hello " + name + "!";
+ *     }
+ * }
+ * </pre>
+ *
+ * And we want to build an invoker that looks up {@code MyService} from the CDI container,
+ * always passes the argument to {@code hello()} as all upper-case, and repeats the return
+ * value twice. To transform the argument, we can use the zero-parameter method
+ * {@code String.toUpperCase()}, and to transform the return value, we write a transformer
+ * as a simple {@code static} method:
+ *
+ * <pre>
+ * class Transformations {
+ *     static String repeatTwice(String str) {
+ *         return str + " " + str;
+ *     }
+ * }
+ * </pre>
+ *
+ * Then, assuming we have obtained the {@code InvokerBuilder} for {@code MyService.hello()},
+ * we can set up the lookup and transformations and build an invoker like so:
+ *
+ * <pre>
+ * builder.setInstanceLookup()
+ *        .setArgumentTransformer(0, String.class, "toUpperCase")
+ *        .setReturnValueTransformer(Transformations.class, "repeatTwice")
+ *        .build();
+ * </pre>
+ *
+ * The resulting invoker will be equivalent to the following class:
+ * <pre>
+ * class TheInvoker implements Invoker&lt;MyService, String&gt; {
+ *     String invoke(MyService ignored, Object[] arguments) {
+ *         MyService instance = CDI.current().select(MyService.class).get();
+ *         String argument = (String) arguments[0];
+ *         String transformedArgument = argument.toUpperCase();
+ *         String result = instance.hello(transformedArgument);
+ *         String transformedResult = Transformations.repeatTwice(result);
+ *         return transformedResult;
+ *     }
+ * }
+ * </pre>
+ *
+ * The caller of this invoker may pass {@code null} as the target instance, because
+ * the invoker will lookup the target instance on its own. Therefore, calling
+ * {@code invoker.invoke(null, new Object[] {"world"})} will return
+ * {@code "Hello WORLD! Hello WORLD!"}.
+ *
+ * <h2>General requirements</h2>
+ *
+ * To refer to a transformer or a wrapper, all methods in this builder accept:
+ * 1. the {@code Class} that that declares the method, and 2. the {@code String} name
+ * of the method.
+ * <p>
+ * Transformers may be {@code static}, in which case they must be declared directly
+ * on the given class, or they may be instance methods, in which case they may be declared
+ * on the given class or inherited from any of its supertypes.
+ * <p>
+ * It is possible to register only one transformer of each kind, or for each argument
+ * position in case of argument transformers. Attempting to register a second transformer
+ * of the same kind, or for the same argument position, leads to an exception.
+ * <p>
+ * Wrappers must be {@code static} and must be declared directly on the given class.
+ * It is possible to register only one wrapper. Attempting to register a second wrapper
+ * leads to an exception.
+ * <p>
+ * It is a deployment problem if no method with given name and valid signature is found,
+ * or if multiple methods with given name and different valid signatures are found. It is
+ * a deployment problem if a registered transformer or wrapper is not {@code public}.
+ * <p>
+ * Transformers and wrappers may declare the {@code throws} clause. The declared exception
+ * types are ignored when searching for the method.
+ * <p>
+ * For the purpose of the specification of transformers and wrappers below, the term
+ * <em>any-type</em> is recursively defined as: the {@code java.lang.Object} class type,
+ * or a type variable that has no bound, or a type variable whose first bound is
+ * <em>any-type</em>.
+ *
+ * <h2>Input lookups</h2>
+ *
+ * For the target instance and for each argument, it is possible to specify that the value
+ * passed to {@code Invoker.invoke()} should be ignored and a value should be looked up
+ * from the CDI container instead.
+ * <p>
+ * For the target instance, a CDI lookup is performed with the required type equal to the bean
+ * class of the bean to which the target method belongs, and required qualifiers equal to the set
+ * of all qualifier annotations present on the bean class of the bean to which the target method
+ * belongs. When the target method is {@code static}, the target instance lookup is skipped.
+ * <p>
+ * For an argument, a CDI lookup is performed with the required type equal to the type of
+ * the corresponding parameter of the target method, and required qualifiers equal to the set
+ * of all qualifier annotations present on the corresponding parameter of the target method.
+ * <p>
+ * Implementations are required to resolve all lookups during deployment. It is a deployment
+ * problem if the lookup ends up unresolved or ambiguous.
+ * <p>
+ * If the looked up bean is {@code @Dependent}, it is guaranteed that the instance will be
+ * destroyed after the target method is invoked but before the the invoker returns. The order
+ * in which the looked up {@code @Dependent} beans are destroyed is not specified.
+ * <p>
+ * The order in which input lookups are performed in not specified and must not be relied upon.
+ *
+ * <h2>Input transformations</h2>
+ *
+ * The target method has 2 kinds of inputs: the target instance (unless the target method is
+ * {@code static}, in which case the target instance is ignored and should be {@code null}
+ * by convention) and arguments. These inputs correspond to the parameters of
+ * {@link Invoker#invoke(Object, Object[]) Invoker.invoke()}.
+ * <p>
+ * Each input can be transformed by a transformer that has one of the following signatures,
+ * where {@code X} and {@code Y} are types:
+ *
+ * <ul>
+ * <li>{@code static X transform(Y value)}</li>
+ * <li>{@code static X transform(Y value, Consumer<Runnable> cleanup)}</li>
+ * <li>{@code X transform()} &ndash; in this case, {@code Y} is the type of the class that
+ * declares the transformer</li>
+ * </ul>
+ *
+ * An input transformer must produce a type that can be consumed by the target method.
+ * Specifically: when {@code X} is <em>any-type</em>, it is not type checked during deployment.
+ * Otherwise, it is a deployment problem if {@code X} is not assignable to the corresponding type
+ * in the declaration of the target method (that is the bean class in case of target instance
+ * transformers, or the corresponding parameter type in case of argument transformers). {@code Y}
+ * is not type checked during deployment, so that input transformers may consume arbitrary types.
+ * TODO this paragraph refers to "assignability", which needs to be defined somewhere!
+ * <p>
+ * When a transformer is registered for given input, it is called before the target method is
+ * invoked, and the outcome of the transformer is used in the invocation instead of the original
+ * value passed to the invoker by its caller.
+ * <p>
+ * If the transformer declares the {@code Consumer<Runnable>} parameter, and the execution
+ * of the transformer calls {@code Consumer.accept()} with some {@code Runnable}, it is
+ * guaranteed that the {@code Runnable} will be called after the target method is invoked but
+ * before the invoker returns. These {@code Runnable}s are called <em>cleanup tasks</em>.
+ * The order of cleanup task execution is not specified. Passing a {@code null} cleanup task
+ * to the {@code Consumer} is permitted, but has no effect.
+ * <p>
+ * If an input transformation is configured for an input for which a lookup is also configured,
+ * the lookup is performed first and the transformation is applied to the looked up value.
+ * If the looked up bean for some input is {@code @Dependent}, it is guaranteed that all
+ * cleanup tasks registered by a transformer for that input are called before that looked up
+ * {@code @Dependent} bean is destroyed.
+ * <p>
+ * The order in which input transformations are performed in not specified and must not
+ * be relied upon.
+ *
+ * <h2>Output transformations</h2>
+ *
+ * The target method has 2 kinds of outputs: the return value and the thrown exception. These
+ * outputs correspond to the return value of {@link Invoker#invoke(Object, Object[]) Invoker.invoke()}
+ * or its thrown exception, respectively.
+ * <p>
+ * Each output can be transformed by a transformer that has one of the following signatures,
+ * where {@code X} and {@code Y} are types:
+ *
+ * <ul>
+ * <li>{@code static X transform(Y value)}</li>
+ * <li>{@code X transform()} &ndash; in this case, {@code Y} is the type of the class that
+ * declares the transformer</li>
+ * </ul>
+ *
+ * An output transformer must consume a type that can be produced by the target method.
+ * Specifically: when {@code Y} is <em>any-type</em>, it is not type checked during deployment.
+ * Otherwise, it is a deployment problem if {@code Y} is not assignable from the return type of
+ * the target method in case of return value transformers, or from {@code java.lang.Throwable}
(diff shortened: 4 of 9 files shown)


tomas-langer:
Is there a way to find instances of parameters using CDI? E.g. if we consider that each parameter is an injection point, could we delegete the whole invocation to CDI?

Something as:
```java
invokable.invoke(myBeanInstance); // all arguments are bound from current CDI context
```

or
```java
invokable.invocationBuilder()
  .injectParam(0)  // injection point
  .param(1, myEntity) // explicit argument value
  .injectParam(2) // injection point
```

In such a case, we could discover injection points in extensions, and correctly prepare producers for them (such as for `@QueryParam` etc.) rather than analyzing each method and creating all parameters by hand.

Ladicek:
The `InvokerBuilder` allows this using `setInstanceLookup` and `setArgumentLookup(int)`.

tomas-langer:
> The `InvokerBuilder` allows this using `setInstanceLookup()` and `setArgumentLookup(int)`.

Thanks a lot!

Ladicek:
BTW, if anyone is interested, we can discuss this proposal on the CDI call next week.

ljnelson:
The name seems problematic (a method can be invoked due to its very nature) and possibly [misspelled](https://www.merriam-webster.com/dictionary/invocable).  Perhaps instead: if a _managed bean_ is a bean you don't create but the container creates for you, and is a bean for which you don't supply dependencies, but for which instead the container supplies dependencies, then so, perhaps, a _managed method_ is a method you don't invoke but the container invokes for you, and is a method for which you don't supply arguments, but for which instead the container supplies arguments? Just a thought.

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
```diff
@@ -0,0 +1,359 @@
...
+
+    /**
+     * Configures an output transformer for the return value.
+     *
+     * @param clazz class that declares the transformer
+     * @param methodName transformer method name
+     * @return this builder
+     * @throws IllegalStateException if this method is called more than once
+     */
+    InvokerBuilder<T> setReturnValueTransformer(Class<?> clazz, String methodName);
```
graemerocher, line 331: These seem too stringly typed and dynamic, can we not use method handles and provide generics for the return type? Why add a requirement for reflection here?

api/src/main/java/jakarta/enterprise/invoke/Invokable.java
```diff
@@ -0,0 +1,49 @@
...
+ * and for example an invoker obtained for one bean cannot be used to invoke
+ * the method on the other bean.
+ * <p>
+ * This annotation may be used as a meta-annotation. Classes and methods annotated with
+ * an annotation that is directly or transitively meta-annotated {@code @Invokable}
+ * are treated as if they were directly annotated {@code @Invokable}.
+ *
+ * @since 5.0
+ */
+@Target({ElementType.METHOD, ElementType.TYPE})
```
graemerocher, line 44: needs to be a meta annotation

graemerocher (commented):
looks like good progress, just a few suggestions

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
Ladicek: It is rather stringly typed, but all alternatives I could think of are as much stringly typed if not more. The idea here is to type check what we can during deployment (which would be during application build for build-time implementations) and let the JVM to sort out the rest. To allow transformers to convert values from one type to another, which seems like a hard requirement to me, we can only type check the output type for input transformers (as the output of an input transformer is passed to the invokable method and hence its type must match) and the input type for output transformers (because the input of an output transformer is obtained from the invokable method and hence its type must match).

Or did you mean the syntax of referring to the transformer method, the tuple `(Class, String)`? I considered 2 alternatives:

1. Require transformers to implement some interface. Not nice, because it requires extra allocations and virtual dispatch. I wanted transformers to be `static` methods mainly to eliminate overhead, but it's also nice in that it tells users that transformers should be stateless. (They can also be instance methods, but that is [or should be] by construction restricted to simple scenarios where the `static` method would just call that instance method.)
2. Use method references (as in, `SomeClass::transformer`). This would require defining functional interfaces for all possible signatures of transformers and provide overloads that accept these functional interfaces. It would also require some way to find the method during deployment (that is, during application build for build-time implementations). Apparently there's a trick to do that if the functional interface extends `Serializable`, which seems strange to me. Finally, we might also want to detect (and forbid?) situation where the caller doesn't pass us a method reference, but an actual lambda.

I'd really love to use method references, but I couldn't figure out how to implement it in a reasonable way. So I settled with just a pair `(Class, String)`, which is basically what I'd need from the method reference, except explicit (and unfortunately not checked during compilation).

Finally, the idea is that all invokers are built during deployment -- there's no way to obtain an `InvokerBuilder` during application runtime. Therefore, there's no requirement for reflection here -- build-time implementations are supposed to find the transformer method using whatever language model they use (e.g. Jandex in case of Quarkus) and build the invoker using whatever mechanism they prefer (e.g. bytecode generation in case of Quarkus).

api/src/main/java/jakarta/enterprise/invoke/Invokable.java
Ladicek: `TYPE` implies `ANNOTATION_TYPE`, at least per `ElementType` javadoc, so this should be enough. But can do that explicitly, yea.

Ladicek:
> The name seems problematic (a method can be invoked due to its very nature)

I used the same argument before, when we were discussing the idea of "executable methods". At the same time, it seems reasonable to refer to the core property of this new feature in its name. A somewhat more correct/descriptive name would be "indirectly-invokable methods", but that's a really ugly name. I went with "invokable" because the core types use this word (`Invokable`, `Invoker`) and the JDK already contains colliding classes for the verbs "call" and "execute" (`Callable`, `Executable`, `Executor`).

> and possibly [misspelled](https://www.merriam-webster.com/dictionary/invocable).

Interesting. I'm not a native speaker and "invocable" looks terribly alien to me :-) Google knows both terms and "invocable" seems more often used (cca 60K results for invokable and 200K results for invocable). There are some other APIs that use "invocable", too:

- https://en.cppreference.com/w/cpp/concepts/invocable
- https://docs.oracle.com/javase/7/docs/api/javax/script/Invocable.html
- https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_classes_annotation_InvocableMethod.htm

I can be persuaded to rename, though I also think public disobedience is in order here. If something can be _invoked_, it should be _invokable_! :-)

> Perhaps instead: if a _managed bean_ is a bean you don't create but the container creates for you, and is a bean for which you don't supply dependencies, but for which instead the container supplies dependencies, then so, perhaps, a _managed method_ is a method you don't invoke but the container invokes for you, and is a method for which you don't supply arguments, but for which instead the container supplies arguments? Just a thought.

Managed bean specifically in CDI is what I call a class-based bean. There are other kinds of beans, producer methods and producer fields, that are created by the container and for which dependencies are provided by the container (though the ultimate constructor call lies in user code). Also, for invokable methods, the caller must supply the target instance and the arguments (unless they build a special kind of invoker with `InvokerBuilder` that looks up the target instance and all arguments from CDI). So I don't think the analogy holds.

I'm certainly not married to "invokable methods", but "managed methods" doesn't cut it in my opinion.

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
graemerocher: ok I am also wondering though why all of this transformation and decoration is necessary? Surely this change to the spec should be purely focused on executable methods and the transformation / decoration side should be in the interceptors spec?

Ladicek:
> Can you say what the property is that distinguishes these methods from other methods?

You can obtain an `Invoker` for them and indirectly invoke them through the invoker. So they are "indirectly-invokable through a CDI-provided mechanism", or "invokable" for short. The very first paragraph of `@Invokable` javadoc says that: https://github.com/jakartaee/cdi/pull/639/files#diff-caa2a3ec23acc164cb3dee736b270531344e7daf84def4acbc8fb210fa7c18ce

> It seems to me the property is that they are _managed_, like other _managed_ things in Jakarta EE, in CDI, JPA, Servlet and elsewhere, where _managed_ means, loosely, "the container does most or all of it for you". Isn't that the property you're going after? If not that, then what?

So if there's anything "managed", it's the invoker. I'd be fine with calling the feature "managed invokers", but that doesn't help with figuring out a name for an annotation to mark methods as "subject to creating managed invokers".

mkouba:
I think that "managed" is (1) overused in the Jakarta ecosystem and (2) does not imply what this API is used for. `@Invokable` is IMO ok and if the javadoc is clear then +1.

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
manovotn: > the transformation / decoration side should be in the interceptors spec?

That was also my initial question when I saw Ladislav's prototype.
However, given the use cases above, I tend to think this is better approach. The idea is that invokeable methods are a feature used by extensions (or integrators, if you will) and as such the integrator might want to do their own specific tweak in how the method is invoked and which parameters are looked up and so on. This would be hard to achieve via interception as multiple extensions trying to handle it differently might easily clash.
Furthermore, unlike interceptors, in this case there is always at most one such transformation per registered invoker (or per extension/integrator that wants to tweak it specifically) - it also doesn't need explicit enablement and doesn't deal with priority meaning it isn't subject to interference from ordering. Last but not least, you can still apply other interceptors onto the method independently of what each extension decides to perform with the invocation.

manovotn:
> and possibly [misspelled](https://www.merriam-webster.com/dictionary/invocable).

I am not a native speaker but "invokable" versus "invocable" both seem valid at least according to dictionaries and some Google-fu. I'd therefore say that it is probably a matter of localized preferences depending on where you live :)

>  managed method is a method you don't invoke but the container invokes for you, and is a method for which you don't supply arguments, but for which instead the container supplies arguments? Just a thought.

I am definitely not a fan of "managed methods", so -1.
The word `managed` is overused already and not too clear either. Even in your own definition, you say that the container supplies the arguments, but that's actually optional. The container does perform the invocation, but the extension/integrator is who controls args, target instance, return value transformation and so on. So in a way, the extension manages these a lot more than the container does (as opposed to the "managed beans" case).

I don't really see an issue with current naming... but then again, naming is hard :)
Either way, I think we should first see if we can agree on the use cases for these methods and see if the API meets expectations and based on that it should be easier to derive a name capturing the intent more closely.

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
Azquelt: > Frameworks often want to let users use multiple signatures of the methods the framework will invoke, but also want to handle them uniformly internally.

Would it be useful then to allow the user to supply a functional interface which defines the uniform shape they're trying to create, and allow the `InvokerBuilder` to return an instance of that interface instead of an `Invoker`? 

Would that allow for better type checking?

> the syntax of referring to the transformer method, the tuple (Class, String)

I must stress that as a Java developer using regular Java without any ahead of time compilation, this seems absolutely horrible. Any transformer seems like it should be a `Function<I, O>` and accept a lambda.

I assume that this is a limitation of doing stuff at build time and I'd need to know more about what limitations that imposes to be able to suggest something better, but I'd hope that we can create something which doesn't seem fully alien to a regular Java developer.

Lastly, on a separate note, I would want the Javadoc for the methods which set a transformer to include the allowed signatures of the transformer method. It's currently listed in the class javadoc, but that information should be referenced or repeated in the method doc.

Azquelt:
I don't like "invokable methods" as a name. It's confusing because all methods are invokable by definition.

However, I think what we're doing here is allowing the user to create "method invokers", which can invoke methods in a special way and can eliminate a lot of complex boilerplate code for frameworks.

I would be happy to talk about "method invokers", although in general conversation I'd probably always refer to them as "CDI method invokers" since the concept is fairly generic, but I'd prefer that the spec and Javadoc never use the phrase "invokable method" and instead talk about "creating an invoker for a method", or a method being "the target for an invoker".

I think the parts of the API that I think would need to change are:
- `@Invokable` annotation, which could perhaps be `@InvokerTarget`
- `InvokableMethodInfo` which could be `InvokerTargetInfo` or `InvokerTargetMethodInfo`
- `InvokableMethod` which could be `InvokerTarget` or `InvokerTargetMethod`

As a side-note, "invoke" is a weird word, I would use "invocation" as the noun so it's not clear whether another derived word should use a 'c' or a 'k'. However, "invoker" and "invokable" seem more natural to me.

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
Ladicek: Some good points there. Indeed I chose the `(Class, String)` representation of method references to enable generating built invokers during application build. I'm not terribly fond of it, but it's reasonably simple and I'm going to claim that anyone writing CDI extensions is very likely to have at least _some_ experience with the Java reflection API, so this will hardly be fully alien to them. Using lambdas probably makes generating invokers during build harder, if not impossible, but I didn't really try it, so take that with a grain of salt.

When it comes to Javadoc, I expect that a huge part of what's there currently will move to the specification text and the javadoc will only contain the more practical parts (and examples). That includes the possible signatures of transformers, if we stay with `(Class, String)`, listed directly on the builder methods, but for now, I wanted to keep that on a single place.

ljnelson:
Do I understand properly that a so-called "invokable method" always belongs to a _managed_ bean?

Ladicek:
> Do I understand properly that a so-called "invokable method" always belongs to a _managed_ bean?

Correct, but note that I use the term _managed bean_ in the very specific CDI sense: https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#managed_beans That is, for example, producer beans are not managed beans.

Ladicek:
A few more points made on the call yesterday.

1. Currently, the `InvokableMethod` interface allows obtaining a direct invoker, which always exists, and also all previously registered invokers, if the caller has the registration key. The registration key is a `Class`, which is supposed to allow for some level of privacy, but that can easily be circumvented using `Class.forName`. The more I think about it, the more I believe we should get rid of `InvokableMethod` and the concept of direct invoker and the concept of registration key, always requiring framework integrators to build the invokers they need during deployment and stash them somewhere for usage at application runtime. That would probably simplify the API quite a bit, and streamline the expected usage path.

2. While the current API only allows building invokers during deployment time, CDI Full could allow building invokers at application runtime. Not sure if that's useful, but it's an option we could add.

Ladicek:
Just added one commit to simplify the invoker registration process.

This commit removes the ability to obtain an invoker without previous registration. It also removes the concept of registration key: whenever an invoker is registered, the extension must remember it for later use.

After that, the `InvokableMethod[Info]` interfaces are useless, hence this commit removes them. Access to invoker registration is through the extension API: `ProcessManagedBean` and `BeanInfo`. Both of these places are provisional; better API for invoker registration likely exists.

This commit does not (yet) remove the `MethodMetadata` and `ParameterMetadata` interfaces, because there is one more possible use case for them: the invoker transformers may want to access the called method, including transformed annotations. The API definition currently does not allow that, but it's something that should be explored.

api/src/main/java/jakarta/enterprise/invoke/Invokable.java
```diff
@@ -0,0 +1,49 @@
...
+ * <p>
+ * Attempting to mark a private method or a constructor of a managed bean class
+ * as invokable is a definition error.
+ * <p>
+ * Methods that are marked as invokable but do not belong to managed beans are ignored.
+ * <p>
+ * Note that multiple managed beans may inherit an invokable method from a common
+ * supertype. In that case, each bean conceptually has its own invokable method
+ * and for example an invoker obtained for one bean cannot be used to invoke
+ * the method on the other bean.
```
manovotn, line 36: I don't think this isolation makes much sense.
You can have something like class `Foo` with method `ping()` and then you can have `n` implementations (`FooImpl1`, `FooImpl2`, ...) of this class that are beans.
Some of the impls you have invokers for might be disabled, specialized, vetoed, ....
You'd need to create and keep reference to a separate invoker for each variant of the bean - despite the fact that you don't care about what else impl variants bring to the table, in fact you might not even care which impl it is, you just want to get the default one that comes via `@Inject Foo` and execute the method on that one.
Determining which invoker you can use at runtime could become a hassle.

I'd rather have an invoker bound to a method and so long as you pass in a "compatible" bean which has such method, you invoke it there.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticObserverBuilder.java
```diff
@@ -334,6 +334,34 @@ public interface SyntheticObserverBuilder<T> {
...
+     * <p>
+     * When looked up from the parameter map in the event notification function, the value will be
+     * an instance of {@link jakarta.enterprise.invoke.Invoker Invoker}, <em>not</em> an {@code InvokerInfo}.
+     *
+     * @param key the parameter key, must not be {@code null}
+     * @param value the parameter value
+     * @return this {@code SyntheticBeanBuilder}
+     * @since 5.0
+     */
+    SyntheticBeanBuilder<T> withParam(String key, InvokerInfo value);
```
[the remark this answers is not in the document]
Ladicek: Thanks!

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticObserverBuilder.java
```diff
@@ -334,6 +334,34 @@ public interface SyntheticObserverBuilder<T> {
...
+     * <p>
+     * When looked up from the parameter map in the event notification function, the values will be
+     * instances of {@link jakarta.enterprise.invoke.Invoker Invoker}, <em>not</em> {@code InvokerInfo}.
+     *
+     * @param key the parameter key, must not be {@code null}
+     * @param value the parameter value
+     * @return this {@code SyntheticBeanBuilder}
+     * @since 5.0
+     */
+    SyntheticBeanBuilder<T> withParam(String key, InvokerInfo[] value);
```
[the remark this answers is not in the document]
Ladicek: Thanks!

manovotn:
@Ladicek there is a javadoc build error
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.3.0:jar (attach-javadocs) on project jakarta.enterprise.cdi-api: MavenReportException: Error while generating Javadoc: 
Error:  Exit code: 1 - /home/runner/work/cdi/cdi/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/InvokerInfo.java:17: error: reference not found
Error:   * Opaque token that stands in for an invoker registered using {@link BeanInfo#registerInvoker(MethodInfo)}.
```

Ladicek:
Hmm, I'll look at the JDK 17 failures tomorrow.

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
graemerocher: what I don't like here is the assumption that reflection will be used. The `Class` parameter should be `ClassInfo`

manovotn: This is a user facing API - how is user going to get `ClassInfo` object for their arbitrary `Foo` class? That class doesn't even need to be a bean, so you cannot get that from another BCE event either.
And as far as I can tell, we don't have a way to create that in the lang model API either.

If we wanted to go without `Class`, we'd probably need to go full `String` mode which is pretty ugly...

Ladicek: So the `Class` parameter does not mean reflection needs to be used. It is a possible implementation, but not the only one. The API is designed so that it can be implemented purely at build time. I actually have a fairly complete implementation of this API in ArC, where all invokers created through the `InvokerBuilder` are generated during build. In such implementation, the `Class` object is only used to obtain the binary name of the class, nothing else. I'd be happy to present how my implementation works, and I'd of course love to hear if this API contains any obstacles for your implementation.

I'll experiment with using method references, but I fear that it isn't exactly easy to extract the necessary information from them during build time, as a method reference is indistinguishable from a lambda expression.

Ladicek: Actually thinking about it, I see one possible problem for your implementation (based on what I know -- if you could confirm, that would be great). The specification currently assumes that an implementation can inspect the class so that it can:

- find the named method, and use its signature for code generation,
- do the limited type checking the specification prescribes.

I didn't notice any issue with that, because our implementation has full access to the deployment. That isn't exactly the case for your implementation, if I remember correctly.

Now, indeed one way to solve it would be to use reflection to inspect the class. That would be reflection during build time, not at runtime -- but this might still be problematic. I'm no expert here, so please correct me if I'm wrong.

I assume if we could pull off some trick with method references, that would fix the issue for you, as the method reference would come with the full signature and you wouldn't have to search for the method on the class. Is that right?

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
```diff
@@ -0,0 +1,363 @@
...
+ * of the same kind, or for the same argument position, leads to an exception.
+ * <p>
+ * Wrappers must be {@code static} and must be declared directly on the given class.
+ * It is possible to register only one wrapper. Attempting to register a second wrapper
+ * leads to an exception.
+ * <p>
+ * It is a deployment problem if no method with given name and valid signature is found,
+ * or if multiple methods with given name and different valid signatures are found. It is
+ * a deployment problem if a registered transformer or wrapper is not visible in the package
+ * that contains the class which declares the invokable method.
```
manovotn, line 111: This is not correct - the visibility of transformer/wrapper method from the package of the bean is irrelevant. Both are called independently by the impl which needs to look up those methods.
In fact, we should require transformers/wrappers to be `public`, unless we want the impl to enforce lookup via something like `MethodHandles.privateLookupIn(...)` which isn't very nice :)

Ladicek: Yeah, in general, the author of the transformer and the author of the managed bean are different, so in practice, this rule collapses to "must be `public`" anyway. I agree we should just demand that transformers and wrappers are always `public`.

api/src/main/java/jakarta/enterprise/invoke/Invokable.java
Ladicek: Good question, conceptually. I probably had more reasons for it, which I don't remember exactly right now, but one is fairly obvious: if I want an invoker to be able to automatically lookup the target instance (`InvokerBuilder.setInstanceLookup()`), then the invoker must be bound not only to the method, but also to the _bean_.

manovotn: Fair enough but you get the invoker from a `ProcessManagedBean` even where you can see bean types (and qualifiers). You can use those for lookup regardless of what the resulting bean is - actually that's exactly what will happen anyway if you have an alternative in place; you'll end up invoking it on different impl instance.

On the topic of instance lookup, assuming the rules weren't this strict, I was thinking if the method could be something like `InvokerBuilder.setInstanceLookup(TypeLiteral<T> t, Annotation... a)` allowing you to specify how to perform the lookup while still leaving dependent instance handling to CDI. Obviously this brings the issue of assignability but that's the same as with many other parameters user can wrongly pass in.

Ladicek: Good point about alternatives -- I don't think the existing spec text (in the javadoc) mentions it, but I think we should only discover invokable methods on enabled beans, or something like that.

manovotn: Well, `ProcessBean` (or managed bean in this case) is fired for enabled beans only, so that's automatically true, at least via portable extensions. However IIRC, that doesn't solve the alternative issue fully as you can have multiple enabled alternatives (meaning alternatives selected for given archive) for `Foo`, each with a different priority.

Ladicek:
I made a little experiment with method references. I declared this functional interface:

```java
@FunctionalInterface
public interface SimpleTransformer<S, T> extends Serializable {
    T transform(S value);
}
```

And then, assuming that `InvokerBuilder` has a method `setArgumentTransformer(int position, Class<?> clazz, String methodName)`, I could easily add this:

```java
public <X, Y> InvokerBuilder setArgumentTransformer(int position, SimpleTransformer<X, Y> transformer) {
    try {
        Method m = transformer.getClass().getDeclaredMethod("writeReplace");
        m.setAccessible(true);
        SerializedLambda lambda = (SerializedLambda) m.invoke(transformer);
        return setArgumentTransformer(position, Class.forName(lambda.getImplClass().replace('/', '.')), lambda.getImplMethodName());
    } catch (ReflectiveOperationException e) {
        throw new RuntimeException(e);
    }
}
```

This allows referring to trasnformer methods using method references: `setArgumentTransformer(0, SomeClass::transform)`. So that's nice, but also a huge huge horrible hack that I would very much hesitate to put into a specification.

Any thoughts? Keep in mind that build-time implementations should be able to generate invokers during build, so the `InvokerBuilder` is actually also called during build.

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
```diff
@@ -0,0 +1,363 @@
...
+ * These outputs correspond to the return value of {@link Invoker#invoke(Object, Object[]) Invoker.invoke()}
+ * or its thrown exception, respectively.
+ * <p>
+ * Each output can be transformed by a transformer that has one of the following signatures,
+ * where {@code X} and {@code Y} are types:
+ *
+ * <ul>
+ * <li>{@code static X transform(Y value)}</li>
+ * <li>{@code X transform()} &ndash; in this case, {@code Y} is the type of the class that
+ * declares the transformer</li>
```
manovotn, line 195: How do you aim to use this with exception transformer? I think this format only applies to return value transformers or am I missing something @Ladicek?

Ladicek: I'm not sure why this couldn't apply to exception transformers...? Could you elaborate?

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
graemerocher: so I see there is `InvokerBuilder<InvokerInfo> createInvoker(MethodInfo method);` so it is probably ok.

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
manovotn: We discussed this over chat and the conclusion is that you can technically use this to register something like `Throwable::getMessage` as a transformer.
However, due to another part of the text here:

> it is a deployment problem if Y is not assignable [...] from java.lang.Throwable in case of exception transformers

You are very limited in use cases for that. In almost all cases, an exception transformer will still be a specific static method with `Throwable` param.

Ladicek: That is correct, the usefulness of this style for exception transformers is very low, but there's no reason to forbid it.

manovotn:
Just for the record, I wanted to provide people with some playground for this feature so I've started working on a Weld prototype supporting current proposed API.
Current state (what's supported/TODOs) along with instructions on how to build it will be in the first comment on Weld PR - https://github.com/weld/core/pull/2864
There is also a growing set of tests showing common/expected use cases.

manovotn:
Just a note WRT to the recent discussion on the call around exception transformer arbitrary return values versus what method handles allow. I have managed to implement a workaround in Weld that can support the API as proposed here - the crux of it is to apply a return value transformation to the exception transformer and then "hide" the actual transformer ret. value within an exception. It's not the prettiest thing, but it can be done conditionally and still using just method handles allowing me to construct one final handle and then invoke it repeatedly without too much overhead.
In case you're curious how it works, see the [third commit in the Weld PR](https://github.com/weld/core/pull/2864/commits).

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
```diff
@@ -0,0 +1,375 @@
...
+ * the description in {@link Invoker#invoke(Object, Object[]) Invoker.invoke()}, depending
+ * on configured lookups, transformers and wrapper. Some types are checked during
+ * deployment, as described in previous sections. Other types are checked during invocation,
+ * at the very least due to the type checks performed implicitly by the JVM. The lookups,
+ * transformers and the wrapper must arrange the inputs and outputs so that when the invokable
+ * method is eventually invoked, the rules described in
+ * {@link Invoker#invoke(Object, Object[]) Invoker.invoke()} all hold.
+ * <p>
+ * TODO specify what happens when a transformer/wrapper declares a parameter of a primitive type
+ *  but the actual value passed to the invoker is `null` (the transformer should get a zero value?)
```
Azquelt, line 286: I'd expect `NullPointerException` in this case, as you would get if you tried to unbox a `null`.

Azquelt:
I know this has been answered before, but I can't remember the answer and couldn't find it written down anywhere.

**Why do we need the `@Invokable` annotation?**

Why can't I create an invoker for any method? Invokers are always created by extensions, which are running at startup. In addition, we already have the capability to get information on any method via `@Enhancement` with `ClassInfo` or `MethodInfo` so I'd have thought that whatever information is needed to create an invoker must already be available to support `@Enhancement`?

Ladicek:
Good question. Arguably `@Invokable` is not as meaningful now as it might have been before, because one must always create an invoker explicitly. Previously, all invokable methods got a direct invoker automatically, for free, but that's no longer the case.

The annotation has never been intended to be used in applications directly, it should be mainly a meta annotation for other annotations that constitute invokable markers, but the only argument for the existence of invokable markers today is that they explicitly designate methods for which an invoker may be built. That is, it's some sort of documentation. I agree that's a fairly weak argument.

I'll sleep on it, but at the moment, I think removing the `@Invokable` annotation would be harmless and I'd be in favor.

manovotn:
> The annotation has never been intended to be used in applications directly, it should be mainly a meta annotation for other annotations that constitute invokable markers, but the only argument for the existence of invokable markers today is that they explicitly designate methods for which an invoker may be built. That is, it's some sort of documentation. I agree that's a fairly weak argument.

That depends on whether we want basically any method to be invokable.
Note that we currently disallow private methods and constructors but everything else is allowed. You could therefore create invokable method off of any package private method that you normally cannot invoke.
Granted, you could do that now if the method is marked `@Invokable` but that's still different from opening up all of them by default.

The only other thing I can think of is annotation processor approach perhaps needing the annotation? But then again, if they can meet the requirements for `@Enhancement`, then I don't suppose this is required (at least not as long as we talk managed class bean's methods).

Ladicek:
Good point about annotation processing-based implementations. I'm no expert, but I think they should be fine -- invokable methods only exist on managed beans, which already must have a bean defining annotation (in CDI Lite), so we should be fine. I'd be glad to be proven wrong (or right) though.

Ladicek:
I removed the `@Invokable` annotation and renamed the entire feature to _method invokers_. This should resolve one of the greatest pain points of the proposal: the correct spelling of "invokable" / "invocable", because that word doesn't appear in the proposal anymore :-)

Ladicek:
I also marked this PR as ready for review. There's still a lot of open questions (notably: should we use the `(Class, String)` pair to identify transformers/wrappers or should we require users to implement an interface, and what exactly are the assignability rules?), but I believe the current state is good enough for inclusion in an alpha release.

api/src/main/java/jakarta/enterprise/inject/spi/ProcessManagedBean.java
```diff
@@ -38,4 +41,16 @@ public interface ProcessManagedBean<X> extends ProcessBean<X> {
...
+     * Returns a new {@link InvokerBuilder} for given method. The builder eventually produces an invoker
+     * for the given method.
+     * <p>
+     * The {@code method} must belongs to the bean being registered, otherwise an exception is thrown.
+     *
+     * @param method method of the bean being registered, must not be {@code null}
+     * @return the invoker builder, never {@code null}
+     * @since 4.1
+     */
+    public InvokerBuilder<Invoker<X, ?>> createInvoker(AnnotatedMethod<? super X> method);
```
Azquelt, line 55: Should this be on `ProcessBean` instead of `ProcessManagedBean`?

I can't think of a reason you shouldn't be able to create an invoker for a method for a producer method, a producer field, a session bean or a synthetic bean.

Possibly you shouldn't be able to create one for an interceptor or decorator.

api/src/main/java/jakarta/enterprise/inject/spi/ProcessManagedBean.java
```diff
@@ -38,4 +41,16 @@ public interface ProcessManagedBean<X> extends ProcessBean<X> {
      * @throws IllegalStateException if called outside of the observer method invocation
      */
     public AnnotatedType<X> getAnnotatedBeanClass();
+
+    /**
+     * Returns a new {@link InvokerBuilder} for given method. The builder eventually produces an invoker
+     * for the given method.
+     * <p>
+     * The {@code method} must belongs to the bean being registered, otherwise an exception is thrown.
```
Azquelt, line 49:
```suggestion
     * The {@code method} must belong to the bean being registered, otherwise an exception is thrown.
```

api/src/main/java/jakarta/enterprise/inject/spi/ProcessManagedBean.java
manovotn: > Should this be on `ProcessBean` instead of `ProcessManagedBean`?

I don't think so; the idea was to support methods declared on class-based beans.
@Ladicek I am pretty sure we discussed this but I don't recall what were the reasons for this exactly?

> I can't think of a reason you shouldn't be able to create an invoker for a method for a producer method, a producer field, a session bean or a synthetic bean.

A producer method has to be declared on a CDI bean - therefore you can create the invoker for it there.
A producer field isn't a method at all; a poor fit for method invokers :)
A session bean will still work, because `ProcessSessionBean` is a subtype of `ProcessManagedBean`
A synthetic bean - I am not sure I follow here. Anything can be a synthetic bean and the logic of these closely follows that of producers. Do you mean the return type of those beans? Note that it can be (just like with producers) even a primitive type.

> Possibly you shouldn't be able to create one for an interceptor or decorator.

Yes, that doesn't make much sense.

api/src/main/java/jakarta/enterprise/invoke/Invoker.java
```diff
@@ -0,0 +1,116 @@
...
+ *     String invoke(MyService instance, Object[] arguments) {
+ *         String argument = (String) arguments[0];
+ *         String result = instance.hello(argument);
+ *         return result;
+ *     }
+ * }
+ * </pre>
+ *
+ * Therefore, calling {@code invoker.invoke(myService, new Object[] {"world"})}
+ * will return {@code "Hello world!"}.
```
Azquelt, line 52: This example is helpful because it shows how the arguments to `invoke` are used, but at the same time it's a bit misleading because there's no information about why you would want to do this.

I can see a few options:
- We could state that this is a _basic_ invoker and that invokers with more complex behaviour can be created with `InvokerBuilder`.
- We could remove the example altogether and direct users to the examples on `InvokerBuilder`
- We could remove the "equivalent class" part and just give a more complete example of creating an invoker with `InvokerBuilder` and then calling it.

Ladicek: Good point, this is indeed a simple invoker with no extra features, which is not exactly apparent in this example. I'll try to expand it.

api/src/main/java/jakarta/enterprise/inject/spi/ProcessManagedBean.java
Azquelt: > A producer method has to be declared on a CDI bean - therefore you can create the invoker for it there.
> A producer field isn't a method at all; a poor fit for method invokers :)

I'm not suggesting creating an invoker for the producer field or method itself, but for methods on the bean types of the bean created by the producer method or field. (i.e. usually the return type of the producer method, or the type of the producer field).

E.g. if I have a producer method
```java
@ApplicationScoped
public Connection getConnection() { /*...*/ }
```
then I should be able to create an invoker for `Connection.prepareStatement(String)`

Azquelt: Possibly we need to clarify what's meant by "The {@code method} must belong to the bean being registered".

I assumed it meant you must be able to call the method on one of the bean types of the bean.

Ladicek: So I originally wanted this to only apply to managed beans, because you know the exact class of the bean and the full set of its methods. It's also the only thing that I have actual use cases for.

It would probably be possible to extend this to producers, where you at least know the supertype of the bean and only the methods declared on that supertype (or inherited from further supertypes) could be invoked.

For synthetic beans, there's no way to obtain anything about the bean implementation class, so that's a no-go.

manovotn: > I'm not suggesting creating an invoker for the producer field or method itself, but for methods on the bean types of the bean created by the producer method or field. (i.e. usually the return type of the producer method, or the type of the producer field).

I see, thanks for clarification.
Currently, invokers are created (in Portable Extensions) from `AnnotatedMethod` - obtaining AM for arbitrary method of arbitrary type might be a little challenging whereas getting them from `ProcessManagedBean` is straightforward.

> Possibly we need to clarify what's meant by The `{@code method}` must belong to the bean being registered.

I think this is in place because otherwise you could observe `ProcessManagedBean<Foo>` and try to register an invoker for an `AnnotatedMethod` from bean `Bar`.

Ladicek: I'll try to expand the word "belong", but @manovotn is right -- this is to prevent using `ProcessManagedBean<Foo>` to create an invoker for a method of a completely different class. How about:

```
The {@code method} must be declared on the bean class of the bean being registered
or inherited from a supertype of the bean class, otherwise an exception is thrown.
```

api/src/main/java/jakarta/enterprise/inject/spi/ProcessManagedBean.java
Ladicek: I'll reword this based on feedback above, so there will be no "belong" anymore.

api/src/main/java/jakarta/enterprise/invoke/Invoker.java
Ladicek: Expanded the example, which prompted moving the _Obtaining an invoker_ section above the _Example_.

api/src/main/java/jakarta/enterprise/inject/spi/ProcessManagedBean.java
Ladicek: Expanded the comment per above.

api/src/main/java/jakarta/enterprise/invoke/Invokable.java
Ladicek: The annotation is gone now.

api/src/main/java/jakarta/enterprise/inject/spi/ProcessManagedBean.java
Azquelt: That's much clearer :+1:

If the intention is that you can only create an Invoker for a managed bean, then the documentation on `BeanInfo` should probably be updated to reflect that.

Ladicek: That's actually specified in `Invoker`, but I added a note to `BeanInfo` as well.

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
```diff
@@ -0,0 +1,369 @@
...
+ * <p>
+ * For the target instance, a CDI lookup is performed with the required type equal to the bean
+ * class of the bean to which the target method belongs, and required qualifiers equal to the set
+ * of all qualifier annotations present on the bean class of the bean to which the target method
+ * belongs. When the target method is {@code static}, the target instance lookup is skipped.
+ * <p>
+ * For an argument, a CDI lookup is performed with the required type equal to the type of
+ * the corresponding parameter of the target method, and required qualifiers equal to the set
+ * of all qualifier annotations present on the corresponding parameter of the target method.
+ * <p>
```
Azquelt, line 131: At some point, we probably need to be more specific about what we mean by a CDI lookup.

I guess at deployment we're performing typesafe resolution, and then when the method is invoked we're obtaining a contextual instance (or contextual reference?) of the Bean?

Azquelt (approved):
This looks good. I'm happy for this to be merged with the remaining TODOs, so that we can get an alpha release out.

I still have some reservations about doing argument, return type and exception transformations, but we can extend those in future releases if it turns out that users want additional functions and it's always possible for users to ignore our methods and use another mechanism (e.g. method handles) if they need more advanced transformations.

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
Ladicek: It would be a contextual reference. But I think I'd like to keep that specified in terms of [programmatic lookup](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#programmatic_lookup), because that shields us from these low-level concerns.

manovotn (approved):
Going to merge this and issue an Alpha release.

Following that, I would like to release Weld version with the prototype so that we can (hopefully) get some feedback.
There are still some TODOs here that we should create a tracking issue for (or keep existing one open) as well as create tracking issue for TCKs.

PR #643, opened 2022-12-24
docs/README 

ekimike:
Before pointing out that the Ruby 2.x version could be the only one for running Jekyll, I think it is better to put it as a recommendation instead of a fact because doing fast research, it seems Jekyll could work with Rubby 3.x, but I didn't test it.

The second change relates to the URL, just added the final part.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -7,7 +7,7 @@ These instructions will get you a copy of the eclipse-ee4j.github.io/cdi/ websit
 ### Installation
 [Jekyll static site generator docs](https://jekyllrb.com/docs/).
 
-1. Install a full [Ruby development environment](https://jekyllrb.com/docs/installation/) (It is preferred version 2.x)
+1. Install a full [Ruby development environment](https://jekyllrb.com/docs/installation/) (Version 2.x is recommended, version 3.x seems to be incompatible)
 2. Install [bundler](https://jekyllrb.com/docs/ruby-101/#bundler)  [gems](https://jekyllrb.com/docs/ruby-101/#gems)
 
         gem install bundler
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -7,7 +7,7 @@ These instructions will get you a copy of the eclipse-ee4j.github.io/cdi/ websit
 ### Installation
 [Jekyll static site generator docs](https://jekyllrb.com/docs/).
 
-1. Install a full [Ruby development environment](https://jekyllrb.com/docs/installation/)
+1. Install a full [Ruby development environment](https://jekyllrb.com/docs/installation/) (It is preferred version 2.x)
```
manovotn, line 10: This sounds a little clumsy, I'd reword it a bit.
```suggestion
1. Install a full [Ruby development environment](https://jekyllrb.com/docs/installation/) (Version 2.x is recommended)
```

manovotn:
> it seems Jekyll could work with Rubby 3.x, but I didn't test it.

I *think* I did try with 3.x some time ago and failed to make it work in current setup but I don't recall the details.
Phrasing it as recommendation is IMO fine.

ekimike:
Could you check this again, please?

docs/README.md
```diff
@@ -7,7 +7,7 @@ These instructions will get you a copy of the eclipse-ee4j.github.io/cdi/ websit
 ### Installation
 [Jekyll static site generator docs](https://jekyllrb.com/docs/).
 
-1. Install a full [Ruby development environment](https://jekyllrb.com/docs/installation/)
+1. Install a full [Ruby development environment](https://jekyllrb.com/docs/installation/) (Version 2.x is recommended, version 3.x seems to be incompatible)
```
Ladicek, line 10: I just tried to follow these instructions locally and hit troubles due to having Ruby 2.5. In particular, the pinned version of Nokogiri seems to require at least Ruby 2.6. I installed Ruby 2.7 and everything worked fine. I didn't have the courage to try Ruby 3.x :-)

```suggestion
1. Install a full [Ruby development environment](https://jekyllrb.com/docs/installation/) (Version 2.7 is recommended, version 3.x seems to be incompatible)
```

Ladicek (approved):

PR #657, opened 2023-03-07
Improve AnnotationLiteral javadoc

### LINKED ISSUE #656: Improve `AnnotationLiteral` javadoc
The javadoc of `AnnotationLiteral` should *discourage* instantiation using anonymous classes.

Definition of `Annotation#equals`, among other thing, says that:

> returns true if the specified object is an instance of the same annotation interface as this instance

However, this can never be honored by anonymous class for `AnnotationLiteral` as it simply doesn't implement given annotation.
Standard class extending `AnnotationLiteral` can and does that just fine.

We could change the javadoc from saying just:

> An instance of an annotation type may be obtained by subclassing `AnnotationLiteral`.

To something like:

> An instance of an annotation type may be obtained by subclassing `AnnotationLiteral`.
The subclass must implement the annotation interface to satisfy the {@link Annotation} contract.

Side note: There are TCK tests that use anonymous `AnnotationLiteral` that we should look into as well. The change will be trivial.

Ladicek:
Fixes #656

### DIFF
--- a/api/src/main/java/jakarta/enterprise/util/AnnotationLiteral.java
+++ b/api/src/main/java/jakarta/enterprise/util/AnnotationLiteral.java
@@ -38,6 +38,7 @@ import jakarta.enterprise.inject.Instance;
  * 
  * <p>
  * An instance of an annotation type may be obtained by subclassing <code>AnnotationLiteral</code>.
+ * The subclass must implement the annotation interface to satisfy the {@link Annotation} contract.
  * </p>
  * 
  * <pre>


manovotn (approved):

mkouba (approved):

api/src/main/java/jakarta/enterprise/util/AnnotationLiteral.java
```diff
@@ -38,6 +38,7 @@
  * 
  * <p>
  * An instance of an annotation type may be obtained by subclassing <code>AnnotationLiteral</code>.
+ * The subclass must implement the annotation interface to satisfy the {@link Annotation} contract.
```
Emily-Jiang, line 41:
```suggestion
 * The subclass must implement the Annotation interface to satisfy the {@link Annotation} contract.
```

Ladicek: One thing I could do is replace "annotation interface" with "annotation type", as "annotation type" is used above this sentence. However, "the subclass must implement the annotation type" is not exactly clear in that "implement" means the `implements` keyword...

Emily-Jiang: ah, okay. It makes sense! Thanks for your explanation!

PR #658, opened 2023-03-09
Clarify how vetoing a specialized bean works

manovotn:
A suggestion for clarification related to recent discussion in https://github.com/jakartaee/cdi-tck/issues/440

It tried to clarify two unclear things:
* It states that bean re-enablement happens if you `veto()` specializing bean
  * I found no other place in spec that would say this happens but I don't think there is a disagreement on that since keeping the original bean disabled provides no benefit.
* It explicitly defines what happens WRT `ProcessBeanAttributes` of the newly re-enabled bean
  * This is in line with [bean discovery](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#bean_discovery_steps_full) chapter requiring that PBA is fired for enabled beans only

This change follows what the TCK test already asserts.

Thoughts are welcome.

CC @Ladicek @jeanouii

### DIFF
--- a/spec/src/main/asciidoc/core/spi_full.asciidoc
+++ b/spec/src/main/asciidoc/core/spi_full.asciidoc
@@ -1300,7 +1300,7 @@ public interface ProcessBeanAttributes<T> {
 * `configureBeanAttributes()` returns a `BeanAttributesConfigurator` (as defined in <<bean_attributes_configurator>>) initialized with the `BeanAttributes` processed by the event to easily configure the  `BeanAttributes` which will be used to replace the original one at the end of observer invocation.
 The method always returns the same `BeanAttributesConfigurator`.
 * `addDefinitionError()` registers a definition error with the container, causing the container to abort deployment after bean discovery is complete.
-* `veto()` forces the container to ignore the bean.
+* `veto()` forces the container to ignore the bean. If invoked on a bean `A` that directly specializes bean `B` (as defined in <<specialize_managed_bean>>), this causes re-enablement of the specialized bean `B` and the container must raise a `ProcessBeanAttribute` event for newly re-enabled bean `B`.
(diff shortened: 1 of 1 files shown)


rmannibucau:
Hi,

I think the spec has a big issue as of today which is to mix "bean" word in all the lifecycle before there is any `Bean<?>`.
Concretely this means the spec does *not* say anywhere there is a disablement of `ProcessBeanAttributes` for any bean (since it happens before beans are a thing so the sentence is a non-sense from a spec perspective even if I "guess" the intent) - likely why TCK were challenged AFAIK.

So conretely I think the adjustment would rather be:

1. Ensure there is no disablement while there is a potential event which can veto any bean (means fire all `ProcessBeanAttributes`). This is 1. to avoid the ambiguity to not be able to modify a bean which will be disabled - but present/usable programmatically and 2. the lifecycle ambiguity for specialization - sorted or not, this is not defined in the spec in current state.
2. Once you can't veto anymore any bean (and therefore decide which ones are disabled or not) then disable the beans accordingly the CDI rules (this is then fully defined).
3. Fix the TCK which are not agreed upon accordingly

The side note there is that not following that can lead to weird cases where assumptions are done on the beans whereas there is actually no guarantee about it at runtime (which is against CDI extension mecanism which intends to validate  the runtime), in particular from an extension point of view.

rmannibucau:
> I am curious to see how exactly you use a disabled bean, especially since there doesn't even need to be a Bean object for it

Did you just explain the spec is absurd cause you dont need a bean to disable a bean? This is exactly one of the points and what makes the spec not implementable and therefore the test a valid challenge.

> Note that this suggestion is breaking backwards compatibility as the specification clearly defines this for enabled beans only.

Spec always had been ambiguous so I agree it can break weld (but weld has other broken cases in that area) so not a big deal, on a spec standpoint no regression, it is still undefined by a very bad wording making it impossible to implement and therefore impossible to test.

manovotn:
Closing my own PR as it's stale.

PR #676, opened 2023-05-31
Move the specification of BeanManager.getInjectableReference() to CDI Full

Ladicek:
We previously [1] moved the `getInjectableReference()` from the `BeanContainer` interface to the `BeanManager` [2], but did not update the specification text accordingly. Thus, the method is specified in the CDI Lite part of the spec, but it in fact belongs to CDI Full. This commit moves the spec text around to fix that.

Additionally, this commit removes the declaration of `resolveObserverMethods()` from `BeanManager`, because that just overrides the same declaration that is present on `BeanContainer`.

Finally, this commit moves the declaration of `getInjectableReference()` from the end of the `BeanManager` source code to the beginning, for symmetry with the `BeanContainer` declaration and to restore the previous layout.

[1] https://github.com/jakartaee/cdi/commit/520bad1989101bf4523e13e8e0c8f2d23cdc34b6
[2] https://github.com/jakartaee/cdi/issues/568

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/spi/BeanManager.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/BeanManager.java
@@ -93,6 +93,22 @@ import jakarta.enterprise.inject.Stereotype;
 public interface BeanManager extends BeanContainer {
 
     /**
+     * <p>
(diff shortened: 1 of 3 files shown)


manovotn (approved):
Thanks, this is an oversight from previously done changes.

PR #690, opened 2023-08-30
Fix guards against repeated external sets of SPI implementations

Ladicek:
We previously [1] added code to guard against repeated calls to these SPIs:

- `BuildServicesResolver.setBuildServices()`
- `CDI.setCDIProvider()`

The guards are unfortunately incorrect. This commit fixes them.

[1] https://github.com/jakartaee/cdi/commit/82ec5d71872212364a321cd3700d0ecadf9162d2


mkouba:
> Fortunately, the code that we previously added did not make it into any release, it's only in the `master` branch now. So we can just fix it.

Good catch! :+1:

manovotn (approved):
Thanks for noticing this!

PR #696, opened 2023-10-25
Allow setting CDI SPI implementations repeatedly

### LINKED ISSUE #695: `CDI.setCDIProvider()` throwing on a repeated set is inconvenient
The `CDI.setCDIProvider()` method has been documented to throw `IllegalStateException` on repeated call since version 1.0, but since version 2.0, the implementation in fact didn't honor this documentation. Multiple implementations started to silently depend on repeated `setCDIProvider()` being allowed.

Now that we actually fixed that for 4.1 (see #619 and #690), it turns out that this is very inconvenient. I know @manovotn and @Azquelt had problems with Weld, I stumbled upon this problem with ArC today as well.

We briefly discussed this on the CDI call yesterday and there was no objection to actually making this officially allowed (it was practically allowed since CDI 2.0).

Ladicek:
The `CDI.setCDIProvider()` method has been documented to throw `IllegalStateException` on repeated calls since version 1.0, but since version 2.0, the implementation in fact didn't honor this documentation. Multiple implementations started to silently depend on repeated `setCDIProvider()` calls being allowed, especially for testing.

We previously [1][2] fixed the implementation to actually do what it documents, planning to release the fix in CDI 4.1, but it turns out this is very inconvenient. Therefore, this commit makes repeated calls of `setCDIProvider()` officially allowed. For consistency, it also allows repeated calls of `BuildServicesResolver.setBuildServices()`.

[1] https://github.com/jakartaee/cdi/commit/82ec5d71872212364a321cd3700d0ecadf9162d2
[2] https://github.com/jakartaee/cdi/commit/8ead70bdd135d277e277dfffc086b58dd92a1de3

Fixes #695

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildServicesResolver.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildServicesResolver.java
@@ -76,16 +76,11 @@ public final class BuildServicesResolver {
      *
      * @param instance a {@link BuildServices} instance that should be used, must not be {@code null}
      * @throws IllegalArgumentException if the provided argument is null
-     * @throws IllegalStateException if the {@link BuildServices} are already set
      */
     public static void setBuildServices(BuildServices instance) {
         if (instance == null) {
             throw new IllegalArgumentException("BuildServices instance must not be null");
         }
-        if (configuredBuildServices == null) {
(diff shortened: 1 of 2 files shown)


manovotn (approved):

Ladicek:
No objections against this were raised on the call today, gonna merge :-)

PR #697, opened 2023-11-06
Initial specification text for method invokers

Ladicek:


Ladicek:
This is an initial specification text for the `Invoker` part of method invokers. It doesn't include anything about `InvokerBuilder` except of the `build()` method. Lookups, transformers and wrappers are for later (I've recently started thinking that it might indeed be best to postpone some of these features to a later release, because there has been little practical experience with them, just my prototypes).

I also started writing TCK tests for this, but I need a few more days as I also need to figure out if some of the requirements stated here are straightforward to implement with Weld. Specifically, the rules about throwing exceptions in certain situations may need to be relaxed. I hope I'll be able to submit a PR to the TCK repo in a few days.

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -0,0 +1,181 @@
...
+The `Invoker` interface contains a single method:
+
+[source,java]
+----
+public interface Invoker<T, R> {
+    R invoke(T instance, Object[] arguments);
+}
+----
+
+This section describes the behavior of the `invoke()` method.
```
manovotn, line 49: This sentence seems superfluous as the intention of the text below is rather obvious.

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -0,0 +1,181 @@
...
+
+Calling `invoke()` invokes the target method on given `instance` of the target bean, passing given `arguments`.
+
+If the target method is `static`, the `instance` is ignored; by convention, it should be `null`.
```
manovotn, line 53: The below list (all of the `if` statements) might be better readable as bullet points.
That being said, all of this creates a hard to read section if you are just trying to read up on how to use invokers.
Personally I'd rather keep these details in the javadoc on the assumption that javadoc text is also considered a source of truth for how CDI should behave (which I am not sure it is). But that might be just me :)

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -0,0 +1,181 @@
...
+
+If the target method is not `static` and `instance` is `null`, a `NullPointerException` is thrown.
+If the target method is not `static` and the `instance` is not <<invoker_assignability,assignable>> to the target bean, a `ClassCastException` is thrown.
+
+Correspondence between given `arguments` and declared parameters of the target method is positional: the Nth element of the `arguments` array is passed as the Nth argument to the target method.
+If the target method is a variable arity method, the last element of the `arguments` array corresponds to the variable arity parameter (and therefore must be an array).
+
+If the target method declares no parameter, `arguments` are ignored.
+If the target method declares any parameter and `arguments` is `null`, `NullPointerException` is thrown.
+If the `arguments` array has fewer elements than the number of parameters of the target method, `ArrayIndexOutOfBoundsException` is thrown.
```
manovotn, line 67: This is one of the points where a method handles implementation won't behave equally (unless you check for this manually) - IIRC, it'd instead throw `IllegalArgumentException`.

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -0,0 +1,181 @@
...
+* if the corresponding argument is not `null`, unboxing conversion is performed on the argument and the result is passed.
+
+If the declared type of a parameter of the target method is not a primitive type, the corresponding argument is passed as is.
+
+// TODO when the `InvokerBuilder` applies transformations, some of the requirements above are no longer strictly necessary, we should reflect that in this text somehow
+
+Invoker implementations must be thread-safe.
+A single invoker instance may be used to perform multiple, possibly concurrent, invocations of the target method, possibly on different instances of the target bean, possibly with different arguments.
+
+Whenever a direct invocation of a method is a business method invocation, an indirect invocation of that method through an invoker is also a business method invocation.
```
manovotn, line 83: This is one of the key behaviors; I'd move it all the way to the `== Method invokers` section.

spec/src/main/asciidoc/core/invokers.asciidoc
manovotn: Also note that this would be even more verbose if we want to mention transformations and wrappers.

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -0,0 +1,181 @@
...
+* `A` is assignable to a primitive type `P` when `A` is the primitive wrapper class of `P`;
+* `A` is assignable to a class or interface type `P` when:
+** `A` is a class or interface type and `A` is a subclass of `P`, or
+** `A` is an array type and `P` is `java.lang.Object`, `java.lang.Cloneable`, or `java.io.Serializable`;
+* `A` is assignable to an array type `P` when `A` is an array type and:
+** the component type of `P` is a primitive type and the component types of `A` and `P` are identical, or
+** the component type of `P` is not a primitive type and the component type of `A` is assignable to the component type of `P` according to these rules.
+
+NOTE: This is a reformulation of the JVM assignability rules as defined by the specification of the `checkcast` and `instanceof` instructions, with addition of primitive types.
+It is expected that CDI implementations do not implement these rules and rely on JVM runtime checks.
```
manovotn, line 178: The above rules are good but given this note and the fact that impls will rely on JVM checks anyway, do we need to explicitly state them? I mean, couldn't we just say they work the same JVM assignability rules with a link there?

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -0,0 +1,181 @@
...
+
+If the target method declares no parameter, `arguments` are ignored.
+If the target method declares any parameter and `arguments` is `null`, `NullPointerException` is thrown.
+If the `arguments` array has fewer elements than the number of parameters of the target method, `ArrayIndexOutOfBoundsException` is thrown.
+If the `arguments` array has more elements than the number of parameters of the target method, the excess elements are ignored.
+If some of the `arguments` is not <<invoker_assignability,assignable>> to the corresponding parameter of the target method, `ClassCastException` is thrown.
+
+If the declared type of a parameter of the target method is a primitive type, then:
+
+* if the corresponding argument is `null`, the default value of the primitive type is passed;
```
manovotn, line 73: Personally, I see this as unexpected and would rather have an exception thrown. The use cases where you'd deliberately use this "shortcut" instead of passing in the default yourself seem much more rare to me than cases in which you accidentally pass in `null` instead of actual non-default value.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
manovotn, line 1: I think the `TODO` in this file can be deleted as well as we agreed that doing this via extensions is the intended way?

manovotn (commented):
Overall looks good, added few comments.
I'd also be +1 for adding docs for transformations/wrapper in this PR as well.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: Fair enough. At some point, it seemed necessary to me, but I'm probably overly cautious here.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: Javadoc is very much a source of truth, which you can see in AtInject (which doesn't have anything else but javadoc). We also did this in Build Compatible Extensions, which are specified in full by the javadoc. Here, I wanted to take the exact opposite approach: put the gory details into the specification text, and leave the javadoc accessible to mere mortals.

Ladicek: Transformers would be in a separate section and we would have to say that the rules specified for `Invoker.invoke()` apply after all transformations and lookups are complete. That should be enough.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: Yes. We'll need to figure out the best way here, based on what both our implementations can easily do, plus taking into account a possible implementation based on classic reflection.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: Agree it is a little hidden here. I'll move it.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: The thing is that they are not exactly the same, because we convert wrapper types to primitive types. It could probably be simplified, but I figured we should just specify this properly :shrug:

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: Yeah. Using default values makes certain things easier, but it is definitely a deviation of how things usually work. We can drop this, for sure.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
Ladicek: If you mean the `BeanInfo.createInvoker()` method, then I'm still not sure that's the best API. I mean, it technically doesn't provide any _information_ about the bean. It could possibly be put on a `BeanConfig`, but we don't have that yet :-)

One other option I considered (but don't have a proper opinion on that yet) is something like:

```java
public interface Invokers {
    InvokerBuilder<InvokerInfo> create(BeanInfo bean, MethodInfo method);
}
```

`@Registration` methods (and maybe also `@Validation`?) could declare a parameter of type `Invokers` (similarly to how they can declare a parameter of type, say, `Messages`).

It's a bit uglier, but maintains proper separation of concerns.

Ladicek:
FYI, I submitted https://github.com/jakartaee/cdi-tck/pull/502 with initial TCK tests that correspond to this PR in its current form. It is passing on my ArC implementation of method invokers, and mostly passes on Weld (with a few modifications that I submitted a PR for).

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: I made this a heading of a subsection that specifies all the boring but necessary details.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: I moved this up in the "Using an `Invoker` chapter, above the newly added subsection that specifies all the boring but necessary details.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: I'm thinking we should allow widening conversion for primitive types, because that's what method handles (as well as reflection) do.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: Agree we should fail on `null` for primitive parameters.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
Ladicek: I'm still not sure about this.

Ladicek:
Rebased and added one commit (to be squashed) that attempts to resolve most of the open questions here. Comments welcome.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: To be clear, I'm not exactly _sure_ about the `InvocationTargetException`. It is something I very much hate about reflection. Maybe we should do this instead:

- specify that for all these validations, a generic `RuntimeException` is thrown
- add `throws Exception` to the `Invoker.invoke()` method
- specify that exceptions thrown by the target method are rethrown

(This would leave custom throwables undefined, which would be consistent with the rest of CDI and Interceptors and everything.)

spec/src/main/asciidoc/core/invokers.asciidoc
manovotn: > I'm thinking we should allow widening conversion for primitive types, because that's what method handles (as well as reflection) do.

That makes sense, +1

spec/src/main/asciidoc/core/invokers.asciidoc
manovotn: I am not sure I understand the ups and downs of wrapping vs re-throwing in this scenario. Would you mind elaborating on that, please?

Ladicek: The upside of wrapping is that callers can easily detect whether the exception was thrown by the invoked method or the invoker mechanism itself failed (presumably because the caller passed an incorrect argument to `invoke()`). The downside is that unwrapping becomes necessary, which is often a source of subtle problems.

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -0,0 +1,175 @@
...
+----
+InvokerBuilder<Invoker<MyService, String>> builder = ...;
+Invoker<MyService, String> invoker = builder.build();
+----
+
+In a build compatible extension, this results in an opaque token that materializes as an `Invoker` at application runtime:
+
+[source,java]
+----
+InvokerBuilder<InvokerInfo> builder = ...;
```
Emily-Jiang, line 107: I think it needs more explanation with the ...

Ladicek: The build compatible extensions case is covered below. The case for portable extensions is covered in the CDI Full section. It would be great to avoid ellipses, but that would require showing much more context, which is detrimental to the purpose of this particular example.

Ladicek:
Rebased and added one commit (to be squashed) that avoids the entire section on assignability for invokers. Turns out that the JLS already defines the exact rules we need (and both reflection and method handles reference those rules as well, under the name of "method invocation conversions").

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: I removed this section entirely, because we can just refer to method invocation conversions, as defined by JLS.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
Ladicek: I still don't like putting `createInvoker` to `BeanInfo`. The above proposal for `Invokers` is ugly, because the name `Invokers` is too similar to `Invoker`, which may lead to frustration. I could call it `InvokerBuilderFactory`, but that name is ugly as well. Maybe this?

```java
public interface InvokerFactory {
    InvokerBuilder<InvokerInfo> createBuilder(BeanInfo bean, MethodInfo method);
}
```

spec/src/main/asciidoc/core/invokers.asciidoc
Emily-Jiang: I suggest adding a comment to link to the relevant section in the doc.

Ladicek: Good point, will do!

Ladicek:
Added one more commit (to be squashed) that:

1. lets `Invoker.invoke()` rethrow target method exceptions directly, instead of wrapping.
2. Uses `InvokerFactory` instead of `BeanInfo.createInvoker()`.

This should be ready now. I also have the necessary TCK changes ready.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: Got rid of `InvocationTargetException`, `Invoker.invoke()` is now supposed to rethrow target method exceptions directly.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
Ladicek: Introduced `InvokerFactory`.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: Done.

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -0,0 +1,154 @@
...
+Invoker implementations must be thread-safe.
+A single invoker instance may be used to perform multiple, possibly concurrent, invocations of the target method, possibly on different instances of the target bean, possibly with different arguments.
+
+Whenever a direct invocation of a method on an object is a business method invocation, an indirect invocation of that method on that object through an invoker is also a business method invocation.
+
+==== Behavior of `invoke()`
+
+If the target method is `static`, the `instance` is ignored; by convention, it should be `null`.
+If the target method is not `static` and `instance` is `null`, a `RuntimeException` is thrown.
+If the target method is not `static` and the class of the `instance` is not a subclass of the bean class of the target bean, a `RuntimeException` is thrown.
```
Ladicek, line 60: > the class of the `instance` is not a subclass of the bean class of the target bean

We probably need to relax this somehow, because contextual references (and therefore injectable references) don't necessarily need to satisfy this constraint. Not yet sure...

Ladicek: I tried to clarify the requirements on the target instance in the last commit, but it probably needs more thinking.

I was originally convinced it should be enough to demand that the target instance is an instance of the bean class, but there are rules (https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#typecasting_between_bean_types) that make this more complex (I vaguely recall something to do with EJBs too).

Ladicek: I think I figured this out, added one more commit. The gist is:

We already define that the target method must be declared by the bean class of the target bean, or inherited from a supertype. That remains, and failure is still a deployment problem (basically a bean/method mismatch). We however also define that the target method should be declared by a type that is present in the set of bean types of the target bean. If not, it is a non-portable behavior, since some implementations can support it. (As an aside, I also made it non-portable to register an invoker for a `final` method. Again, some implementations can support it.)

This lets us mandate minimum requirements on the target instance in case it is a contextual reference for the target bean (likely the most common case), in line with the rest of the specification. If the target method is declared on an interface, the contextual reference can always be cast to it, otherwise the contextual reference must be for the type that declares the target method. For a non-contextual instance, the class simply must declare the method or inherit from a supertype. Implementations may support more.

Took me a while to figure out, but I feel pretty happy about it.

Ladicek: > As an aside, I also made it non-portable to register an invoker for a `final` method. Again, some implementations can support it.

Added one more commit where I expanded this to all target methods declared on unproxyable bean types.

My reasoning was that if a bean type is unproxyable, it does not need to be possible to obtain a contextual reference for a bean for that bean type, and invoking a method through a contextual reference is likely to be the most common case, so it's better to be open about it.

I however realize that this might be too broad. There are 3 cases when CDI cares about proxyability:

- normal scoped beans
- beans with bound interceptors
- beans with bound decorators

In the 1st case, CDI mandates that all bean types for which a contextual reference is ever obtained are proxyable. In the 2nd and 3rd case, CDI mandates just the _bean class_ to be a proxyable type.

I hope we can tighten the requirement here somehow, but I'm not sure how much.

Ladicek:
Squashed all commits, I believe this is now ready to merge.

I started working on specification text for lookups, which I'll submit as another PR.

Azquelt:
It looks good to me!

I had a couple of questions:

1. Am I correct in my reading that there's no requirement that the method belongs to any of the bean types of the bean, as long as it is non-private and declared on the bean class or one of its supertypes?
    - However, if a contextual reference is passed to `invoke`, then the method must be on one of the bean types
2. I don't see any _technical_ reason that method invokers couldn't be extended to beans other than managed beans. If we wanted to extend this functionality to other beans in a future release, I think we could do that without any breaking changes?
    - some parts of the spec would have to be loosened (e.g. requiring that the method is _either_ on the bean class of a managed bean or on one of the bean types) and some parts would not apply to non-managed beans (e.g. invoking methods on non-contextual instances)
    - at the moment I don't see any need to do this, I'm just trying to think through whether we've painted ourselves into a corner if someone were to come up with a good use-case in the future.

Azquelt (approved):

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/InvokerFactory.java
```diff
@@ -0,0 +1,24 @@
...
+
+import jakarta.enterprise.invoke.InvokerBuilder;
+import jakarta.enterprise.lang.model.declarations.MethodInfo;
+
+/**
+ * Factory for {@link InvokerBuilder}s.
+ *
+ * @since 4.1
+ */
+public interface InvokerFactory {
```
manovotn, line 11: I have to say I am not a big fan of this.
You still need `BeanInfo` as an argument, so this just adds an extra step along the way where your `@Registration` method needs an extra argument.
Technically, you could also attempt to pass in some custom impl of `BeanInfo` which I don't suppose we want users to try.

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -0,0 +1,173 @@
...
+=== Building an `Invoker`
+
+The CDI container allows building an `Invoker` for a method of an <<enablement,enabled>> <<managed_beans,managed bean>>.
+The method for which the invoker is built is called the _target method_ of the invoker, and the managed bean is called the _target bean_ of the invoker.
+
+Invalid target methods are:
+
+* private methods,
+* constructors,
+* methods declared on the `java.lang.Object` class, except of the `toString()` method,
```
manovotn, line 21:
```suggestion
* methods declared on the `java.lang.Object` class, except for the `toString()` method,
```

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -0,0 +1,173 @@
...
+public interface Invoker<T, R> {
+    R invoke(T instance, Object[] arguments) throws Exception;
+}
+----
+
+Calling `invoke()` invokes the target method on given `instance` of the target bean, passing given `arguments`, and propagates back the return value or thrown exception.
+The `instance` and `arguments` may be contextual or non-contextual objects.
+
+Invoker implementations must be thread-safe.
+A single invoker instance may be used to perform multiple, possibly concurrent, invocations of the target method, possibly on different instances of the target bean, possibly with different arguments.
```
manovotn, line 56: I am not sure mentioning concurrent invocations is smart here.
While you can definitely do that, many beans are no built for concurrent usage by default (such as req. scoped beans) yet are a valid candidate for an invoker.  A user cannot know which beans will have invokers and so cannot guarantee concurrency works if attempted.
So, I'd either drop the note or add it into a separate sentence mentioning it's possible yet whether it's safe depends on the respective bean implementation.

manovotn (approved):
Looks good; especially the spec text is pretty impressive in how accurate it is while still remaining readable!
I left some two minor comments and a grumble about `InvokerFactory`, but I understand I got outvoted there :)

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/InvokerFactory.java
Ladicek: Given that a `@Registration` method must declare exactly one parameter of type `BeanInfo` or `InterceptorInfo` or `ObserverInfo`, I considered specifying that an `InvokerFactory` parameter may be declared on a `@Registration` method that declares a parameter of type `BeanInfo` and that `BeanInfo` is the target bean of the invokers created by the factory, and then `InvokerFactory.createInvoker()` would only have a single parameter of type `MethodInfo`. In the end, it felt a little too magical to me.

The main reason why I don't want `BeanInfo.createInvoker()` is that `BeanInfo` (like all other `*Info` types) is a "passive" object that only hands out information. It shouldn't have any "active" methods that configure behavior. 

An alternative would be something like

```java
/**
 * Allows configuring a bean.
 *
 * @see Registration
 * @since 4.1
 */
public interface BeanConfig {
    /**
     * Returns the {@link BeanInfo} corresponding to this configured bean.
     *
     * @return the {@link BeanInfo} corresponding to this configured bean, never {@code null}
     */
    BeanInfo info();

    /**
     * Returns a new {@link InvokerBuilder} for given method of this bean. The builder
     * eventually produces an opaque representation of the generated invoker.
     * <p>
     * If an invoker may not be built for this bean or for given {@code method},
     * an exception is thrown.
     *
     * @param method target method of the invoker, must not be {@code null}
     * @return the invoker builder, never {@code null}
     */
    InvokerBuilder<InvokerInfo> createInvoker(MethodInfo targetMethod);
}
```

I'm a little hesitant about this, because it would also be a natural place for other bean configurations (bean attribute transformations, injection point transformations) and I have doubts we can do that in the `@Registration` phase, as currently specified.

spec/src/main/asciidoc/core/invokers.asciidoc
manovotn: > I could add one more sentence if we wanted to be more clear about that:

:+1: sounds good

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: Done.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: Done.

PR #699, opened 2023-11-09
Specify @Priority on producers

### LINKED ISSUE #556: @Priority on producers
Now that `@Priority` can be put on anything, we could (should?) support `@Priority` on producers (methods? fields? both?).

Ladicek:
Fixes #556

### DIFF
--- a/spec/src/main/asciidoc/core/injectionandresolution.asciidoc
+++ b/spec/src/main/asciidoc/core/injectionandresolution.asciidoc
@@ -45,8 +45,12 @@ An alternative is not available for injection, lookup or name resolution to clas
 An alternative may be given a priority for the application:
 
 * by placing the `@Priority` annotation on the bean class of a managed bean, or
+* by placing the `@Priority` annotation on the producer method, field or resource, or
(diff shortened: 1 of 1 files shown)


Ladicek:
Submitting as a draft because I need to think a little more about this formulation:

> To determine the priority of a producer method or field, the `@Priority` annotation on the producer method or field is considered first.
> If the producer method or field does not declare the `@Priority` annotation, the `@Priority` annotation declared on the bean class is used.

The question is: should this be a general principle, or should this only be taken into account during ambiguity resolution? (The CDI specification doesn't use the priority for anything else, but I know of implementations that _do_ use the priority value of beans e.g. for ordering.)

manovotn (approved):

Emily-Jiang (approved):

manovotn:
> Submitting as a draft because I need to think a little more about this formulation

Can we swap this from draft to ready for review?
I think the current formulation is precise and would work fine.

Ladicek:
Yeah, I totally forgot I updated this PR with a more targeted text.

PR #700, opened 2023-11-13
Add API for programmatic access to assignability rules for observer methods and typesafe resolution.

### LINKED ISSUE #498: Programmatic access to Assignability rules
Some of the important algorithms outlined in pseudo-code in the spec are already available to developers of extensions, e.g. BeanManager#resolve, BeanManager#resolveDecorators and BeanManager#resolveObserverMethods. However, underlying those are other algorithms which are currently not made available to the programmer without depending on specific vendors, the type assignability rules. It is of course possible to implement the algorithms by hand when needed, the spec is clear enough to do that, but the rules are subtle enough that this is error prone and not completely trivial.

Specific use case that I have in mind: I'd like to add to the CDI container's event mechanism by also delivering events to consumers which can be registered and unregistered at runtime. In order to be consistent, I'd like the consumers to conform to the same strict type rules that observer methods have. Unfortunately, there is no quick and simple way to apply these rules except for diving into Weld's source code and stealing what I need hoping that I don't inadvertently break some subtle edge case in the process.

manovotn:
Fixes #498 

There was a request to have access to CDI assignability rules programmatically so this is the first thing coming to my mind.
**I am not 100% convinced that this feature is very useful (in that it can find some audience) so I would appreciate some feedback on whether we want to add it or not.**

Few notes:
* Delegate assignability is deliberately left out as I cannot imagine anyone extending that.
* The varargs argument for qualifiers might instead be a collection because there isn't really a case where this could be entirely left out due to presence of `@Default` - at least not the way it's worded right now as users are expected to list all required qualifiers.
* We could use the `TypesAndQualifiers` to also represent the event payload/qualifiers and injection point type/qualifiers but I somehow found it cleaner this way as the required type would always be a singleton set.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
@@ -251,4 +251,40 @@ public interface BeanContainer {
      * @since 2.0
      */
     Instance<Object> createInstance();
+
+    /**
+     * Returns true if a bean with given bean types and qualifiers would be assignable
+     * to an injection point with given required type and required qualifiers, false otherwise.
+     * <p>
+     * Callers do not need to include implied qualifiers ({@code @Default}, {@code Any}).
+     * These will be automatically added where applicable.
+     * <p>
+     * Throws {@link IllegalArgumentException} if any of the arguments is {@code null}.
+     *
+     * @param beanTypes bean types of a bean; must not be {@code null}
+     * @param beanQualifiers qualifiers of a bean; must not be {@code null}
+     * @param requiredType required type of an injection point; must not be {@code null}
+     * @param requiredQualifiers required qualifiers of an injection point; must not be {@code null}
+     * @return true if a bean with given bean types and qualifiers would be assignable
+     * to an injection point with given required type and required qualifiers, false otherwise
+     */
+    boolean isMatchingBean(Set<Type> beanTypes, Set<Annotation> beanQualifiers, Type requiredType, Set<Annotation> requiredQualifiers);
+
+    /**
+     * Returns true if an event object with given type and qualifiers would match
+     * an observer method with given observed event type and observed event qualifiers, false otherwise.
+     * <p>
+     * Callers do not need to include implied qualifiers ({@code @Default}, {@code Any}).
+     * These will be automatically added where applicable.
+     * <p>
+     * Throws {@link IllegalArgumentException} if any of the arguments is {@code null}.
+     *
+     * @param eventType type of an event object; must not be {@code null}
+     * @param eventQualifiers event qualifiers; must not be {@code null}
+     * @param observedEventType observed event type of an observer method; must not be {@code null}
+     * @param observedEventQualifiers observed event qualifiers on an observer method; must not be {@code null}
+     * @return true if an event object with given type and qualifiers would result in notifying
+     * an observer method with given observed event type and observed event qualifiers, false otherwise
+     */
+    boolean isMatchingEvent(Type eventType, Set<Annotation> eventQualifiers, Type observedEventType, Set<Annotation> observedEventQualifiers);
 }
--- a/spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
+++ b/spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
@@ -260,3 +260,14 @@ Instance<Object> createInstance();
 Instances of dependent scoped beans obtained with this `Instance` object must be explicitly released by calling `Instance.destroy()` method.
 
 If no qualifier is passed to `Instance.select()` method, the `@Default` qualifier is assumed.
+
+==== Assignability of beans and observers
+
+The methods `BeanContainer.isMatchingBean()` and `isMatchingObserver()` provide access to assignability rules defined in <<typesafe_resolution>> and <<observer_resolution>>.
+
+[source, java]
+----
+public boolean isMatchingBean(Set<Type> beanTypes, Set<Annotation> beanQualifiers, Type requiredType, Set<Annotation> requiredQualifiers);
+
+public boolean isMatchingEvent(Type eventType, Set<Annotation> eventQualifiers, Type observedEventType, Set<Annotation> observedEventQualifiers);
+----


api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
```diff
@@ -251,4 +252,40 @@ public interface BeanContainer {
...
+     * Throws {@link IllegalArgumentException} if any of the arguments is {@code null}.
+     *
+     * @param bean Represent the type and qualifiers of a bean
+     * @param injectionPointType Type of injection point
+     * @param qualifiers Injection point qualifiers
+     * @return true if a bean with supplied {@link TypesAndQualifiers} is assignable to an injection point of given
+     * type and qualifiers; false otherwise
+     * @param <T> represents injection point type
+     */
+    <T> boolean isMatchingBean(TypesAndQualifiers bean, T injectionPointType, Annotation... qualifiers);
```
Ladicek, line 290: The type parameter here is practically useless. You can pass a `String` as the parameter and noone will complain.

We could have 2 variants of the method, one taking `Class<T>` and the other taking `TypeLiteral<T>`. Or we could just use `TypeAndQualifiers` for both inputs.

Which brings me to the other idea: why is it `TypesAndQualifiers`? I believe most usages would need exactly 1 type, no?

Ladicek:
I think the current state is decent. I tried to improve it a little here: https://github.com/Ladicek/cdi-spec/commits/programmatic-assignability-access It's mainly about consistency and using terms defined by the CDI specification.

manovotn:
> I think the current state is decent. I tried to improve it a little here: https://github.com/Ladicek/cdi-spec/commits/programmatic-assignability-access It's mainly about consistency and using terms defined by the CDI specification.

I've added them here and, as the commit suggested, squashed them into my commit.

Ladicek:
Thanks!

The only remaining question from me is: should we really expect users to correctly add implied qualifiers (`@Default` and `@Any`)? It feels like CDI implementations should take care of that, like everywhere else.

If I understand correctly, the rules for that are fairly simple. To all set of qualifiers, the CDI implementation would:

- add the `@Any` qualifier, always
- add the `@Default` qualifier if and only if the set of qualifiers contains no other qualifiers but `@Any`, `@Default` and `@Named(...)`

That should be it I think?

manovotn:
> The only remaining question from me is: should we really expect users to correctly add implied qualifiers (@Default and @Any)? It feels like CDI implementations should take care of that, like everywhere else.

Right, that crossed my mind as well.
The reason I did it like this is to preserve the option to input literally any combination and see how CDI works in that case.
Also, I am not sure if it would then allow to cover every case as you might be able to get a bean (or an injection point) with some highly imaginative (read botched) qualifier combinations via extensions - such as deliberately defining such IP or stripping off its qualifiers etc.
Also note that this is slightly different for events as am OM with `@Default` will only get notified for events that have `@Default` or no qualifiers at all (as opposed to having the subset that the OM defines).
That being said, we can surely add that capability.

Ladicek:
Ah yeah, I always forget that the `@Default` qualifiers work differently for observers. Well, I guess we should put together a list of uncommon situations that could occur when we automatically add implied qualifiers...?

Looking at it more thoroughly, here are some additional thoughts.

`@Any`:
- typesafe resolution:
    - beans: all beans have `@Any` (=> we can safely add it to the set of bean qualifiers)
    - injection points: no rules (=> we shouldn't change the set of required qualifiers)
- observer resolution:
    - events: all events have `@Any` (=> we can safely add it to the set of event qualifiers)
    - observer methods: no rules (=> we shouldn't change the set of observed event qualifiers)

`@Default`:
- typesafe resolution:
    - beans: `@Default` is implied when there's no other qualifier other than `@Any` or `@Named(...)`, but can be declared explicitly with other qualifiers too  (=> we can safely add it to the set of bean qualifiers under the same rules)
    - injection points: `@Default` is implied if the set of required qualifiers is empty (=> we can safely add it to the set of required qualifiers if it's empty)
- observer resolution:
    - events: no rules (=> we shouldn't change the set of event qualifiers)
    - observer methods: no rules (=> we shouldn't change the set of observed event qualifiers)

manovotn:
@Ladicek I have reworded the text so that it no longer requires users to specify all qualifiers explicitly.

api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
```diff
@@ -251,4 +251,42 @@ public interface BeanContainer {
      * @since 2.0
      */
     Instance<Object> createInstance();
+
+    /**
+     * Returns true if a bean with given bean types and qualifiers would be assignable
+     * to an injection point with given required type and required qualifiers, false otherwise.
+     * In other words, this method provides access to the typesafe resolution rules - the process of matching a bean to an injection point.
```
Ladicek, line 258: In my previous wording, I tried to _not_ make an impression that this method represents the entirety of typesafe resolution rules, because there's more to it than just assignability. The current wording is a little more ambiguous.

But maybe that's not a big deal and this sentence could be dropped entirely?

(Same for observers, although unfortunately the spec doesn't have a nice single term for what `isMatchingEvent()` does, unlike `isMatchingBean()`, which just implements _assignability_.)

api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
```diff
@@ -251,4 +251,42 @@ public interface BeanContainer {
...
+     * <p>
+     * Callers do not need to include implicit qualifiers such as {@code @Default} or {@code Any}.
```
Ladicek, line 260: I'm no English expert, but "such as" implies that there may be more than just what you list, which AFAIK is not true. How about this?

```suggestion
Callers do not need to include implied qualifiers ({@code @Default}, {@code Any}).
```

Ladicek:
And could you please change the commit message? "Draft an API for ..." -> "Add API for ..."

api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
manovotn: Hmm, I was trying to explain it plainly but you're right that it introduces ambiguity.
Will remove the sentence.

manovotn:
> And could you please change the commit message? "Draft an API for ..." -> "Add API for ..."

All of the comments should be addressed now.

Ladicek:
Ah and we should mention these methods in the spec text, chapter _Programmatic access to container_ (`beanmanager_lite.asciidoc`).

Ladicek:
I think something like this should be enough:

```asciidoc
==== Assignability of beans and observers

The methods `BeanContainer.isMatchingBean()` and `isMatchingObserver()` provide access to assignability rules defined in <<typesafe_resolution>> and <<observer_resolution>>.

[source, java]
----
public boolean isMatchingBean(Set<Type> beanTypes, Set<Annotation> beanQualifiers, Type requiredType, Set<Annotation> requiredQualifiers);

public boolean isMatchingEvent(Type eventType, Set<Annotation> eventQualifiers, Type observedEventType, Set<Annotation> observedEventQualifiers);
----
```

manovotn:
> Commit message, not [just] PR title :-)

Eeh, yea. Apparently it helps to save the changes when you perform them locally :facepalm: 

> Ah and we should mention these methods in the spec text, chapter Programmatic access to container (beanmanager_lite.asciidoc).

That's good point. Will do!

manovotn:
@Ladicek I've added the text to `beanmanager_lite.asciidoc`

Ladicek (approved):
LGTM

manovotn:
Since we agreed to include this feature in during last meeting, I am going to merge it.
There is a tracking issue for TCKs that shouldn't be too hard to do.

PR #707, opened 2023-11-28
Add entry for jdk 21 into Github Actions building matrix 

bmscomp:
The main purpose of this pull request is to add an entry for building the project using JDK 21 with and update of github actions tools for checkout and build project


bmscomp:
@manovotn  Sorry I missed that,  thanks so much

PR #773, opened 2024-02-26
fix license headers and some other small things in the website

Ladicek:


Ladicek:
CC @Azquelt I had to move the license headers into the Jekyll front matter block, as YAML comments, otherwise Jekyll would not notice the front matter at all.

Azquelt (approved):
Thanks, I hadn't noticed that the .html and .md files are not _just_ html and markdown.

manovotn (approved):

PR #787, opened 2024-03-01
Move EE dependency version properties to parent pom

### LINKED ISSUE #786: Update the jakarta.enterprise.cdi-el-api EL version
**Describe the bug**
The cdi/el module has its own uel.api.version property that was not updated for the 4.1.0 release.

**Expected behavior**
This needs to be updated and should be coming from the parent pom so we are only setting this in one place.

**Additional context**
The 4.1.0 release is only staged, so it will have to be dropped and the tag moved.

starksm64:
Fixes #786 for the 4.1 branch

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,17 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Additional context**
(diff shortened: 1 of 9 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,3 +1,6 @@
+<!--
+    SPDX-License-Identifier: Apache-2.0
+-->
```
Ladicek, line 3: I see there's a YAML front matter below. This comment will make it _not_ work. I'd suggest moving the comment into the front matter, using  YAML comments:

```
---
# SPDX-License-Identifier: Apache-2.0

name: Bug report
about: Create a report to help us improve
title: ''
labels: ''
assignees: ''
---

...
```

That's what I did with the Jekyll templates in the `docs` directory.

PR #794, opened 2024-03-07
Create beans_4_1.xsd

starksm64:
Add the 4.1 beans.xml schema to main branch


Ladicek:
I think we need this entire commit: https://github.com/jakartaee/cdi/pull/793/commits/e8611381736c5da804ef9ad0e82c77083f66d182 Otherwise, we'll lose the new paragraph in `preface.adoc`.

starksm64:
I'll do that in another commit as I had the pom.xml version changes that we don't want.

PR #796, opened 2024-03-08
Add a maven file to build a combined javadoc view of the CDI api sources

starksm64:


Ladicek:
I guess we'll eventually also want this in the `main` branch.

PR #817, opened 2024-05-21
Allow registering BCE in CDI SE without discovery

### LINKED ISSUE #813: Add the ability to register BuildCompatibleExtension in CDI SE without discovery
Some time ago I realized that while we added build compatible extensions, we didn't allow users to register them in SE programmatically.
The only option in SE is to discover them which is unfortunate as you often want a purely synthetic bean archive (to keep it minimal).

The `SeContainerInitializer` already allows this for portable extensions, see [this code](https://github.com/jakartaee/cdi/blob/main/api/src/main/java/jakarta/enterprise/inject/se/SeContainerInitializer.java#L200-L207).
We could easily add an extra method with the same capability.

```
    public abstract SeContainerInitializer addBuildCompatibleExtensions(Class<? extends BuildCompatibleExtension>... extensions);
```

Note that I am currently adding the same in Weld under [this issue](https://issues.redhat.com/browse/WELD-2772) but I really think this should be in spec API.

manovotn:
Fixes #813 

Adds a new method to `SeContainerInitializer` which allows users to register build compatible extensions (BCE) without the need for discovery.

I am aware the `SeContainerInitializer` has [two variants of the `add` method for portable extensions (PE)](https://github.com/jakartaee/cdi/blob/4.1.0/api/src/main/java/jakarta/enterprise/inject/se/SeContainerInitializer.java#L192-L207), one of which allows you to pass in pre-initialized PE.
I intentionally skipped that for BCE as those were designed with stricter environments in mind. However, if there is need/want for this we can add it as well.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/se/SeContainerInitializer.java
+++ b/api/src/main/java/jakarta/enterprise/inject/se/SeContainerInitializer.java
@@ -20,6 +20,7 @@ import java.util.Map;
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
 
+import jakarta.enterprise.inject.build.compatible.spi.BuildCompatibleExtension;
 import jakarta.enterprise.inject.spi.Extension;
 
 /**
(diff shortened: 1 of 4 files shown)


Ladicek (approved):

Ladicek:
Apparently the formatter doesn't like the file...?

manovotn:
> Apparently the formatter doesn't like the file...?

Eh, not just that, I merely forgot to push the rebased version which also adapts tests with dummy initializer impls.
Should be good now.

Ladicek (approved):

Ladicek (approved):

manovotn:
I've rebased this against current `main` since it has been hanging here for a while.
Since there was already an approval, I am going to merge this once the CI passes.

PR #818, opened 2024-05-21
Updates

Ladicek:
- move from 5.0.1.Alpha1-SNAPSHOT to 5.0.0.Alpha2-SNAPSHOT
- move the base Java level to 17
- move a few forgotten references from 4.1 to 5.0
- remove deprecated methods from `BeanManager`


manovotn (approved):

manovotn:
> move from 5.0.1.Alpha1-SNAPSHOT to 5.0.0.Alpha2-SNAPSHOT

Shouldn't this be `5.0.0.Alpha1-SNAPSHOT`?

api/src/main/java/jakarta/enterprise/inject/spi/BeanManager.java
```diff
@@ -221,28 +221,6 @@ public interface BeanManager extends BeanContainer {
      */
     public int getInterceptorBindingHashCode(Annotation interceptorBinding);
 
-    /**
-     * Returns a {@link jakarta.el.ELResolver} that resolves beans by EL name.
-     *
-     * @deprecated use {@code ELAwareBeanManager}, this method will be removed in CDI 5.0
-     * @return the {@link jakarta.el.ELResolver}
-     */
-    @Deprecated(since = "4.1", forRemoval = true)
```
manovotn, line 230: Also remove imports related to `ElResolver` from this class.

manovotn (changes requested):
Since you are removing the BM methods, we should be able to also remove the EL dep from CDI API `pom.xml` and only keep it in the CDI EL API.

api/src/main/java/jakarta/enterprise/inject/spi/BeanManager.java
Ladicek: Good point, I forgot about that yesterday.

Ladicek: Done.

Ladicek:
> Shouldn't this be `5.0.0.Alpha1-SNAPSHOT`?

That's what I also thought, but there have apparently been several attempts to release 5.0.0.Alpha1 already. None of them appeared in Maven Central, but the commits are there. Not sure what to do about this...?

manovotn:
> > Shouldn't this be `5.0.0.Alpha1-SNAPSHOT`?
> 
> That's what I also thought, but there have apparently been several attempts to release 5.0.0.Alpha1 already. None of them appeared in Maven Central, but the commits are there. Not sure what to do about this...?

With no release in Central and no tag either, does it matter?

Ladicek:
Good point, I didn't notice there's no tag. I'll move back to `5.0.0.Alpha1-SNAPSHOT`.

Ladicek:
Done.

Azquelt (approved):

PR #828, opened 2024-09-18
Add the TCK challenge template

Emily-Jiang:


Ladicek:
This should be in https://github.com/jakartaee/cdi-tck IMHO.

manovotn:
> This should be in https://github.com/jakartaee/cdi-tck IMHO.

I agree.
Besides, CDI TCKs actually already have a file capturing that - https://github.com/jakartaee/cdi-tck/blob/master/impl/src/main/resources/tck-tests.xml
Implementations are supposed to use that file (with testng) to skip any tests that are assumed to be under a valid challenge.

Emily-Jiang:
okay, good point. Let me move over there.

PR #833, opened 2025-01-13
add SyntheticBeanBuilder.withInjectionPoint()

Ladicek:


Ladicek:
This is a draft, because it's very much not ready. The biggest open question is: does `SyntheticBeanBuilder.withInjectionPoint()` in any way affect the `Instance<Object>` lookup parameter in `SyntheticBeanCreator` / `SyntheticBeanDisposer`?

One way would be to say something like:

> Implementations are allowed to only support looking up beans from {@code lookup} that have previously been registered using {@code withInjectionPoint()}.

That is obviously a breaking change, but I don't see a better way that wouldn't be more breaking.

Ladicek:
CC @manovotn @mkouba

Azquelt:
What are you trying to fix by adding this?

I don't see an explanation of the problem or a linked issue.

mkouba:
> What are you trying to fix by adding this?
> 
> I don't see an explanation of the problem or a linked issue.

In a build-time environment, such as Quarkus, injection point metadata has to be recorded for any `@Dependent` synthetic bean since we don't know if it defines an `InjectionPoint` dependency or not. In other words, the current API prevents some build-time optimizations and enforces unnecessary bytecode generation and reflection.

Ladicek:
What @mkouba says is just the reason why I'm submitting this PR now.

The ultimate cause is feature parity -- Portable Extensions allow this, while Build Compatible Extensions do not.

Ladicek:
> This is a draft, because it's very much not ready. The biggest open question is: does `SyntheticBeanBuilder.withInjectionPoint()` in any way affect the `Instance<Object>` lookup parameter in `SyntheticBeanCreator` / `SyntheticBeanDisposer`?
> 
> One way would be to say something like:
> 
> > Implementations are allowed to only support looking up beans from {@code lookup} that have previously been registered using {@code withInjectionPoint()}.
> 
> That is obviously a breaking change, but I don't see a better way that wouldn't be more breaking.

I eventually decided that this is something we should keep doing in Quarkus, outside of the specification. The spec should stay silent on this.

Azquelt:
Thanks @Ladicek for the explanation on the call. As I understand it now:

With portable extensions, you can create a synthetic bean by implementing [`Bean<T>`][bean-t] which also allows you to implement the `getInjectionPoints` method. The injection points specified here aren't used in the creation of the bean instance, but are validated by the container and are available for inspection and validation by other extensions observing `ProcessBean`.

Other extensions can also modify the injection points by observing `ProcessInjectionPoint`, but any modifications would likely be ignored by the synthetic bean.

As far as I can see, there's no way to link an injection point to an injected dependent bean to allow the dependent bean to inspect where it's been injected.

This PR would add equivalent functionality to build compatible extensions. Injection points can be added to the synthetic bean and will be validated by the container and available for other extensions to inspect with [`@Registration`][registration] and [`BeanInfo`][bean-info].

Additionally, Quarkus want the ability to do some additional out-of-spec optimisations (such as removing any beans which aren't needed), and having this functionality in build compatible extensions would add a way for the user to indicate that certain beans are used by their synthetic bean and can't be optimised away.

[bean-t]: https://jakarta.ee/specifications/platform/10/apidocs/jakarta/enterprise/inject/spi/bean
[registration]: https://jakarta.ee/specifications/platform/10/apidocs/jakarta/enterprise/inject/build/compatible/spi/registration
[bean-info]: https://jakarta.ee/specifications/platform/10/apidocs/jakarta/enterprise/inject/build/compatible/spi/beaninfo

Ladicek:
Actually, let me push this PR back to _draft_, because there's a bit of a conflict we need to resolve: the Portable Extensions API allows this, but expect a direct implementation of `InjectionPoint`, which contains a lot more than just type and qualifiers. And at least in Weld, those extra members are used in validation, for example. So we need to unify those two APIs somehow...

manovotn:
> I found some time and updated the proposal with a somewhat better (I believe) API. The main question remains (and it's why I'm keeping this PR as a draft): how can we bridge this with Portable Extensions, which expect full blown `InjectionPoint`? Need to discuss with @manovotn, because I'm no expert in that area.

Pardon my ignorance, I am pretty sure we discussed this but I don't see it recorder here - what was the issue with introducing `SyntheticInjectionPoint` interface (a parent of `jakarta.enterprise.inject.spi.InjectionPoint`) that would have just getters for type and qualifiers?
I know Weld ATM assumes the other properties of the IP are non-null but that seems mendable.

manovotn:
> I don't think there's an issue with that. (Besides the name. We need a good name.) If you think that's a reasonable way forward, I'll try to figure something out.

It does sound reasonable to me, but we can discuss it further in the mtg today (assuming you are coming).

manovotn:
> For an unexplainable reason, I'm quite fond of the idea of `SyntheticInjectionPoint` _extending_ `InjectionPoint` :-)

Sure, works for me.

Ladicek:
I was thinking it would be quite acceptable to do a small breaking change and say that `InjectionPoint.getMember()` may return `null` in case the bean declaring the injection point is a synthetic bean. But then, I found this: https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#inter_module_injection_full Apparently, the spec relies upon `getMember()` never returning `null`, even in case of synthetic beans.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticInjections.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright (c) 2025 Red Hat and others
```
manovotn: Wrong copyright header?

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticInjections.java
```diff
@@ -0,0 +1,54 @@
...
+ * that were previously registered using {@code SyntheticBeanBuilder.withInjectionPoint()}.
+ *
+ * @since 5.0
+ */
+public interface SyntheticInjections {
+    /**
+     * Returns an injectable reference to a bean of given {@code type} with given {@code qualifiers},
+     * assuming an injection point with the same type and qualifiers was registered
+     * on the {@link SyntheticBeanBuilder}. Returns {@code null} if the type/qualifiers
+     * combination was not registered on the {@code SyntheticBeanBuilder}.
```
manovotn: Shouldn't it be an error to try and retrieve a non-registered bean?
Since it is just one party registering and then using the synth. injection point, they should know what to look for and a wrong input should give them some meaningful error.

Besides, there is always the edge case of a [dependent scoped producer returning `null`](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#producer_method_lifecycle) as a bean which is then indistinguishable from not having any such bean.

Or we could return an `Optional` which would be empty if there was no bean.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
```diff
@@ -435,6 +435,66 @@ public interface SyntheticBeanBuilder<T> {
...
+     * <p>
+     * The registered injection point will be validated. If no matching bean exists,
+     * the container treats it as a deployment problem.
+     *
+     * @param type the type of the bean that may be looked up in the creation/destruction function
+     * @param qualifiers the qualifiers of the bean that may be looked up in the creation/destruction function
+     * @return this {@code SyntheticBeanBuilder}
+     * @since 5.0
+     */
+    SyntheticBeanBuilder<T> withInjectionPoint(Type type, Annotation... qualifiers);
```
manovotn, line 513: A `Type` will allow arbitrary input including those that are not legal bean types.
I was thinking whether we could use `TypeLiteral` instead as that one is at least documented as meant for instantiation of param types with actual params (which I assume is what you added `Type` here for?).

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticInjections.java
Ladicek: Indeed. I'll fix.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticInjections.java
Ladicek: Good point. We could throw `IllegalArgumentException ` I guess.

Ladicek: Aside: I think I specified this to behave the same as the corresponding API we have in ArC, but I think throwing an exception is a better idea.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
Ladicek: I don't think we would want to use `TypeLiteral` here, because the only thing it provides is a `java.lang.reflect.Type` and we would have to construct the build-time representation of types from that. It feels better to have the user construct the build-time representation of the type directly.

manovotn: Hm, you're right

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticInjections.java
mkouba: Are you talking about `io.quarkus.arc.SyntheticCreationalContext.getInjectedReference(Class<R>, Annotation...)`? If so, then we also throw an `IllegalArgumentException` if a corresponding synthetic injection point was not defined...

Ladicek: Do we? Then I have no idea why I wrote this the way I wrote it :-)

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticInjections.java
Ladicek: Fixed.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticInjections.java
Ladicek: Fixed.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanCreator.java
```diff
@@ -17,11 +17,45 @@
  * CDI container will create an instance of the creation function every time when it needs
  * to obtain an instance of the synthetic bean. Implementations must be {@code public}
  * classes with a {@code public} zero-parameter constructor; they must not be beans.
+ * <p>
+ * Starting with CDI 5.0, the {@link #create(Instance, Parameters)} method is deprecated
+ * for removal; suggested replacement is {@link #create(SyntheticInjections, Parameters)}.
+ * Exactly one of these methods must be implemented. The CDI container must check which method
```
manovotn: Should we explicitly state what kind of error is expected if none/both methods are implemented?

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticInjections.java
```diff
@@ -0,0 +1,56 @@
...
+     * Returns an injectable reference to a bean of given {@code type} with given {@code qualifiers},
+     * assuming an injection point with the same type and qualifiers was registered
+     * on the {@link SyntheticBeanBuilder}. Throws {@link IllegalArgumentException}
+     * if no such injection point was registered.
+     * <p>
+     * If no qualifier is passed, {@code @Default} is assumed.
+     *
+     * @param type the type of the bean to lookup, must not be {@code null}
+     * @param qualifiers the qualifiers of the bean to lookup, may be {@code null} or empty
+     * @return injectable reference to the bean, never {@code null}
```
manovotn: Soo, I will be a little evil and question the `never null` part by pointing out that dependent scoped beans *can* return `null` 😇

Ladicek: Good point, will fix.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanCreator.java
Ladicek: Non-portable behavior I guess?

Ladicek: Done, this is specified to be non-portable now.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticInjections.java
Ladicek: Done.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanCreator.java
```diff
@@ -17,11 +17,46 @@
...
+    /**
+     * Creates an instance of the synthetic bean.
+     * May only return {@code null} if the synthetic bean is {@code @Dependent}.
+     * <p>
+     * The {@link SyntheticInjections} parameter may be used to simulate producer method parameter injection.
+     * All injectable references looked up from {@code SyntheticInjections} have to previously be registered using
+     * {@code SyntheticBeanBuilder.withInjectionPoint()}. All {@code @Dependent} bean instances obtained from
+     * {@code SyntheticInjections} during execution remain managed until the synthetic bean instance is destroyed.
+     * Therefore, implementations are encouraged to destroy unneeded {@code @Dependent} bean instances obtained
+     * from {@code SyntheticInjections}.
```
Azquelt: How do implementations destroy unneeded `@Dependent` instances? Should `SyntheticInjections` have a `destroy` method?

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticInjections.java
```diff
@@ -0,0 +1,56 @@
...
+import jakarta.enterprise.util.TypeLiteral;
+
+/**
+ * Contains injectable references for injection points registered for a synthetic bean.
+ * A synthetic bean creation/destruction function can look up beans in this container
+ * that were previously registered using {@code SyntheticBeanBuilder.withInjectionPoint()}.
+ *
+ * @since 5.0
+ */
+public interface SyntheticInjections {
```
Azquelt, line 27: "Contains injectable references" suggests that the injectable references for injection points are always created prior to bean creation, which would in turn suggest that any `@Dependent` scoped beans are always created, even if they're never retrieved by the user.

Is this actually the case?

If not, I suggest changing the wording to "Provides injectable references".

Ladicek: We don't really specify whether the injectable references are created eagerly or lazily, and I think that's on purpose. I agree that "provides" is a better verb than "contains", for sure.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanCreator.java
Ladicek: Great point. Let me think about that...

Ladicek: I pushed a small commit with my proposal. I removed this sentence, because the set of beans provided by `SyntheticInjections` is bounded (pre-defined by `SyntheticBeanBuilder`) and so we don't really have to care that much. The `SyntheticInjections` javadoc now says:

> All `@Dependent` bean instances created by `SyntheticInjections` are destroyed when the corresponding synthetic bean instance is destroyed.

PR #874, opened 2025-07-08
specification of standbys

### LINKED ISSUE #871: Introduce a concept of a "default" bean
[I think we touched on this topic on several occasions already but I didn't find an issue at a glance]

There is a common scenario where a library wants to allow users to define their own bean but in case no such bean is present, the library wants to add its own "default" bean.

In order to achieve that, such library always needs to have an extension, monitor for given bean and conditionally register its own. There can also be limitations to how the user can register the bean as to avoid double definition if both beans are defined in the ABD phase for instance (at least in Weld, see https://github.com/weld/core/discussions/3191).

However, Quarkus introduced a more elegant solution some time ago - [a concept of "default beans".](https://quarkus.io/guides/cdi-reference#default_beans) and I think we should try to introduce a similar concept to CDI specification itself.

Please bear in mind that the name "default" is subject to change but for the sake of description, I will stick to it here.

A default bean changes the typesafe resolution algorithm [the same way alternatives do](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#unsatisfied_and_ambig_dependencies), but one step sooner. If no other bean than the default one is present, all is good. In case of ambig. dependency (in terms of the presented use case this is equivalent to user defining their own bean), we first eliminate all "default" beans and try again. Should that fail, we use the current algorithm that operates on alternatives (and priority).
If we choose to enable priority on default beans as well, it would take one more step, but nothing we couldn't manage.

An addition such as this would make it much easier for library authors and potentially remove a lot of boilerplate from their extensions. Thoughts?

Ladicek:
Standbys are the opposite of alternatives: where alternatives take precedence over non-alternatives, non-standbys take precedence over standbys. This pattern has classically been known as "default beans", but that is a bad name (consists of two words, both of them being used in the specification already).

In this commit, standbys can only be selected for application, they cannot be selected just for a bean archive (in `beans.xml`). Perhaps this will have to change, but present state should be good enough for now. Similarly, standby producers currently do not inherit priority from their declaring beans, which perhaps will have to change (for the sake of consistency), but present state should be good enough for now.

Resolves #871


ljnelson:
Namingwise, "standby" really grates, not just because it's clumsy, but mostly because it is an adjective that is used here as a noun, where its counterparts, "stereotype" and "alternative", are both nouns.

I understand why we can't use "default" and I understand that names are hard. Nevertheless, getting them right in specifications is really important.

Maybe "backstop" (it's a noun, like alternative and stereotype; among its definitions is "something that bolsters")? "Exemplar" ("one that is typical or representative"; mild yuck)?

api/src/main/java/jakarta/enterprise/inject/Standby.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2025, Red Hat, Inc., and individual contributors
```
manovotn: Nitpicking but this is no longer true :)

manovotn:
> Namingwise, "standby" really grates, not just because it's clumsy, but mostly because it is an adjective that is used here as a noun, where its counterparts, "stereotype" and "alternative", are both nouns.
> 
> I understand why we can't use "default" and I understand that names are hard. Nevertheless, getting them right in specifications is really important.
> 
> Maybe "backstop" (it's a noun, like alternative and stereotype; among its definitions is "something that bolsters")? "Exemplar" ("one that is typical or representative"; mild yuck)? "Template" (not quite right, overloaded meaning, but captures the "starting point" nature of the issue)?

I am aware we have `@Default` already, but I still find using `@DefaultBean` much more self explanatory than any of the options discussed so far. Simply because every developer knows what "default" means instantly as we are all aware of default configs and/or other uses which are similar in their nature.

As for proposed names - "Template" I agree is too overloaded to convey clear meaning and "Exemplar" is too close to saying its an "example bean" which is IMO not the case.
"Backstop" is an interesting one - on that note I am also thinking of "@BackupBean` or perhaps `@Backfill`.
There is also `@FallbackBean` which makes sense in CDI sense but clashes with MP FT perception of it.

Ladicek:
If we didn't have `@Fallback` in MicroProfile Fault Tolerance, I would have called this "fallback" in an instant. "Backup" is quite overloaded, but "backfill" seems nice.

api/src/main/java/jakarta/enterprise/inject/Standby.java
```diff
@@ -0,0 +1,57 @@
...
+import jakarta.enterprise.util.AnnotationLiteral;
+
+/**
+ * <p>
+ * Specifies that a bean is a standby. May be applied to a bean class, producer method or field or
+ * {@linkplain Stereotype stereotype}.
+ * </p>
+ *
+ * <pre>
+ * &#064;Standby
```
manovotn: Unless `@Standby` is meant to be bean defining, the example should also include a scope.
Furthermore, in the spec text you state that the bean needs to be explicitly selected - so a priority declaration should be added as well.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
```diff
@@ -121,14 +121,31 @@ public interface SyntheticBeanBuilder<T> {
...
 
     /**
-     * Sets a priority of this synthetic bean. To make this synthetic bean an enabled alternative,
-     * call both {@code alternative(true)} and {@code priority(some priority)}.
+     * Marks this synthetic bean as a standby if desired. To make this synthetic bean
+     * an enabled standby, call both {@code standby(true)} and {@code priority(some priority)}.
      * <p>
-     * If this synthetic bean is an alternative, not setting a priority means
+     * If this synthetic bean is a standby, not setting a priority means
+     * that it is not enabled, which is equivalent to not registering it at all.
```
manovotn: > ...which is equivalent to not registering it at all.

This doesn't seem very useful, should we detect it and consider it an error is the synth. bean is standby and does not have priority?

spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -721,6 +723,32 @@ Alternatively, an alternative may be declared by annotating a bean, producer met
 
 If an interceptor is an alternative, non-portable behavior results.
 
+[[standbys]]
+
+=== Standbys
+
+A _standby_ is a bean that must be explicitly selected if it should be available for lookup, injection or name resolution.
```
manovotn: Having to explicitly enable the beans at all time seems like a hassle we can avoid - I'd state that if no priority is present, `@Priority(0)` (or other such number) is assumed.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -58,15 +60,34 @@ An alternative may be given a priority for the application:
 * by placing the `@Priority` annotation on the bean class that declares the producer method, field or resource, or
 * by placing the `@Priority` annotation on a stereotype that is applied to the bean class, producer method or producer field.
 
+[[declaring_selected_standbys]]
+
+==== Declaring selected standbys
+
+{cdi_lite} defines one method of selecting standbys: the `@Priority` annotation allows a standby to be selected for an entire application.
```
manovotn: I'd generalize this to CDI, not just Lite as we are not defining any other means of enablement anyway.

spec/src/main/asciidoc/core/spi_full.asciidoc
```diff
@@ -974,6 +977,8 @@ public interface AfterTypeDiscovery {
 
 * `getAlternatives()` returns the ordered list of enabled alternatives for the application, sorted by priority in ascending order.
 Alternatives enabled for a bean archive are not included in the list.
+* `getStandbys()` returns the ordered list of enabled standbys for the application, sorted by priority in ascending order.
+Standbys enabled for a bean archive are not included in the list.
```
manovotn: Since we are not allowing per-archive enablement, I think this sentence can be removed, right?

spec/src/main/asciidoc/javase/javase.asciidoc
```diff
@@ -7,13 +7,10 @@ https://www.apache.org/licenses/LICENSE-2.0.
 
 SPDX-License-Identifier: Apache-2.0
 ////
-[partintro]
```
manovotn, line 10: These small adjustments (this and the EL/bean name) might be worth sending as a separate tiny change as they are unrelated to the PR anyway.

manovotn (commented):
Thank you for the draft @Ladicek!

api/src/main/java/jakarta/enterprise/inject/Standby.java
Ladicek: Ah, I didn't even bother reading the full line, just updated the year :-) Will fix.

api/src/main/java/jakarta/enterprise/inject/Standby.java
Ladicek: This is mostly a copy of the javadoc of `Alternative`. Will look again and update both.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
Ladicek: This is mostyl a copy of the text we have for `alternative()`. I think we could safely remove the "which is equivalent to not registering it at all" part, but I would be wary of additional validations.

spec/src/main/asciidoc/core/definition.asciidoc
Ladicek: I would prefer we didn't do this, because it would be yet another difference to alternatives.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
Ladicek: Good point, will fix.

spec/src/main/asciidoc/core/spi_full.asciidoc
Ladicek: Very true, thanks! Will remove.

spec/src/main/asciidoc/javase/javase.asciidoc
Ladicek: Yeah, good point. I was lazy :-) Will separate them.

api/src/main/java/jakarta/enterprise/inject/Standby.java
Ladicek: Done.

api/src/main/java/jakarta/enterprise/inject/Standby.java
Ladicek: Done.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
Ladicek: Done.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
Ladicek: Done.

spec/src/main/asciidoc/core/spi_full.asciidoc
Ladicek: Done.

spec/src/main/asciidoc/core/definition.asciidoc
manovotn: The reason I like having some default priority is that this is bound to be a user trap - many people will first try to define a standby bean without adding the priority which will give them a needless pause as to what's going on in their app.

This IMO goes hand in hand with the difference already in place - you cannot have this for alternatives because those can be enabled per-archive so having a non-enabled alternative makes sense. That's not true for a standby bean though. The only case I can think of in which you might use non-enabled standby bean is to then conditionally enable such bean in an extension - and you can achieve the same by conditionally vetoing the bean instead :shrug:

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
manovotn: This is also linked to my other comment with default priority - if we had it, then this sentence is redundant.
However, if we decide to not have it, I agree the additional validation might be too much.

spec/src/main/asciidoc/javase/javase.asciidoc
Ladicek: Done. The change "EL name" -> "bean name" is submitted in #875. The removal of `[partintro]` is gone, because it actually is OK, even though my IDE shows an error.

spec/src/main/asciidoc/core/definition.asciidoc
Ladicek: I agree, but if we define that a standby has by default a priority, it is always enabled for application, which prevents us from ever adding per bean archive enablement. I don't think I'm ready to make that call just yet.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
Ladicek: ("Done" above means "I removed the _which is equivalent to not registering it at all_ part", FTR.)

api/src/main/java/jakarta/enterprise/inject/Standby.java
```diff
@@ -0,0 +1,65 @@
...
+ * &#064;Priority(1)
+ * &#064;Dependent
+ * public class DefaultOrder extends Order { ... }
+ * </pre>
+ *
+ * <p>
+ * A standby is not available for injection, lookup or name resolution in a module unless the module
+ * is a bean archive and the standby is explicitly <em>selected</em> for the application (by adding
+ * the {@link Priority} annotation). An alternative is never available for injection, lookup
```
Azquelt:
```suggestion
 * the {@link Priority} annotation). A standby is never available for injection, lookup
```

Ladicek: Agh! Good point, will fix, thanks.

Ladicek: Done.

spec/src/main/asciidoc/core/definition.asciidoc
manovotn: We have touched on this during the CDI meeting on Jul 29 and agreed that we want to keep the current wording so that we have more options in the future.

ljnelson:
I'll keep thinking. Another problem I have with "standby" is that it sounds like something that could be energized or activated or enabled or switched on at any point for any reason, which isn't really what this is. This is a default—something that is used when other options don't exist or can't work—but we just can't use that word.

Another requirement the name has, if we're trying to keep the spirit of the specification, is something that answers the question, "What kind of bean is this?" (A _managed_ bean, a _producer_ bean, an _alternative_ bean, a _synthetic_ bean….) So I agree with you that in general `XYZBean`, as some kind of descriptor, if it can be avoided, should be avoided.

manovotn:
> What about "failsafe" (as a noun)? Something that keeps the system from breaking when (in this case) nobody else supplied The Thing That Was Supposed To Be There? Then you can talk about failsafes and alternatives in the same sorts of sentences.

Failsafe is way too close to fallback (fault tolerance) IMO.
Also, the alternatives are slightly different in behavior and we should highlight that difference if possible so as to avoid confusion.

BTW is there a reason why you don't like `@Backfill`? I am not a native speaker so just curious if there is something wrong with it.

ljnelson:
> Failsafe is way too close to fallback (fault tolerance) IMO

OK; I think we agree to disagree on this particular one.

> BTW is there a reason why you don't like @Backfill? I am not a native speaker so just curious if there is something wrong with it.

Understood. Backfill has connotations of uselessness. It is the stuff—usually material unsuitable for any other purpose—you put back in a hole you just dug.

ljnelson:
Right; I understand. You asked if there was something wrong with it and the noun version is definitely problematic and its bad connotations would occur to native speakers relatively quickly. Put as slang: it leaves a slightly bad taste in the mouth.

ljnelson:
> It should really be a "default bean that users can override if need be"

Two words that fit that definition fairly closely are "prototype" and "archetype". ("Archetype" feels more like "example", but something that is a "prototype" can be a thing that makes copies of itself (not how it would be used here), or a default that could be used or extended.

A prototype is also something that typically could benefit from some improvement (i.e. users could supply their own) but could be relied on as-is.

Ladicek:
I'd personally not use _prototype_, because it is a Spring scope that is very similar to (but not exactly the same as) `@Dependent`. (I actually have an implementation of the prototype scope in CDI in the ArC test suite. It's quite simple.)

Overall, it is my opinion that the best words (_default_ or _fallback_) cannot be used, so we gotta settle with something slightly awkward. To me, _standby_ or _backfill_ sound reasonable, as well as _spare_ or _reserve_. I might also be fine with _last resort_, but with the ability to use priorities, certain "last resorts" wouldn't really be "last" :-)

Emily-Jiang:
> I'd personally not use _prototype_, because it is a Spring scope that is very similar to (but not exactly the same as) `@Dependent`. (I actually have an implementation of the prototype scope in CDI in the ArC test suite. It's quite simple.)
> 
> Overall, it is my opinion that the best words (_default_ or _fallback_) cannot be used, so we gotta settle with something slightly awkward. To me, _standby_ or _backfill_ sound reasonable, as well as _spare_ or _reserve_. I might also be fine with _last resort_, but with the ability to use priorities, certain "last resorts" wouldn't really be "last" :-)

How about `@DefaultBean` or `FallbackBean`?

Emily-Jiang:
> I would prefer a single word that goes well with "bean". If you have to add "bean" to the term itself, you declare defeat :-)

How about `Backup`?

Azquelt:
> To me, _standby_ or _backfill_ sound reasonable, as well as _spare_ or _reserve_.

I really like "Reserve" - something kept back unless it is needed. In some ways it's better than "Default" because having more than one reserve makes sense whereas having more than one default doesn't.

The typesafe resolution steps would then be:
- first look for the highest priority enabled alternative bean
- if there are none, look for a normal bean
- if there are none, look for the highest priority reserve bean

Reading that back, it makes intuitive sense to me.

I'm ok with "Standby" but it has a bit more of a connotation of something ready and waiting to take over a responsibility if something goes wrong, which isn't quite how this feature would work.

ljnelson:
Ooh, I very much like reserve. Certainly much, much better than standby. Well done.

Ladicek:
@manovotn any thoughts about _reserve_? :-)

api/src/main/java/jakarta/enterprise/inject/Alternative.java
```diff
@@ -33,20 +34,23 @@
...
  * </pre>
  *
  * <p>
- * An alternative is not available for injection, lookup or EL resolution to classes or JSP/JSF pages in a module unless the
- * module is a bean archive and the alternative is explicitly <em>selected</em> in that bean archive. An alternative is never
- * available for injection, lookup or EL resolution in a module that is not a bean archive.
+ * An alternative is not available for injection, lookup or name resolution in a module unless the module
+ * is a bean archive and the alternative is explicitly <em>selected</em> for the application (by adding
+ * the {@link Priority} annotation) or for that bean archive. An alternative is never available for injection, lookup
+ * or name resolution in a module that is not a bean archive.
```
Azquelt: > An alternative is never available for injection, lookup or name resolution in a module that is not a bean archive.

Is this still true? I assumed that was related to it needing to be selected using beans.xml, which would need it to be a bean archive. If you enable it for the whole application, I'm unsure whether it should be injectable or look-up-able in modules which are not bean archives.

Not really related to the substance of this PR, so we could defer discussing it.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
```diff
@@ -122,6 +123,13 @@ public interface BeanInfo {
      */
     boolean isAlternative();
 
+    /**
+     * Returns whether this bean is a {@linkplain Reserve reserve}.
+     *
+     * @return whether this bean is a {@linkplain Reserve reserve}
+     */
+    boolean isReserve();
```
Azquelt, line 130: Should we mention reserves in the doc for `priority` as well?

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
```diff
@@ -48,6 +49,14 @@ public interface StereotypeInfo {
...
+     * Returns whether this stereotype is meta-annotated {@link Reserve @Reserve}.
+     * This means that all beans with this stereotype are reserves.
+     *
+     * @return whether this stereotype is meta-annotated {@link Reserve @Reserve}
+     */
+    boolean isReserve();
+
     /**
      * Returns the priority value this stereotype declares using the {@link jakarta.annotation.Priority @Priority}
      * meta-annotation. Beans with this stereotype will be enabled for the application and ordered using this priority
```
Azquelt: The doc here for `priority` mentions enabling beans for the application. That's kind of true because stereotypes are bean defining annotations, but should this bit of doc be talking about enabling alternatives and reserves instead?

api/src/main/java/jakarta/enterprise/inject/spi/configurator/BeanAttributesConfigurator.java
```diff
@@ -199,4 +199,14 @@ public interface BeanAttributesConfigurator<T> {
      */
     BeanAttributesConfigurator<T> alternative(boolean value);
 
+    /**
+     *
+     * Change the reserve status of the configured bean.
```
Azquelt, line 188: Stray blank line

spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -705,6 +704,8 @@ If an interceptor has a name, non-portable behavior results.
 
 An _alternative_ is a bean that must be explicitly selected if it should be available for lookup, injection or name resolution.
 
+Alternatives take precedence over other, non-alternative beans.
+
```
Azquelt: I think this line is helpful, but:
- should we say this is during typesafe resolution?
- should we make it a NOTE or an INFO, to indicate that this is just useful information, the actual specification for this part is elsewhere.

Something like
```asciidoc
INFO: During <<typesafe_resolution>>, alternatives take precedence over other, non-alternative beans.
```

spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -721,6 +722,32 @@ Alternatively, an alternative may be declared by annotating a bean, producer met
 
 If an interceptor is an alternative, non-portable behavior results.
 
+[[reserves]]
+
+=== Reserves
+
+A _reserve_ is a bean that must be explicitly selected if it should be available for lookup, injection or name resolution.
+
+Other, non-reserve beans take precedence over reserves.
```
Azquelt: As above,

```
INFO: During <<typesafe_resolution>>, other non-reserve beans take precedence over reserves.
```

spec/src/main/asciidoc/core/definition.asciidoc
```diff
@@ -867,6 +893,26 @@ Following sample shows a stereotype that can be used to mark bean as globally en
 public @interface Mock {}
 ----
 
+[[reserve_stereotype]]
+
+===== Declaring a `@Reserve` stereotype
+
+A stereotype may declare a `@Reserve` annotation, which specifies that every bean with the stereotype is a reserve.
+
+We may specify that all default impl objects are reserves:
```
Azquelt:
```suggestion
For example, we may specify that all beans annotated with `@DefaultImpl` are reserves:
```

I had a momentary confusion as to what a "default impl" object was. It's less clear than the "mock" example for alternatives.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -116,21 +138,25 @@ A bean is eligible for injection to a certain injection point if:
...
-The container eliminates all eligible beans that are not alternatives, except for producer methods and fields of beans that are alternatives.
-If:
-
-* there is exactly one bean remaining, the container will select this bean, and the ambiguous dependency is called resolvable.
-* all the beans left are alternatives with a priority, or producer methods or fields of beans that are alternatives with a priority, then the container will determine the highest priority value and eliminate all beans, except for:
+When an ambiguous dependency exists, the container attempts to resolve the ambiguity:
+
+. The container eliminates all eligible beans that are reserves.
+. If there is exactly one bean remaining, the container selects this bean, and the ambiguous dependency is called resolvable.
+. Otherwise, if there is no bean remaining and all eligible beans are reserves, then:
```
Azquelt: It looks like we're eliminating all reserves in step 1 but then expecting to still find reserves in step 3?

I'm not sure whether this is the best way of fixing this:

```suggestion
. If any eligible beans are not reserves, the container eliminates all eligible beans that are reserves.
. If there is exactly one bean remaining, the container selects this bean, and the ambiguous dependency is called resolvable.
. Otherwise, if there is no bean remaining and all eligible beans are reserves, then:
```

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -284,22 +310,35 @@ For a custom implementation of the `Bean` interface defined in <<bean>>, the con
...
-If:
 
-* there is exactly one bean remaining, the container will select this bean, and the ambiguous dependency is called _resolvable_.
-* all the beans left are alternatives with a priority, then the container will determine the highest priority value, and eliminate all beans, except for producer methods and fields of beans that are alternatives with the highest priority value.
-If there is exactly one bean remaining, the container will select this bean, and the ambiguous dependency is called _resolvable_.
+When an ambiguous name exists, the container attempts to resolve the ambiguity:
+
+. The container eliminates all eligible beans that are reserves.
+. If there is exactly one bean remaining, the container selects this bean, and the ambiguous name is called resolvable.
+. Otherwise, if there is no bean remaining and all eligible beans are reserves, then:
```
Azquelt: Same comment as above, I find it confusing to eliminate all reserves first and then go back to them.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -284,22 +310,35 @@ For a custom implementation of the `Bean` interface defined in <<bean>>, the con
...
+For the purpose of determining the priority of any producer method or field during step 6. a. above, the priority of the producer method or field is considered first.
+If the producer method or field does not have a priority, the priority of the managed bean that declares the producer method or field is used.
 
 All unresolvable ambiguous names are detected by the container when the application is initialized.
 Suppose two beans are both available for injection in a certain module, and either:
 
-* the two beans have the same bean name and the name is not resolvable, or
-* the bean name of one bean is of the form `x.y`, where `y` is a valid bean name, and `x` is the bean name of the other bean,
+* the two beans have the same bean name and the ambiguous name is not resolvable, or
+* the bean name of one bean is of the form `x.y`, where `y` is a valid bean name, and `x` is the bean name of the other bean.
```
Azquelt: I think this section was clearer without the word ambiguous here.

We're specifying what it means for a name to be ambiguous, so having the word "ambiguous" as part of the description is confusing. The first half of the line makes it clear that we're talking about two beans having the same name.

spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
```diff
@@ -149,17 +149,6 @@ In addition to rules defined in <<assignable_parameters>>, the following rules a
...
-
-[[ambig_names_full]]
-
-==== Ambiguous names in {cdi_full}
-
-In addition to rules defined in <<ambig_names>>, the following rules apply.
-
-When an ambiguous name exists, the container attempts to resolve the ambiguity.
-The container eliminates all eligible beans that are not alternatives selected for the bean archive or selected for the application, except for producer methods and fields of beans that are alternatives.
-
```
Azquelt, line 162: I don't like the way this is specified but I'm not sure we can just get rid of it.

With your changes, I think:
- Alternatives are enabled if they're selected for _any_ bean archive
- Enabled beans are eligible for injection

I think it's only this part that would prevent typesafe resolution from resolving an alternative in a bean archive that it isn't selected for.

api/src/main/java/jakarta/enterprise/inject/Alternative.java
Ladicek: I'm not even sure why this sentence is there, it's not really related to alternatives. But I think it's still true; if an archive is not a bean archive, _nothing_ is available for injection, lookup or name resolution. Or I'm seriously mistaken :-)

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
Ladicek: Absolutely! Will do.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
Ladicek: It could, but then it would also have to mention interceptors.

Since this sentence is kinda superfluous (this is specified elsewhere), we could also remove it altogether.

api/src/main/java/jakarta/enterprise/inject/spi/configurator/BeanAttributesConfigurator.java
Ladicek: It's there in almost all methods in this class, so I felt it's appropriate here too :-) But you're right, it's wrong and needs to be removed in the entire class. Will do.

spec/src/main/asciidoc/core/definition.asciidoc
Ladicek: I agree with prefixing the sentence with "During <<typesafe_resolution>>", but not sure about making it an admonition. We almost never use them in the CDI spec and it doesn't seem like this place deserves one.

spec/src/main/asciidoc/core/definition.asciidoc
Ladicek: Yeah, I unfortunately wasn't able to come up with a better example. Will use your wording.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
Ladicek: I structured the algorithm into a numbered list, but other than that, I tried to not rewrite it too much. This is how it was originally written, and while it might seem weird, it is actually exactly how we implement the algorithm in ArC (not sure about Weld).

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
Ladicek: I actually think this whole paragraph could use a rewrite. The sentence above already says that

> All unresolvable ambiguous names are detected by the container when the application is initialized.

So we don't have to repeat it here.

How about I rewrite this to the following?

> All unresolvable ambiguous names are detected by the container when the application is initialized and treated as a deployment problem.
>
> If there are two beans that are both available for injection in a certain module, and the bean name of one bean is of the form `x.y`, where `y` is a valid bean name, and `x` is the bean name of the other bean, then the container treats it as a deployment problem.

spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
Ladicek: Yeah, so I was thinking about this for a while and concluded that we don't need this because:

- the ambiguity resolution algorithm works on _eligible beans_
- _eligible beans_ are, quote:

  > A bean is eligible for injection to a certain injection point if:
  >
  > * it is available for injection in the module that contains the class that declares the injection point, and
  > * it is assignable to the injection point (using <<assignable_parameters>>).
- when is a bean _available for injection_ is defined in the _Modularity_ chapter (specifically the _Inter-module injection_ section) for CDI Lite, and for CDI Full, there's an additional provision in _Modularity in CDI Full_ that states, quote:

  > An alternative is not available for injection, lookup or name resolution to classes in a module unless the module is a bean archive and the alternative is explicitly _selected_ for the bean archive or the application.

Further, since you're specifically pointing out ambiguous names, the _Name resolution_ section says:

> A name resolves to a bean if:
>
> * the bean has the given bean name, and
> * the bean is available for injection in the module where the name resolution is requested.

Once again, the additional provision in _Modularity in CDI Full_ kicks in: the alternative is not available for injection in the module in which it is not selected (unless it's selected for application), so the ambiguity resolution algorithm should not even bother with considering it.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
Ladicek: Done.

api/src/main/java/jakarta/enterprise/inject/spi/configurator/BeanAttributesConfigurator.java
Ladicek: Done.

spec/src/main/asciidoc/core/definition.asciidoc
Ladicek: Done.

spec/src/main/asciidoc/core/definition.asciidoc
Ladicek: Done.

spec/src/main/asciidoc/core/definition.asciidoc
Ladicek: Done.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
Ladicek: Done.

spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
Ladicek: To be frank, the concept of eligible beans makes no sense for name resolution, because beans are only eligible for injection. The previous text of name ambiguity resolution used the term eligible bean, so the new text uses it as well, but I think we could just drop it. Name ambiguity resolution clearly operates on beans to which the name resolves, and that should be enough. No need to bother with eligibility for injection, which is not what happens here.

api/src/main/java/jakarta/enterprise/inject/Alternative.java
manovotn: > I'm not even sure why this sentence is there

Same, it seems superfluous.
To me, the wording here indicates that it might be related to `CDI.current()` as you can invoke that from modules that are not bean archives. However, even then you either gain access to the CDI container and can see and resolve whatever is enabled there, or you cannot see anything. I don't see any particular connection to just alternatives here :shrug:

api/src/main/java/jakarta/enterprise/inject/Reserve.java
```diff
@@ -0,0 +1,65 @@
...
+ * &#064;Priority(1)
+ * &#064;Dependent
+ * public class DefaultOrder extends Order { ... }
+ * </pre>
+ *
+ * <p>
+ * A reserve is not available for injection, lookup or name resolution in a module unless the module
+ * is a bean archive and the reserve is explicitly <em>selected</em> for the application (by adding
+ * the {@link Priority} annotation). A reserve is never available for injection, lookup
+ * or name resolution in a module that is not a bean archive.
```
manovotn: I know this is copied from `Alternative` text, but unless we know what does the last sentence stand for, I'd avoid adding it just for the sake of it?

> A reserve is never available for injection, lookup or name resolution in a module that is not a bean archive.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
manovotn: Perhaps we could instead just say something like this?

> The priority can be used for enablement and ordering of beans, reserves and interceptors

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
```diff
@@ -113,23 +113,37 @@ public interface SyntheticBeanBuilder<T> {
...
      * @return this {@code SyntheticBeanBuilder}
      * @throws IllegalStateException if this method is called multiple times
+     *         or if the bean was previously marked as {@linkplain #reserve(boolean) reserve}.
      */
     SyntheticBeanBuilder<T> alternative(boolean isAlternative);
 
     /**
-     * Sets a priority of this synthetic bean. To make this synthetic bean an enabled alternative,
-     * call both {@code alternative(true)} and {@code priority(some priority)}.
+     * Marks this synthetic bean as a reserve if desired. To make this synthetic bean
```
manovotn: I'd change the `reserve` to a `{@link Reserve}`

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
```diff
@@ -113,23 +113,37 @@ public interface SyntheticBeanBuilder<T> {
...
+     * If this synthetic bean is a reserve, not setting a priority means that it is not enabled.
+     * <p>
+     * If not called, this synthetic bean will not be a reserve.
+     *
+     * @param isReserve whether this synthetic bean should be a reserve
+     * @return this {@code SyntheticBeanBuilder}
+     * @throws IllegalStateException if this method is called multiple times
+     *         or if the bean was previously marked as {@linkplain #alternative(boolean) alternative}.
+     */
+    SyntheticBeanBuilder<T> reserve(boolean isReserve);
```
manovotn, line 142: Should we also add `enabledReserve(boolean, int)` method? I assume you mostly want to add a reserve that has a priority. But then we should have the same for alternatives I suppose...

Or perhaps even replace this one with `enabledReserve(boolean, Integer)` where the second parameter can be `null` meaning you don't want a priority set?

(similar approach could OFC be used in `BeanConfigurator`, I am just curious what others think about it in general)

Ladicek:
I'm not saying it has to be illegal. I'm saying it should be illegal, because it's probably a user error.

api/src/main/java/jakarta/enterprise/inject/spi/Prioritized.java
```diff
@@ -22,8 +22,8 @@
  * <p>
  * For instance {@link ObserverMethod} extends this interface to set the observer priority.
  *
- * A custom alternative {@link Bean}, {@link Interceptor} or {@link Decorator} may implement this interface to be activated
- * with a given priority
+ * A custom alternative or reserve {@link Bean}, {@link Interceptor} or {@link Decorator} may implement
+ * this interface to be activated with a given priority
```
manovotn: Unrelated to the content of this PR but perhaps we should replace `activated` for `enabled`?
That's the term we are using throughout the spec.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
manovotn: Weld doesn't have this concept yet.
However, when a container deals with ambig. dependency, you have a view of all eligible beans and you can shuffle them freely to figure it out. So technically speaking there is no issue with the algorithm, it's more a matter of a reformulating it to remove the "wait...what?!" effect when you first read it :)

Perhaps it would be enough to not talk about eliminating the beans? So the first sentence would be something like this:

> The container first considers all eligible beans that are not reserves.

api/src/main/java/jakarta/enterprise/inject/Alternative.java
Ladicek: I can remove it.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
Ladicek: You mean something like

> The priority is used for enablement and ordering of beans, especially alternatives, reserves and interceptors.

?

This suggests that other beans (not alternatives, not reserves, not interceptors) could be enabled (which is not true) and/or ordered (which is not true in the specification, but may be true in implementations). We could also say just:

> The priority is used for enablement and ordering of alternatives, reserves and interceptors.

That makes sense.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
Ladicek: Note that if we remove the notion of eliminating beans, we'd also have to rewrite the specification of `Instance.iterator()` .

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
Ladicek: I'm -1 to that. It provides nearly zero value over the existing API, conflates multiple concepts together and makes the API more complex. There's nothing wrong with people writing `reserve(true).priority(1)`.

api/src/main/java/jakarta/enterprise/inject/spi/Prioritized.java
Ladicek: Good point.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
Ladicek: Done. Also for alternatives.

api/src/main/java/jakarta/enterprise/inject/spi/Prioritized.java
Ladicek: Done.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
manovotn: Sorry, I do not recognize the difference of your last suggested sentence and mine.

I only added the word `can` to indicate the optionality of adding a priority (via stereotype) but both options we stated have the same meaning in my head :)

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
manovotn: >  There's nothing wrong with people writing reserve(true).priority(1).

True, I just recalled we had a similar Quarkus-only API at some point (which we deprecated later IIRC) so I wanted to bring it up.
Though I am totally fine not having this option.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
Ladicek: The main difference is you wrote "beans" and I wrote "alternatives" :-)

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
Ladicek: Indeed we did have a similar API, and I'm personally super happy we got rid of it :-)

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
manovotn: BTW you can keep the wording starting with:

> Otherwise, the container eliminates all eligible beans that are not alternatives, except for producer methods and fields of beans that are alternatives.

As that's where you actually can eliminate all beans.
But you're right that in context of the `iterator()` it's not any more clear that what we have now :thinking:

api/src/main/java/jakarta/enterprise/inject/Alternative.java
Azquelt: > if an archive is not a bean archive, nothing is available for injection, lookup or name resolution.

I'm not sure this is true. Beans [must be declared in a bean archive to be enabled][spec-enabled], but I think they can be _available for injection_ in non-bean archives. I can't see anything in the modularity rules in either [lite][spec-mod-lite] or [full][spec-mod-full] that would prevent that.

In practise, I think the most likely scenario for this is a Jakarta EE component which is not a bean being located in an archive which doesn't have a beans.xml and doesn't have any classes with bean defining annotations. Such an archive would not be a bean archive but I would still expect to be able to inject CDI beans into that class (just as I would be able to inject other EE resources).

I think the correct rules should be that an alternative is available for injection in a module if either:
- the alternative is ~enabled~ selected for the whole application
- the module is a bean archive and the alternative is ~enabled~ selected in that bean archive

[spec-enabled]: https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#enablement
[spec-mod-full]: https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#inter_module_injection_full
[spec-mod-lite]: https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#inter_module_injection

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
Azquelt: > > The priority is used for enablement and ordering of alternatives, reserves and interceptors.
> 
> That makes sense.

That does make sense.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
```diff
@@ -122,10 +122,18 @@ public interface BeanInfo {
...
+     * @return whether this bean is a {@linkplain jakarta.enterprise.inject.Reserve reserve}
+     */
+    boolean isReserve();
+
     /**
      * Returns the {@linkplain jakarta.annotation.Priority priority} declared on this bean, or {@code null}
      * if this bean does not declare a priority. Declaring a priority on an alternative bean makes it an enabled
-     * alternative. Similarly, declaring a priority on an interceptor makes it an enabled interceptor.
+     * alternative. Declaring a priority on a reserve bean makes it an enabled reserve. Declaring a priority
+     * on an interceptor makes it an enabled interceptor.
```
Azquelt: Should the alternatives and reserves be "selected" rather than "enabled" here?

Looks like the current spec for alternatives mostly says "selected" but sometimes says "enabled" :disappointed: 

I think interceptors still get enabled.

Azquelt: Ah, spotted a bunch more places where we also say "enabled", maybe we should clear that up later.

api/src/main/java/jakarta/enterprise/inject/Alternative.java
manovotn: > In practise, I think the most likely scenario for this is a Jakarta EE component which is not a bean being located in an archive which doesn't have a beans.xml and doesn't have any classes with bean defining annotations. Such an archive would not be a bean archive but I would still expect to be able to inject CDI beans into that class (just as I would be able to inject other EE resources).

I did a similar quick test setup before writing my review as I had the same thought - and it worked as you say. My case was with EAR that had a shared library which wasn't a CDI archive and from there I did `CDI.current()` to check some bean availability.
That said, my guess was that this is more work of an integrator to achieve this than anything guaranteed by specification. I also checked TCKs and didn't really see a case for or against it.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
manovotn: > The main difference is you wrote "beans" and I wrote "alternatives" :-)

Ah, I feel stupid, now I see why you didnt like it 🤦 
I did mean alternatives OFC, so +1 to your version

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
manovotn: Ok, let's drop that idea, marking as resolved

api/src/main/java/jakarta/enterprise/inject/Alternative.java
Ladicek: I was looking through the specification and I did find exactly the same wording for bean archives, but I _didn't_ find anything close to archives that are not bean archives. Perhaps this was left inadvertently in the javadoc, while it should have been removed.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
Ladicek: If an alternative/reserve is selected [for the application / for the bean archive], it also becomes enabled. Interceptors do not get selected, but they are still enabled.

Ladicek: It is true though that when it comes to priority, we should speak about selecting alternatives/reserves for the application. They become enabled due to being selected.

api/src/main/java/jakarta/enterprise/inject/Alternative.java
Ladicek: > I'm not sure this is true. Beans [must be declared in a bean archive to be enabled](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#enablement), but I think they can be _available for injection_ in non-bean archives. I can't see anything in the modularity rules in either [lite](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#inter_module_injection) or [full](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#inter_module_injection_full) that would prevent that.
> 
> In practise, I think the most likely scenario for this is a Jakarta EE component which is not a bean being located in an archive which doesn't have a beans.xml and doesn't have any classes with bean defining annotations. Such an archive would not be a bean archive but I would still expect to be able to inject CDI beans into that class (just as I would be able to inject other EE resources).
> 
> I think the correct rules should be that an alternative is available for injection in a module if either:
> 
>     * the alternative is ~enabled~ selected for the whole application
> 
>     * the module is a bean archive and the alternative is ~enabled~ selected in that bean archive

This is probably true, but the CDI spec says otherwise and has been saying exactly this since version 2.0:

> An alternative is not available for injection, lookup or name resolution to classes in a module unless the module is a bean archive and the alternative is explicitly selected for the bean archive or the application.

Versions 1.2 and 1.1 said:

> An alternative is not available for injection, lookup or EL resolution to classes or JSP/JSF pages in a module unless the module is a bean archive and the alternative is explicitly selected for the bean archive or the application.

Version 1.0 doesn't have selection for application:

> An alternative is not available for injection, lookup or EL resolution to classes or JSP/JSF pages in a module unless the module is a bean archive and the alternative is explicitly selected in that bean archive. An alternative is never available for injection, lookup or EL resolution in a module that is not a bean archive.

This is where the javadoc thing comes from. It's been removed in CDI 1.1, so it's about time we remove it in javadoc as well. It's superfluous.

I think we should discuss this elsewhere, because it's important, but completely unrelated to this PR. Care to file an issue?

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
Ladicek: I'm going to use this for both `BeanInfo.priority()` and `StereotypeInfo.priority()`:

> Priority is used to select alternatives and reserves for application and order them during typesafe resolution, and to enable interceptors and order them during invocation.

api/src/main/java/jakarta/enterprise/inject/Alternative.java
Ladicek: Removed that one line from javadoc. The contentious wording remains though.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
Ladicek: Done.

api/src/main/java/jakarta/enterprise/inject/Reserve.java
Ladicek: Done.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
Ladicek: Done.

api/src/main/java/jakarta/enterprise/inject/Alternative.java
Azquelt: Yep ok, I'll raise a separate issue for this.

Azquelt: Raised #882

Ladicek:
Added an extra commit (to be squashed if agreed upon) that makes it impossible for alternatives (resp. reserves) to declare reserve (resp. alternative) producers.

Ladicek:
I agree, so I squashed the two commits. It is my firm belief that alternatives declaring reserve producers and reserves declaring alternative producers is meaningless and should be forbidden. If someone ever comes with a meaningful, well, meaning, we can reconsider.

api/src/main/java/jakarta/enterprise/inject/Reserve.java
```diff
@@ -0,0 +1,63 @@
...
+ * <pre>
+ * &#064;Reserve
+ * &#064;Priority(1)
+ * &#064;Dependent
+ * public class DefaultOrder extends Order { ... }
+ * </pre>
+ *
+ * <p>
+ * During typesafe resolution, other, non-reserve beans take precedence over reserves.
+ * </p>
```
Azquelt, line 40:
```suggestion
 * <p>
 * During typesafe resolution, other non-reserve beans take precedence over reserves.
 * </p>
```

I realise that this is symmetric with the `@Alternative` javadoc, but when it's written with the two commas close together like this is looks wrong to me.

Ladicek: Good point. Since I like the symmetry with `Alternative` :-), I slightly updated both.

PR #875, opened 2025-07-09
use the term "bean name" instead of "EL name" in the API

Ladicek:
The specification text uses the term _bean name_. Some parts of the API used the old term _EL name_, which this commit rectifies.


manovotn (approved):
LGTM

PR #889, opened 2025-09-16
Clarify 'superclass' usage in the specification. Change `runtime event class` to `runtime event type`.

### LINKED ISSUE #884: Possible clarification needed in specification section on event types
The [CDI 4.1 specification reads, in part](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#event_types_and_qualifier_types):

> An event object is an instance of a concrete Java class with no unresolvable type variables. The event types of the event include all superclasses and interfaces of the runtime class of the event object.

Given this:
```
Object eventObject = new ArrayList<String>();
```
…if the part of the specification that reads "the event types of the event [object] include all superclasses and interfaces of the runtime class of the event object" is true, then the event types would not include `ArrayList<String>`, but only what `eventObject.getClass()` returns (`ArrayList.class`) and its transitive supertype closure.

Does this section of the specification therefore need to be rewritten/clarified, since Weld, at least, (properly, IMHO) includes the parameterized type `ArrayList<String>` in the collection of event types for this example?

(Clearly the event types in real-world CDI implementations such as Weld are _actually_ derived from the type argument supplied to the "broadcaster" (e.g. the `ArrayList<String>` in `jakarta.enterprise.event.Event<ArrayList<String>>`), not "the runtime class of the event object" (which cannot yield `ArrayList<String>` by itself), no matter what the specification says. It would be nice to bring the specification into line with real-world implementations.)

### LINKED ISSUE #886: Inspect usage of the `superclass` term throughout the specification
Inspired by https://github.com/jakartaee/cdi/issues/884 and the CDI meeting discussion from Sep 9, 2025.

In multiple places, the specification uses the term `superclass` while what it really _means_ is `superclass types`.
This is potentially a remnant of old times where the JLS did not differentiate between the two.
(As was rightly pointed out [here](https://github.com/jakartaee/cdi/issues/884#issuecomment-3269140244))
One such example is the wording in events chapter mentioned above.

This work needs to be done manually though, as some occurrences are bound to be correct while others should be changed.

manovotn:
Fixes #886

Did some `grep`-ing in the spec text plus javadocs and from the occurrences that came up, I thought these are the ones that need changing.

EDIT: I have added a second commit which is related and fixes #884 - namely it changes the language to talk about `runtime event type` instead of `runtime event class`

### DIFF
--- a/spec/src/main/asciidoc/cdi-spec.asciidoc
+++ b/spec/src/main/asciidoc/cdi-spec.asciidoc
@@ -101,4 +101,7 @@ include::core/packagingdeployment_full.asciidoc[]
 
 include::javase/javase.asciidoc[]
 
+[[part_3]]
+= Part III - Glossary
 
+include::glossary.asciidoc[]
--- a/spec/src/main/asciidoc/glossary.asciidoc
+++ b/spec/src/main/asciidoc/glossary.asciidoc
@@ -1,0 +1,45 @@
+////
+Copyright (c) 2025 Contributors to the Eclipse Foundation
+
+This program and the accompanying materials are made available under the
+Apache Software License 2.0 which is available at:
+https://www.apache.org/licenses/LICENSE-2.0.
+
+SPDX-License-Identifier: Apache-2.0
+////
+
+This glossary includes terms used throughout the CDI specification that are either defined in other specifications, in which case links are provided, or are used in the CDI specification but not defined precisely, in which case the terms are defined here.
+Under no circumstances should the definitions here be interpreted as affecting other specifications that might use the same terms.
+
+The other specifications referenced here are:
+
+- link:https://docs.oracle.com/javase/specs/jls/se17/html/index.html[The Java Language Specification, Java SE 17 Edition] (_JLS_ for short)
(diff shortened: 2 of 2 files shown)


manovotn:
> Suggestion: perhaps use "supertypes" instead of "superclass types" as the former is [defined rigorously by the JLS](https://docs.oracle.com/javase/specs/jls/se24/html/jls-4.html#jls-4.10).

My reading of the JLS is that `supertypes` would indeed be correct and would mean both - superclass types as well as superinterface types. 
However, it's not quite obvious without looking at the JLS, which we don't directly reference (not to mention it's not an easy read for ordinary user). OTOH, the terms `superclass type` and `superinterface type` seem fairly self-explanatory while still present in the JLS.

I can be outvoted of course, this is a just a draft to get some momentum on the issue before it gathers too much dust :)

Ladicek:
I would also like to highlight that using the term _supertype_ as defined in the JLS would (subtly, but still) change the meaning here. This is because [4.10.2](https://docs.oracle.com/javase/specs/jls/se17/html/jls-4.html#jls-4.10.2) explicitly says that a parameterized type `C<T1, ..., Tn>` has, among others, the raw type `C` in the set of its direct supertypes.

Ladicek:
I'll do that.

manovotn (commented):
The glossary is really nice :+1:

PR #892, opened 2025-09-18
Forbid injecting Event/Instance with a wildcard type

### LINKED ISSUE #888: Explicitly forbid wildcards in Event/Instance injection points
Recently I came across a corner cases regarding injection of `Event` and `Instance` with wildcards in them.

So cases such as:

```java
@Inject
Event<?> wildEvent;

@Inject
Event<? extends Foo> extendsFooEvent;

@Inject
Instance<?> wildInstance;
```

I think the specification is currently not clear on whether this should work and I propose that we explicitly forbid it. Note that this includes:
* Validating the IPs like I showed in the code above
* As well as runtime validation for cases such as `BeanManager.createInstance().select(new TypeLiteral<Event<?>>(){}`
  * This should be doable as you can only get to wildcards using the `select(TypeLiteral, Annotation...)` variant

Looking at existing implementations, I found that:
* Weld already forbids this for `Instance` and `Event` (throwing a `DefinitionException`)
* Quarkus (Arc) forbids this for `Instance` (`DefinitionException`) but permits `Event`

I think we should forbid it because of several reasons.
Firstly, in case of `Instance`, the specification IMO already indirectly suggests it is not permitted via [*required type*](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#dynamic_lookup) of `Instance` combined with [Legal injection point types](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#legal_injection_point_types) combined with [Legal bean types](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#legal_bean_types) . Granted, this does not *forbid* it, but certainly doesn't list wildcard as allowed "type". Note that this does not apply to `Event`.

Secondly,  usability - just having a wildcard (`wildEvent` field in the code above) or one with upper bound (`extendsFooEvent` in the code above) has basically no use. You cannot fire an actual type (i.e. `String`)  event from it and attempting to `select()` a subclass won't compile either. For `Event`, the only "usable" case would be a wildcard with lower bound but even then it has dubious benefits as you might well have event with given actual type. The wildcard will also not help in infering any type variables found in the actual runtime payload type.
Similar applies to `Instance` where looking for a bean with wildcard type (which isn't allowed) doesn't make sense.

Thirdly, the type variable `T` of `Event<T>` and `Instance<T>` should represent a selected/required type and you normally cannot declare a type that would be a wildcard so you shouldn't be able to do it this way either.

Thoughts?

Cc @ljnelson whose question got me digging into this.

manovotn:
Fixes #888 

A proposal for the spec phrasing.

The initial draft forbids the explicit injection of Event/Instance with a wildcard type but leaves one scenario as deliberately non-portable which is the following:

```java
@Inject
Instance<Object> instance;

public void doSomething() {
    // Select Event with wildcard type from an Instance
    final Event<? super List<String>> e = instance.select(new TypeLiteral<Event<? super List<String>>>() {}).get();
}
```

I do not like leaving this as non-portable. However, validating it is only possible at runtime either when you attempt the `select()/get()` or when you then `fire()` such an event. And at the moment I am not convinced that adding a mandatory runtime validation overhead used in any invocation of that method just to catch this rare "use case" is worth it.
Stating that it is non-portable already signifies that it should not be relied upon without enforcing it via TCKs.

I can be convinced otherwise of course; this should be specifiable along the lines of `If a programmatically obtained Instance has a required type of `Event` that has a specified type that is a wildcard type, ...`.


spec/src/main/asciidoc/core/events.asciidoc
```diff
@@ -167,6 +167,10 @@ Event<AdminLoggedInEvent> admin = any.select(
 
 If the specified type contains a type variable, an `IllegalArgumentException` is thrown.
 
+The specified type must not be a wildcard type.
+If an injected `Event` has a specified type that is a wildcard type, the container treats it as a definition error.
+If a programmatically obtained `Event` has a required type that is a wildcard type, non-portable behavior results.
```
Ladicek:
```suggestion
If a programmatically obtained `Event` has a specified type that is a wildcard type, non-portable behavior results.
```

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -653,6 +653,9 @@ Instance<AsynchronousPaymentProcessor> async = anyPaymentProcessor.select(
             AsynchronousPaymentProcessor.class, new AsynchronousQualifier() );
 ----
 
+The required type must not be a wildcard type.
+If an injected `Instance` has a required type that is either a wildcard type or an `Event` with a wildcard type, the container treats it as a definition error.
```
Ladicek:
```suggestion
If an injected `Instance` has a required type that is either a wildcard type or an `Event` whose specified type is a wildcard type, the container treats it as a definition error.
```

Ladicek:
LGTM otherwise.

spec/src/main/asciidoc/core/events.asciidoc
manovotn: Done

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
manovotn: Good point, done

Ladicek:
Here's a way how to programmatically obtain an `Instance` with a wildcard required type:

```java
@Inject
BeanContainer c;

public void xyz() {
    var x = c.createInstance().select(new TypeLiteral<Instance<? extends Number>>() {});
    var y = x.get();
    var z = y.get();
}
```

Crazy, but doable :-)

manovotn:
Ah, that's sneaky.


> Here's a way how to programmatically obtain an `Instance` with a wildcard required type:
> 
> ```java
> @Inject
> BeanContainer c;
> 
> public void xyz() {
>     var x = c.createInstance().select(new TypeLiteral<Instance<? extends Number>>() {});
>     var y = x.get();
>     var z = y.get();
> }
> ```
> 
> Crazy, but doable :-)

Nice once, that's sneaky and very questionable usage! :)
I have added the suggested sentence.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -653,6 +653,11 @@ Instance<AsynchronousPaymentProcessor> async = anyPaymentProcessor.select(
             AsynchronousPaymentProcessor.class, new AsynchronousQualifier() );
 ----
 
+The required type must not be a wildcard type.
+If an injected `Instance` has a required type that is either a wildcard type or an `Event` whose specified type is a wildcard type, the container treats it as a definition error.
+If a programmatically obtained `Instance` has a required type that is a wildcard type, non-portable behavior results.
```
Ladicek, line 658: Sorry, my bad -- we should repeat the previous sentence here too.

```suggestion
If a programmatically obtained `Instance` has a required type that is either a wildcard type or an `Event` whose specified type is a wildcard type, non-portable behavior results.
```

It is relatively straightforward to do that:

```java
@Inject
BeanContainer c;

public void xy() {
    Instance<Event<? super Number>> x = c.createInstance().select(new TypeLiteral<Event<? super Number>>() {});
    Event<? super Number> y = x.get();
}
```

manovotn: Not sure I understand. These steps give you a programmatically obtained `Event` with a wildcard type and that's already covered in the previous section. Or am I missing something?

Ladicek: Oh, good point! I didn't realize that. You're right, the current provisions are enough.

Ladicek (approved):

PR #899, opened 2025-10-02
Add beans XSD file for 5.0

### LINKED ISSUE #896: Provide `beans_5_0.xsd` for upcoming 5.0
Every release should have a version of this file corresponding to the release number and it appears we are missing one for CDI 5.0.

manovotn:
Fixes #896 

Contents are identical to 4.1, I have just changed references and default that mentioned 4.1 to now mention 5.0.

### DIFF
--- a/api/src/main/resources/beans_5_0.xsd
+++ b/api/src/main/resources/beans_5_0.xsd
@@ -1,0 +1,369 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Copyright 2025, Contributors to the Eclipse Foundation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


api/src/main/resources/beans_5_0.xsd
```diff
@@ -0,0 +1,369 @@
...
+         https://jakarta.ee/xml/ns/jakartaee/beans_5_0.xsd"
+         version="5.0">
+            ...
+         </beans>
+
+        The deployment descriptor may indicate the published version of
+        the schema using the xsi:schemaLocation attribute for the Java EE
+        namespace with the following location:
+
+        https://jakarta.ee/xml/ns/jakartaee/beans_4_0.xsd
```
Ladicek: Forgotten?

manovotn: You're right. Fixed.

PR #907, opened 2025-10-21
specification of asynchronous invokers

### LINKED ISSUE #859: Invoker support for asynchronous methods
**Is your feature request related to a problem? Please describe.**

Invokers cannot be used to portably call asynchronous methods. This would prevent Invokers being used to call Jakarta REST resource methods which execute asynchronously and return `CompletionStage` in a safe way without relying on non-spec features of the CDI container in use.

**Background**

Invokers allow the target instance and arguments for the method being invoked to be looked up from CDI. It looks up the bean matching the type and qualifiers, uses that as the instance or argument, calls the method and then destroys the reference afterwards.

For dependent scoped beans, this means that the bean will live for the duration of the method invocation and then be destroyed.

This is a problem if the method represents an asynchronous operation which may still be ongoing when the method returns and still require access to the beans used as the instance or arguments.

The spec [does allow][spec-impl-specific] for implementation-specific solutions to this problem, but such solutions are not portable.

Additionally, some practical considerations I think are important here:
- this only affects `@Dependent` scoped beans
  - normal-scoped beans will have their lifecycle tied to their context. The reference injected will be a proxy object which will delegate to the bean instance from the current context.
- some scopes (including `@RequestScoped`) may not be usable for asynchronous use anyway
- as far as I'm aware, the only consequence of destroying a `@Dependent` scoped bean is that any `@PreDestroy` methods of that bean or its dependents will be run. If there are no pre-destroy lifecycle callbacks, there may not be an issue with continuing to use the bean instance after it's been destroyed.

**Describe the solution you'd like**

An API is added which allows the looked up bean references to remain valid until after the asynchronous operation ends. There are various options for what this API may look like discussed in #749 and #768.

**Describe alternatives you've considered**
TODO add options discussed previously

**Additional context**

Ladicek:
Fixes #859

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/InvokerValidation.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/InvokerValidation.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2025, Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * Apache Software License 2.0 which is available at:
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package jakarta.enterprise.inject.build.compatible.spi;
+
+import java.util.function.Supplier;
+
+/**
+ * Includes methods for validating invokers during {@link Validation @Validation}.
+ *
+ * @since 5.0
+ */
+public interface InvokerValidation {
+    /**
+     * Ensures that an {@linkplain jakarta.enterprise.invoke.AsyncHandler async handler} exists for the given async type.
+     * If it does not exist, registers a deployment problem as if {@link Messages#error(Exception)} was called.
+     *
+     * @param asyncType the required async type, must not be {@code null}
+     */
+    default void ensureAsyncHandlerExists(Class<?> asyncType) {
+        ensureAsyncHandlerExists(asyncType, () -> null);
+    }
+
+    /**
+     * Ensures that an {@linkplain jakarta.enterprise.invoke.AsyncHandler async handler} exists for the given async type.
+     * If it does not exist, registers a deployment problem as if {@link Messages#error(Exception)} was called.
+     * The deployment problem includes the given {@code message} in a suitable, implementation-defined manner.
+     *
+     * @param asyncType the required async type, must not be {@code null}
+     * @param message supplier of the error message, must not be {@code null} but may return {@code null}
+     */
+    void ensureAsyncHandlerExists(Class<?> asyncType, Supplier<String> message);
+}
--- a/api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
+++ b/api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2025, Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * Apache Software License 2.0 which is available at:
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package jakarta.enterprise.invoke;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Flow;
+
+/**
+ * Includes logic for determining when an asynchronous action started by an invocation
+ * of an asynchronous method completes. Service providers of this interface are called
+ * <em>async handlers</em>. For the purpose of method invokers, a method is called
+ * <em>asynchronous</em> if it matches an async handler.
+ * <p>
+ * An <em>async type</em> of an async handler is the erasure of the type argument
+ * to the {@code AsyncHandler} direct superinterface type.
+ * <p>
+ * An async handler must be annotated either {@link ReturnType} or {@link ParameterType}.
+ * This annotation determines how methods are matched and how the async handler is used.
+ * <p>
+ * The {@link #transform(Object, Runnable) transform()} method of a matching async handler
+ * is called exactly once by {@link Invoker#invoke(Object, Object[]) Invoker.invoke()}.
+ * It takes an {@code original} asynchronous action and returns a transformation of it.
+ * This transformation calls {@code completion.run()} exactly once, when the asynchronous
+ * action completes.
+ * <p>
+ * For example, an async handler for {@link java.util.concurrent.CompletionStage} may look like:
+ *
+ * <pre>
+ * &#064;ReturnType
+ * public class CompletionStageAsyncHandler&lt;T&gt; implements AsyncHandler&lt;CompletionStage&lt;T&gt;&gt; {
+ *     public CompletionStage&lt;T&gt; transform(CompletionStage&lt;T&gt; original, Runnable completion) {
+ *         return original.whenComplete((value, error) -> {
+ *             completion.run();
+ *         });
+ *     }
+ * }
+ * </pre>
+ *
+ * The CDI container must provide an async handler for the return types of
+ * {@link java.util.concurrent.CompletionStage CompletionStage},
+ * {@link java.util.concurrent.CompletableFuture CompletableFuture} and
+ * {@link Flow.Publisher Flow.Publisher}.
+ *
+ * @param <T> the type that represents the asynchronous action
+ */
+public interface AsyncHandler<T> {
+    /**
+     * Takes the {@code original} asynchronous action and returns a variant of it, transformed
+     * such that {@code completion.run()} is called on completion. Note that {@code completion.run()}
+     * must be called exactly once.
+     * <p>
+     * It is recommended that {@code completion.run()} is called <em>before</em> completion
+     * is propagated to the caller.
+     *
+     * @param original the original asynchronous action
+     * @param completion action that must be executed when the asynchronous action completes
+     * @return the transformed async action
+     */
+    T transform(T original, Runnable completion);
+
+    /**
+     * If an async handler is annotated {@code @ReturnType}, the target method matches when
+     * the erasure of its return type is identical to the async type of the async handler.
+     * <p>
+     * In such case, the {@code AsyncHandler.transform()} method is called after the target
+     * method returns. The original return value is passed to {@code transform()} as
+     * {@code original} and the return value of {@code transform()} is returned to the caller.
+     * <p>
+     * If an async handler is annotated {@code @ReturnType}, it must not be annotated
+     * {@code @ParameterType}, otherwise deployment problem occurs.
+     */
+    @Target(ElementType.TYPE)
+    @Retention(RetentionPolicy.RUNTIME)
+    @interface ReturnType {
+    }
+
+    /**
+     * If an async handler is annotated {@code @ParameterType}, the target method matches when
+     * it declares exactly one parameter whose type's erasure is identical to the async type
+     * of the async handler.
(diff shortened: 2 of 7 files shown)


api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
```diff
@@ -0,0 +1,89 @@
+package jakarta.enterprise.invoke;
```
Azquelt: Need a copyright header

api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
```diff
@@ -0,0 +1,89 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Includes logic for determining when an asynchronous action started by an invocation
+ * of an asynchronous method completes. Service providers of this interface are called
+ * <em>async handlers</em>. For the purpose of method invokers, a method is called
+ * <em>asynchronous</em> if it matches an async handler.
+ * <p>
+ * An <em>async type</em> of an async handler is the erasure of the type argument
+ * to the {@code AsyncHandler} direct superinterface type.
```
Azquelt, line 28: So that means if you implement `AsyncHandler<CompletionStage<String>>`, the async type is just `CompletionStage`?

Should we replace "the type argument" with `{@code T}`

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -213,24 +213,64 @@ It is a deployment problem if an attempt to identify a looked up bean results in
...
+
+[source,java]
+----
+public interface AsyncHandler<T> {
+    T transform(T original, Runnable completion);
+}
+----
+
+An _async handler_ is a service provider of `jakarta.enterprise.invoke.AsyncHandler`.
+An async handler must not declare a provider method; it must declare a provider constructor.
```
Azquelt: Provider constructor?

api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
Ladicek: Correct. This is just to make the specification of matching slightly shorter, the term _async type_ is not used anywhere else.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: This is specified in `java.util.ServiceLoader`.

I'm actually not sure if we need this provision, but if we allow provider methods, the service provider doesn't have to implement `AsyncHandler` and figuring out the async type would be harder. I'm not sure if anyone actually uses provider methods; the other parts of the CDI spec ignore them completely...

Azquelt:
>     1. Currently, we specify that `withAsync()` must not be called for non-async methods. We could allow calling `withAsync(false)`; should we?

I think we should allow `withAsync(false)` to state that the method should be treated non-asynchronously.

>     2. Currently, we specify that "it is recommended that completion is signalled before it is propagated to the caller of `Invoker.invoke()`"; should we make that _mandatory_? I _think_ it's straightforward to signal completion before propagating it to the caller for all types except JAX-RS `AsyncResponse`. I don't know what happens if I register an `InvocationCallback` on the `AsyncResponse`, so I went with a recommendation for now.

I think a recommendation makes sense. If you have a listener model, I don't think you can guarantee that.

>     3. Similarly to the previous question, I'm also not sure about async methods throwing synchronously. Again, it seems straightforward for all types except JAX-RS `AsyncResponse`.

It's straightforward when the return type represents the async task. If it's an argument, then I don't think there's an obvious answer as to whether it'll call the completion task or not.

I would suggest that we say that cleanup will happen immediately if the invocation throws an exception, and that the container must tolerate the completion being run as well.

Azquelt:
> One other question I don't currently have an answer for: what if there's multiple async handler for the same async type?

I'm a bit concerned about this too since I think it's likely that multiple libraries will provide an async handler for common async types where support from the container is not required.

We could say that any of them may be called, which is a bit unpredictable, but in theory fine as long as all the deployed async handlers are correct.

We could say that it's a deployment error, though that might be difficult for the user to resolve if the handlers are provided by libraries.

I'm also a bit concerned about the scope of an async handler in a multi-module scenario. If I deploy an async handler in a module, is it available for use by the whole application, or only invocations of beans in modules which have visibility of the async handler module.

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
```diff
@@ -81,6 +88,21 @@ public interface InvokerBuilder<T> {
...
     InvokerBuilder<T> withArgumentLookup(int position);
 
+    /**
+     * Sets the {@link AsyncHandler} to use when invoking the asynchronous method. when
+     * {@code asyncHandler} is not {@code null}, the built invoker is asynchronous. It may
+     * be {@code null} in the (unlikely) case of an asynchronous method that should
+     * <em>not</em> be invoked using an asynchronous invoker.
+     * <p>
+     * This method must be called when the target method is asynchronous and must not be called
+     * when the target method is not asynchronous.
```
Azquelt: Do we want to restrict calling `withAsync(null)` for non-asynchronous methods?

I would change this to "must not be called with a non-null value when the target method is not asynchronous"

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
```diff
@@ -81,6 +88,21 @@ public interface InvokerBuilder<T> {
...
+     *
+     * @param asyncHandler the {@link AsyncHandler} to use; may be {@code null}
```
Azquelt:
```suggestion
     * @param asyncHandler the {@link AsyncHandler} to use, or {@code null} to indicate that the invoker is synchronous
```

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -213,24 +213,65 @@ It is a deployment problem if an attempt to identify a looked up bean results in
...
+
+[source,java]
+----
+public interface InvokerBuilder<T> {
+    InvokerBuilder<T> withAsync(Class<? extends AsyncHandler> asyncHandler);
+}
+----
+
+If the target method is asynchronous, the `withAsync()` method on `InvokerBuilder` must be called, otherwise deployment problem occurs.
+If the target method is not asynchronous, the `withAsync()` method on `InvokerBuilder` must not be called, otherwise deployment problem occurs.
```
Azquelt:
```suggestion
If the target method is not asynchronous, the `withAsync()` method on `InvokerBuilder` must not be called with a non-`null` value, otherwise deployment problem occurs.
```
If we make the corresponding change in the Javadoc.

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -213,24 +213,65 @@ It is a deployment problem if an attempt to identify a looked up bean results in
...
+[source,java]
+----
+public interface AsyncHandler<T> {
+    T transform(T original, Runnable completion);
+}
+----
+
+An _async handler_ is a service provider of `jakarta.enterprise.invoke.AsyncHandler`.
+An async handler may not declare a provider method; it must declare a provider constructor.
+An async handler must have a direct superinterface type that is a parameterized type whose generic interface is `AsyncHandler` and its sole type argument is a class type, interface type or parameterized type, otherwise deployment problem occurs.
```
Azquelt, line 242: This pair of restrictions seem unusual for a service provider. Is this necessary so that we can reliably find out what `T` is, or is there some other reason?

The first restriction seems difficult to detect or enforce if the handler was loaded with `ServiceLoader.load`, so should I infer that you're not expecting implementations to do that?

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
Ladicek: I would prefer to keep it, because it's nicely symmetric: if the target method is async, this method must be called, and if the target method is not async, this method may not be called.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: I do very much expect implementations to use `ServiceLoader.load()`, which is not an issue in runtime-oriented implementations, which will find async handlers at runtime and may easily just hold onto the implementations obtained from the service loader. Build-time oriented implementations will call `ServiceLoader.load()` during build (at deployment time), and if we allowed provider methods, build-time oriented implementations would have a hard time matching between what they obtained at build time and what they obtained again at runtime. If we only allow provider constructors, build-time oriented implementations can call `ServiceLoader.load()` just once, during build, and then just call the provider constructor at runtime at suitable place, making the matching trivial.

I actually think all other places in CDI spec that rely on `ServiceLoader` also expect only provider constructors and no provider methods, but I don't think anyone has ever tested this.

The second provision is to prevent situations such as `class Foo implements AsyncHandler` or `class Bar<T> implements AsyncHandler<T>`.

api/src/main/java/jakarta/enterprise/invoke/InvokerBuilder.java
Azquelt: I get that, but it does mean that if another library is added which adds an async handler for your return type, you need to call `withAsync(null)` if that library is present and not call it if it isn't present.

Ladicek: I think that situation is weird enough that users want to know about it instantly. (Also it seems quite unlikely. There's not that many async types out there.)

Azquelt: It does seem unlikely, but I really dislike the idea of having to call the method if an unrelated library is present and not call it if the library is not present and having no way of writing the code that works in both situations.

I guess you could argue that the developer can use reflection to detect whether the other library is present, but that seems more nasty and error prone.

Ladicek: I would say if you found yourself in such situation, you want know about it and you need to think about it. Using a `null` handler may be a valid solution, but I'd argue that more often than not, there will be other things you want to do.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/InvokerValidation.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Ensures that an {@linkplain jakarta.enterprise.invoke.AsyncHandler async handler} exists for the given async type.
+     * If it does not exist, registers a deployment problem as if {@link Messages#error(Exception)} was called.
+     * The deployment problem includes the given {@code message} in a suitable, implementation-defined manner.
+     *
+     * @param asyncType the required async type, must not be {@code null}
+     * @param message the error message, may be {@code null}
+     * @throws IllegalStateException if called outside of the observer method invocation
+     */
+    void ensureAsyncHandlerExists(Class<?> asyncType, String message);
```
Azquelt: Should we change this to `Supplier<String>` for the message?

If someone is providing a message with more detail, I could see them wanting to use a dynamically created message, including information about the affected async methods.

Ladicek: This occurs during `@Validation` (or `AfterDeploymentValidation` in PEs), so I find it hard to believe people will stash all invoked methods to be able to use them here. When I thought about this, I imagined this would be just a simple constant that e.g. identifies what framework has asked for the async type, or something like that. But using a `Supplier` cannot hurt I guess :-) Will change and squash the commits.

Ladicek: Done.

manovotn: @Ladicek you also need to change the spec text

Ladicek: Ah, good point, thanks! Fixed.

PR #911, opened 2025-10-30
add `@Eager`

### LINKED ISSUE #835: Consider introducing @Eager or similar annotation for bean eager init
This has been discussed and requested repeatedly (last I know of was [here](https://github.com/jakartaee/cdi/issues/834#issuecomment-2599463154)). 
Note that this is *not* a new feature as the same can be achieved via declaration of `Startup` event observer in any given bean.
However, that is often perceived as a workaround instead of a solution and is admittedly longer than using an annotation.

The idea would be to introduce an annotation such as `@Eager` (_name is subject to change_), declarable at bean class level, which would make CDI initialize the bean on container startup.

Some thoughts:
* This makes most sense for `@ApplicationScoped` (or `@Singleton`) beans but can technically work for all scopes
  * Should we forbid it for some scopes? Doesn't seem to make much sense for dependent beans for instance.
  * What about contexts that aren't active at startup? We should probably align this with how a startup event observer would behave.
* Nothing in CDI specification requires impl to behave lazily (not create a bean until its method is invoked) yet this features indirectly assumes it as it would otherwise have no effect. This doesn't prevent us from adding it; just saying :)

FTR, some discussion about implementation took place in the aforementioned issue ([here](https://github.com/jakartaee/cdi/issues/834#issuecomment-2601941256)). It boils down to:
* Gathering information about eager-init beans during discovery 
* Resolving them after container starts (along the lines of `Instance#get()`)
  * Beans without client proxies will be created as soon as they are resolved.
  * Bean with client proxies will need to be "forcefully" initialized.
    * Dumb version is invoking their `toString()` which would work in 99% cases.
    * Otherwise needs to manually invoke their creation cycle for which some impls already have code anyway ([here's Weld bit](https://github.com/weld/api/blob/master/weld/src/main/java/org/jboss/weld/proxy/WeldClientProxy.java#L48-L54))

Ladicek:
This annotation may be put on `@ApplicationScoped` beans to trigger eager initialization. This occurs at the same time as the `Startup` event.

Putting the annotation on a bean of any other scope leads to definition error. This may be relaxed in the future, if we find suitable meaning.

For example, eagerly initializing `@RequestScoped` beans typically only makes sense for _some_ beans, depending on the nature of the request (such as the URL in case of an HTTP request). My personal opinion is that such conditions are best expressed in code, which is already possible, because all beans can declare observer methods, which may include arbitrary code.

`@Eager` is a shortcut for such observer in the most common scenario. It also serves as a direct replacement of EJB `@Startup`, which also only makes sense in case of application-wide beans (EJB `@Singleton`).

Fixes #835

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
@@ -138,6 +138,7 @@ public interface SyntheticBeanBuilder<T> {
      * @return this {@code SyntheticBeanBuilder}
      * @throws IllegalStateException if this method is called multiple times
      *         or if the bean was previously marked as {@linkplain #alternative(boolean) alternative}.
+     * @since 5.0
      */
     SyntheticBeanBuilder<T> reserve(boolean isReserve);
 
--- a/api/src/main/java/jakarta/enterprise/inject/Reserve.java
+++ b/api/src/main/java/jakarta/enterprise/inject/Reserve.java
@@ -48,6 +48,8 @@ import jakarta.enterprise.util.AnnotationLiteral;
  * <p>
  * Unlike {@linkplain Alternative alternatives}, reserves cannot be selected for a bean archive in {@code beans.xml}.
  * </p>
+ *
+ * @since 5.0
  */
 @Target({ TYPE, METHOD, FIELD })
 @Retention(RUNTIME)
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BeanInfo.java
@@ -126,6 +126,7 @@ public interface BeanInfo {
      * Returns whether this bean is a {@linkplain jakarta.enterprise.inject.Reserve reserve}.
      *
      * @return whether this bean is a {@linkplain jakarta.enterprise.inject.Reserve reserve}
+     * @since 5.0
      */
     boolean isReserve();
 
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/StereotypeInfo.java
@@ -54,6 +54,7 @@ public interface StereotypeInfo {
      * This means that all beans with this stereotype are reserves.
      *
      * @return whether this stereotype is meta-annotated {@link jakarta.enterprise.inject.Reserve @Reserve}
+     * @since 5.0
      */
     boolean isReserve();
 
--- a/api/src/main/java/jakarta/enterprise/inject/spi/AfterTypeDiscovery.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/AfterTypeDiscovery.java
@@ -51,6 +51,7 @@ public interface AfterTypeDiscovery {
     /**
      * @return the list of enabled reserves for the application, sorted by priority in ascending order.
      * @throws IllegalStateException if called outside of the observer method invocation
+     * @since 5.0
      */
     public List<Class<?>> getReserves();
 
--- a/api/src/main/java/jakarta/enterprise/inject/spi/BeanAttributes.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/BeanAttributes.java
@@ -86,6 +86,7 @@ public interface BeanAttributes<T> {
      *
      * @return <code>true</code> if the bean is a {@linkplain Reserve reserve}, and <code>false</code>
      *         otherwise.
+     * @since 5.0
      */
     public default boolean isReserve() {
         // `default` to avoid breaking older `Bean` implementations
--- a/api/src/main/java/jakarta/enterprise/inject/spi/configurator/BeanAttributesConfigurator.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/configurator/BeanAttributesConfigurator.java
@@ -190,6 +190,7 @@ public interface BeanAttributesConfigurator<T> {
      *
      * @param value value for reserve property
      * @return self
+     * @since 5.0
      */
     BeanAttributesConfigurator<T> reserve(boolean value);
 
--- a/api/src/main/java/jakarta/enterprise/inject/spi/configurator/BeanConfigurator.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/configurator/BeanConfigurator.java
@@ -323,6 +323,7 @@ public interface BeanConfigurator<T> {
      *
      * @param value value for reserve property
      * @return self
+     * @since 5.0
      */
     BeanConfigurator<T> reserve(boolean value);
 


api/src/main/java/jakarta/enterprise/context/Eager.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Marks the annotated bean as eagerly initialized.
+ *
+ * @since 5.0
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD, ElementType.FIELD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface Eager {
```
Azquelt, line 36: My gut feeling is that many more people read the javadoc than read the spec so I think we should include the following information here:
- `Eager` is equivalent to observing `Startup`
- `Eager` may only be used on `ApplicationScoped` beans
- `Eager` can be applied to a managed bean class, a producer method or a producer field

Azquelt (commented):
> I'm sure my decision to only allow `@Eager` on `@ApplicationScoped` beans will spark a controversy, but well, such is life. I actually originally specified it for all scopes, but decided against it for reasons mentioned in the PR description.

> For example, eagerly initializing `@RequestScoped` beans typically only makes sense for some beans, depending on the nature of the request (such as the URL in case of an HTTP request). My personal opinion is that such conditions are best expressed in code, which is already possible, because all beans can declare observer methods, which may include arbitrary code.

This is an interesting case. A `RequestScoped` bean may be instantiated in any request scope, whether it's associated with an HTTP request or not.

If you wanted to conditionally instantiate a `RequestScoped` bean, I guess you could do that by observing `@Initialized(RequestScoped.class)` on a _different_ bean, checking whether it's an HTTP request and if so using `BeanContainer.unwrapClientProxy` to instantiate the bean? That seems like an unusual case which `@Eager` doesn't express.

I think I'd have still made it apply to all scopes, even if there are some cases where it's not appropriate, but I don't feel too strongly about it. This covers the most common use case and if there are strong arguments for other scopes, we can allow it at a later time.

> One open question I've got: what's the best package for the Eager annotation? I tentatively put it into `jakarta.enterprise.context`, because I couldn't find a better place, but it still doesn't feel exactly right.

I think `jakarta.enterprise.context` makes sense, since eagerness is associated with the lifecycle of a context and it already contains `Initialized` and `Destroyed`. `jakarta.enterprise.inject` might be another option, alongside `Stereotype` and `Alternative` - other annotations which affect the definition of the bean.

api/src/main/java/jakarta/enterprise/context/Eager.java
Ladicek: You're right. I had a similar text to the spec in the javadoc, but decided against it, because it's relatively dense and thick. I'll add something lighter back.

Also, this reminds me I have to add support for synthetic beans -- I had it in mind originally, but forgot. I'll fix that too.

Ladicek: I think I addressed this particular comment.

Locally, I have code that allows synthetic beans to be eagerly initialized, but I think I need to prototype an implementation at least to some extent to see whether the API I've got makes sense.

ljnelson:
I had one question: if `@Eager` is sugar for (e.g.) `private void frob(@Observes @Initialized(ApplicationScoped.class) Object x) {}`, is there a more general case here? Should there instead be a meta-annotation that effectively says, "the annotated annotation is shorthand for a particular kind of observer method"? Then eager initialization is just an instance of the more general case? Or no?

manovotn:
> I think I'd have still made it apply to all scopes, even if there are some cases where it's not appropriate, but I don't feel too strongly about it. This covers the most common use case and if there are strong arguments for other scopes, we can allow it at a later time.

This very much sums up my stance on it as well except I feel rather neutral on applying this to all scopes.
We initially started the discussion with app scoped beans so it does make sense to start there and see if there is interest in expanding it later. Plus with the addition of `BeanContainer.unwrapClientProxy` we at least allow users to cover those cases themselves if they want to.

I know we also discussed this for singletons, is there a reason why you chose to leave those out now?

> One open question I've got: what's the best package for the Eager annotation? I tentatively put it into `jakarta.enterprise.context`, because I couldn't find a better place, but it still doesn't feel exactly right.

That's where I would put it as well.
Other option would be a separate package but that would then make it inconsistent with other annotations we already have so +1 for `context` package.

manovotn:
> I had one question: if `@Eager` is sugar for (e.g.) `private void frob(@Observes @Initialized(ApplicationScoped.class) Object x) {}`, is there a more general case here? Should there instead be a meta-annotation that effectively says, "the annotated annotation is shorthand for a particular kind of observer method"? Then eager initialization is just an instance of the more general case? Or no?

What other observers would you want to cover?
The reason we want to do this is that `@Initialized(XYZ.class)` was a special case in which users declared an observer which was often completely empty, just as an "intended workaround" for eager init.
Other observers in CDI are declared for the logic you want to place into them; at least in so far as I can think of.

Ladicek:
> I had one question: if `@Eager` is sugar for (e.g.) `private void frob(@Observes @Initialized(ApplicationScoped.class) Object x) {}`

Aside: it's a shortcut for `void eagerInit(@Observes Startup startup) {}`.

> is there a more general case here? Should there instead be a meta-annotation that effectively says, "the annotated annotation is shorthand for a particular kind of observer method"? Then eager initialization is just an instance of the more general case?

I was thinking about that, as you can see here: https://github.com/jakartaee/cdi/issues/835#issuecomment-2619428479 Let me rephrase.

We could allow declaring "observer shortcuts", which would be annotations meta-annotated with `@Shortcut(observed event type)` and possibly observed event qualifiers.

`Shorcut` would be declared as

```java
@Retention(RUNTIME)
@Target(ANNOTATION_TYPE)
@Documented
public @interface Shortcut {
    Class<?> value();
}
```

An observer shortcut could be added on a class, which would mean the class declares an empty observer method. It could also be added on a method, which would mean the method is effectively an observer method with an unused event parameter.

For example:

```java
@Shortcut(Startup.class)
// no qualifiers here
@Retention(RUNTIME)
@Target(TYPE, METHOD)
public @interface OnStartup {
}
```

Open questions:

- how do we specify parameterized observed event types? (e.g. you can declare an observer shortcut for `String`, but not for `List<String>`)
- in case the shortcut is put on a class, how do we know whether the hypothetical observer method is static or non-static?
- how do we know whether the hypothetical observer is synchronous or asynchronous?

Specifically in case of `@Eager`, we currently specify it's equivalent to observing `Startup` for `@ApplicationScoped` beans, but we might want to generalize it to other scopes in the future, in which case it should most likely be equivalent to observing `@Initialized(scope class) Object`, so an observer shortcut would no longer be a good fit.

If you like the idea of observer shortcuts, feel free to file another issue. If you have any answers to the open questions, feel free to include them as well :-) (My tentative ideas would be: we don't care about parameterized types, because users can always just declare an observer method directly. For sync/async, we could just add an annotation member `boolean async() default false`. For static/non-static, we could add an annotation member `boolean nonStatic() default true`, but that breaks when you add the shortcut on a method. We could specify that this member only makes sense for observer shortcuts put on a class and is ignored on observer shortcuts on methods, but it feels clumsy. We could also add an annotation member to specify priority.)

Ladicek:
> I know we also discussed this for singletons, is there a reason why you chose to leave those out now?

The only reason is that the rest of CDI spec ignores `@Singleton` completely, and I didn't want to start breaking that rule :-) So I specified that if a pseudo-scoped bean (except `@Dependent`) is `@Eager`, non-portable behavior occurs.

I did think about adding something like:

> (It is recommended that `@Eager @Singleton` beans are treated as `@Eager @ApplicationScoped` beans.)

But I didn't feel too strongly about it.

api/src/main/java/jakarta/enterprise/context/Eager.java
hantsy: > My gut feeling is that many more people read the javadoc than read the spec so I think we should include the following information here:
> 
>     * `Eager` is equivalent to observing `Startup`
> 
>     * `Eager` may only be used on `ApplicationScoped` beans
> 
>     * `Eager` can be applied to a managed bean class, a producer method or a producer field

And the JavaDoc here confused me.

>  Contextual instance of an eagerly initialized bean is created at the same time
> the {@link jakarta.enterprise.event.Startup Startup} event is fired.

If the eager is inverting the lazy proxy creation. I do not think it is equivalent to `@Observes StarupEvent`.

It should be related to how to handle the bean instantiation. Firstly creating an empty proxy like normal, then immediately execute more steps:
1. Inject the dependencies
2. Call the managed bean lifecycle `@PostConstruct` method.

In theory, any scoped beans could be operated like this. Even `applicationScoped` bean looks more reasonable.

In contrast to Spring Boot `ApplicationReady` event, the bean initialization does not depend on it.

manovotn: > If the eager is inverting the lazy proxy creation. I do not think it is equivalent to @Observes StarupEvent.

I am not sure I understand your comment but it is equivalent.
In order for CDI container to notify a bean of such an event, it first creates its instance which includes injecting into it and calling its post construct method. This is a shorthand for the same.

> In theory, any scoped beans could be operated like this. Even applicationScoped bean looks more reasonable.

See the original CDI issue, it mentioned several variants but in the end we decided to start smaller and only define it for app scoped beans which are the most common usage we know/heard of.
FTR, you can achieve the same for any scope and with the addition of `BeanContainer#unwrapClientProxy` it should be fairly simple.

Ladicek: > If the eager is inverting the lazy proxy creation. I do not think it is equivalent to `@Observes StarupEvent`.

It absolutely is. An empty observer of `Startup` only has one effect: forces the bean contextual instance into existence.

> It should be related to how to handle the bean instantiation. Firstly creating an empty proxy like normal, then immediately execute more steps:
> 
> 1. Inject the dependencies
> 2. Call the managed bean lifecycle `@PostConstruct` method.

This is all executed when a contextual instance of a bean is created. See for example https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#managed_bean_lifecycle There's exactly zero need to say all that again.

hantsy: What I mean here is that the `@Eager` should complete the instance's existence before triggering `StartupEvent`.

Ladicek: What's the difference between "before `Startup`" and "during `Startup` at priority `Integer.MIN_VALUE`"? :-)

hantsy: If there is a custom CDI bean observes `StartupEvent` which calls an eager bean, the `Eager` been **maybe** not ready for being called, because it is also triggered by `StartupEvent`.

Ladicek: This is very much not true. As mentioned above, the only thing that `@Eager` does is forcing the contextual instance of the bean into existence early on. If there's another bean calling into the `@Eager` bean before that eager bean was initialized (most likely because the other bean is also `@Eager` and calls into the first bean from a `@PostConstruct` callback or something like that), the instance will be created during that call.

Azquelt: Yeah, this is the nice thing about making an `@Eager` bean with a `@PostConstruct` method, vs observing `Startup` and doing any setup logic in the observer method; `@PostConstruct` is guaranteed to run before the bean is used, whereas observer methods are called in priority order.

All that `@Eager` does is make sure that an instance of the bean is created when the `Startup` event is fired.

Would it be clearer to say this?
```
A contextual instance of an eagerly initialized bean is created at the same time
the {@link jakarta.enterprise.event.Startup Startup} event is fired, if one does
not exist.
```

I can't decide if that detail is helpful to have in the Javadoc or just makes the sentence more confusing.

Edit: this might be better:
```
The contextual instance of an eagerly initialized bean is created at the same time
the {@link jakarta.enterprise.event.Startup Startup} event is fired, if it has not
already been created.
```

Ladicek: I think I'd rather say

> Contextual instance of an eagerly initialized bean is created ~~at the same time~~ _no later than_ the {@link jakarta.enterprise.event.Startup Startup} event is fired.

Or maybe reword it more:

> Eagerly initialized bean is guaranteed to be created at latest when the `Startup` event is fired.

hantsy: Ideally, I think the CDI implementor has a mechanism that can track all `@Eager` beans and make sure all are initialized successfully, then fire the `StartUpEvent`.

Ladicek: I still don't understand the difference between "immediately before `Startup`" and "during `Startup` at priority `Integer.MIN_VALUE`", but sure, we can say "immediately before firing `Startup`".

Aside:

> fire the `StartUpEvent`

Please stop calling it `StartupEvent` or `StartUpEvent`, it's called `Startup`.

api/src/main/java/jakarta/enterprise/context/Eager.java
```diff
@@ -0,0 +1,65 @@
...
+ * producer method or field or {@linkplain jakarta.enterprise.inject.Stereotype stereotype}.
+ * <p>
+ * Only {@link ApplicationScoped @ApplicationScoped} beans may be eagerly initialized.
+ * <p>
+ * Contextual instance of an eagerly initialized bean is created no later than at the moment
+ * the container fires the {@link jakarta.enterprise.event.Startup Startup} event.
+ *
+ * @since 5.0
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD, ElementType.FIELD })
```
manovotn, line 33:
```suggestion
@Target({ ElementType.TYPE, ElementType.METHOD, ElementType.FIELD, ElementType.ANNOTATION_TYPE })
```
Since we want to allow this on stereotypes as well.

api/src/main/java/jakarta/enterprise/context/Eager.java
```diff
@@ -0,0 +1,65 @@
...
+
+import jakarta.enterprise.util.AnnotationLiteral;
+
+/**
+ * Marks the annotated bean as eagerly initialized. May be applied to a bean class,
+ * producer method or field or {@linkplain jakarta.enterprise.inject.Stereotype stereotype}.
+ * <p>
+ * Only {@link ApplicationScoped @ApplicationScoped} beans may be eagerly initialized.
+ * <p>
+ * Contextual instance of an eagerly initialized bean is created no later than at the moment
```
manovotn:
```suggestion
 * A contextual instance of an eagerly initialized bean is created no later than at the moment
```

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
```diff
@@ -155,6 +156,21 @@ public interface SyntheticBeanBuilder<T> {
      */
     SyntheticBeanBuilder<T> priority(int priority);
 
+    /**
+     * Marks this synthetic bean as a {@linkplain jakarta.enterprise.context.Eager eagerly initialized} if desired.
```
manovotn:
```suggestion
     * Marks this synthetic bean as an {@linkplain jakarta.enterprise.context.Eager eagerly initialized bean} if desired.
```

api/src/main/java/jakarta/enterprise/context/Eager.java
Ladicek: We don't need that, `ElementType.TYPE` includes annotation types as well. See also `@Reserve` or `@Alternative`, for example.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
Ladicek: Ah, I think I fixed this somewhere else (and forgot to backport the fix) to read "Marks this synthetic bean as eagerly initialized if desired". Duplicating the word "bean" doesn't make much sense to me. Will fix like that if you agree?

api/src/main/java/jakarta/enterprise/context/Eager.java
Ladicek: I'm not sure what I think about this, but my English is pretty bad, so I'm gonna trust you on this one :-)

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
manovotn: Sure, sounds good

api/src/main/java/jakarta/enterprise/context/Eager.java
manovotn: Ah, you're right, I didn't realize that.

api/src/main/java/jakarta/enterprise/context/Eager.java
Ladicek: Done.

api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanBuilder.java
Ladicek: Done.

PR #913, opened 2025-11-12
Change GAV of CDI artifacts, provide relocation POMs for easier migration

### LINKED ISSUE #908: Make Maven coordinates consistent with rest of EE
CDI has one of the most inconsistent Maven coordinates currently.

Most every API in Jakarta EE has `jakarta.[spec name]:jakarta.[spec name]-api`.

Where "spec name" is written fully if its a simple single word name, or using an abbreviation if its multiple words:

E.g.

- jakarta.ejb:jakarta.ejb-api
- jakarta.interceptor:jakarta.interceptor-api
- jakarta.servlet:jakarta.servlet-api
- jakarta.activation:jakarta.activation-api
- jakarta.mail:jakarta.mail-api
- jakarta.transaction:jakarta.transaction-api
- jakarta.el:jakarta.el-api
- ..

CDI however uses `jakarta.enterprise:jakarta.enterprise.cdi-api`

I propose using `jakarta.cdi:jakarta.cdi-api`.

cc @keilw @JanWesterkamp-iJUG @ivargrimstad

manovotn:
Fixes #908

Changes group ID of all artifacts to `jakarta.cdi` and artifact IDs to `jakarta.cdi-[something]`.

Relocation modules are provided for all of the artifacts for ease of migration.
They should all have packaging `pom` to avoid creating empty JARs and should be built (and hence released by the Jenkins) along with the whole project build.

We will want to remove these modules as we get 5.0 out of the gate.
Their main purpose is to allow users to update their existing versions smoothly at which point they will start seeing a warning during the build informing them of the relocation.

Comments are welcome, this is the first time I am dealing with relocation modules so I might easily be wrong (although it does seem to work locally).

### DIFF
--- a/lang-model/pom.xml
+++ b/lang-model/pom.xml
@@ -11,12 +11,12 @@
 	<modelVersion>4.0.0</modelVersion>
 
 	<parent>
-		<groupId>jakarta.enterprise</groupId>
-		<artifactId>jakarta.enterprise.cdi-parent</artifactId>
+		<groupId>jakarta.cdi</groupId>
+		<artifactId>jakarta.cdi-parent</artifactId>
 		<version>5.0.0.Beta1-SNAPSHOT</version>
 	</parent>
 
-	<artifactId>jakarta.enterprise.lang-model</artifactId>
+	<artifactId>jakarta.cdi-lang-model-api</artifactId>
 	<packaging>jar</packaging>
 
 	<name>CDI Language Model</name>
--- a/relocation/api/pom.xml
+++ b/relocation/api/pom.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2025, Contributors to the Eclipse Foundation
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~ http://www.apache.org/licenses/LICENSE-2.0
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>jakarta.enterprise</groupId>
+        <artifactId>jakarta.enterprise.cdi-parent</artifactId>
+        <version>5.0.0.Beta1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jakarta.enterprise.cdi-api</artifactId>
+    <packaging>pom</packaging>
+
+    <name>CDI APIs (Relocation)</name>
+    <description>APIs for CDI (Contexts and Dependency Injection for Java)</description>
+
+    <licenses>
+        <license>
+            <name>Apache License 2.0</name>
+            <url>https://repository.jboss.org/licenses/apache-2.0.txt</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
+
+    <distributionManagement>
+        <relocation>
+            <groupId>jakarta.cdi</groupId>
+            <artifactId>jakarta.cdi-api</artifactId>
+        </relocation>
+    </distributionManagement>
+</project>
--- a/relocation/spec/pom.xml
+++ b/relocation/spec/pom.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2025, Contributors to the Eclipse Foundation
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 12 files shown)


lang-model/pom.xml
```diff
@@ -11,12 +11,12 @@
...
 	<parent>
-		<groupId>jakarta.enterprise</groupId>
-		<artifactId>jakarta.enterprise.cdi-parent</artifactId>
+		<groupId>jakarta.cdi</groupId>
+		<artifactId>jakarta.cdi-parent</artifactId>
 		<version>5.0.0.Beta1-SNAPSHOT</version>
 	</parent>
 
-	<artifactId>jakarta.enterprise.lang-model</artifactId>
+	<artifactId>jakarta.lang-model</artifactId>
```
Ladicek: This should IMHO be `jakarta.lang-model-api` or even better `jakarta.cdi-lang-model-api`.

relocation/ide-config/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+        </license>
+    </licenses>
+
+    <distributionManagement>
+        <relocation>
+            <groupId>jakarta.cdi</groupId>
+            <artifactId>jakarta.cdi-ide-config</artifactId>
+        </relocation>
+    </distributionManagement>
+</project>
```
Ladicek: I'm fairly sure we don't need a relocation for `cdi-ide-config`, that's just an internal artifact noone should use.

relocation/spec/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+        </license>
+    </licenses>
+
+    <distributionManagement>
+        <relocation>
+            <groupId>jakarta.cdi</groupId>
+            <artifactId>jakarta.cdi-spec-doc</artifactId>
+        </relocation>
+    </distributionManagement>
+</project>
```
Ladicek, line 47: This relocation feels useless, do we actually have anyone consuming the spec from Maven?

If we don't know, let's keep it, for sure, but it feels weird.

Ladicek:
Just a few small comments, but looks good. I'll take a proper look tomorrow, but I don't expect any other finding.

lang-model/pom.xml
manovotn: I was about to add a comment here myself - I was considering that but I know we originally wanted to offer this API outside of CDI so stripping of the `cdi` affix in the artifact ID made sense to me.
That said, we haven't had that usage so far so I am not feeling strongly about it :shrug:

relocation/spec/pom.xml
manovotn: I agree it feels useless but it should cost us basically zero effort to have it which is why I added it as well.
OFC I have no clue whether someone uses it or not.

relocation/ide-config/pom.xml
manovotn: You're probably right, I went a bit over the top with this one :)

lang-model/pom.xml
manovotn: On second thought, I have changed it to `jakarta.cdi-lang-model-api`. 
It is, after all, more consistent with the rest of artifacts.

relocation/spec/pom.xml
Ladicek: Let's keep it then. No problem having one more relocation in one release :-)

relocation/api/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>jakarta.enterprise</groupId>
+        <artifactId>jakarta.enterprise.cdi-parent</artifactId>
```
Ladicek, line 23: These coordinates will no longer exist, no? I don't think we're relocating the parent as well (and I'm not sure if that would even work). We should use the new parent coordinates here IMHO.

Ladicek: Actually, it looks like we _do_ relocate the parent. I'm not sure if that makes sense and if that works. Did you test this?

I'm thinking it probably will work, because the relocations don't require anything from the parent, so it doesn't matter the parent is itself a relocation, but I'm not sure.

Ladicek: This post https://www.lorenzobettini.it/2025/10/how-we-used-maven-relocation-for-xtend/ suggests that relocating a parent works.

It also suggests that all artifacts in one Maven Central deployment must share the group ID, which means that deploying our relocations (just like theirs) won't work and will have to be done separately.

manovotn: We do relocate the parent - I have taken this over from other projects/PRs I have seen using this approach.
That said, it is not required and we could skip that but it felt more natural to mimic the original structure.

> It also suggests that all artifacts in one Maven Central deployment must share the group ID, which means that deploying our relocations (just like theirs) won't work and will have to be done separately.

Hm, so that means I cannot have the relocation parent module be built automatically and we will need to run a separate release job (or script in the current one) to release the relocations. That's annoying :-/

Ladicek: It is annoying indeed.

Ladicek: Also, I really don't understand why we should relocate the parent, but I guess it's possible someone out there used it one way or another, so let's stay conservative and relocate it too.

manovotn: Theoretically, the release tweak should be as simple as re-running the same command but pointing to the relocation pom with `-f relocation/pom.xml`... hopefully.
But I am pretty sure the release plugin will be happy to prove me wrong :)

I suppose we could/should try to do some Alpha/Beta before we hit M2 so that we don't botch that one.

Ladicek: Agree we can do an Alpha2 to figure this out.

PR #915, opened 2025-11-18
add `@AutoClose`

### LINKED ISSUE #876: AutoCloseable beans
I've recently been trying out Spring for the first time, in order to get Hibernate Data Repositories working with it. To my surprise, Spring's support for dependency injection is on the whole quite inelegant and still, after all these years, lacking in type safety compared to CDI. It looks like at one point in history they basically just put a veneer of annotations over their old XML stuff, and then sorta left it at that. The annotations themselves are a real hodgepodge strewn about various packages, with quite bad naming, For example, their equivalent of `@Produces` is called `@Bean` and has seven weird and messy-looking members, most of them stringly-typed.

But there is so far _one single thing_ I've run into that I liked, and that is missing in CDI.

It seems that by default Spring will automatically call the `close()` method of any bean which implements `AutoCloseable`.  In CDI I would have two options:

1. declare a `@PreDestroy` method if it's a regular bean, or
2. write a `@Disposes` method if it's a producer method.

Option 1 is in general completely fine, and arguably `@PreDestroy` is easier than implementing `AutoCloseable`. But option 2 for producer methods is actually significantly worse, since it's very common that the object returned by a producer method is defined by a library or framework, and _already_ implements `AutoCloseable`. For example, Hibernate's `StatelessSession`, or whatever.

I think there should be some way to indicate to CDI that it should treat the `close()` method as a `@PreDestroy` callback. Ideally this would be automatic, but that would break backward compatibility, so I guess we need some sort of annotation for that. I'm not sure of the best way to expose that, but perhaps something like this is OK, if not perfect:

```java
@Produces @Closeable @TransactionScoped
StatelessSession statelessSession(SessionFactory factory) {
    return factory.openStatelessSesssion();
}
```

Does anyone have a better idea?

Ladicek:
Fixes #876

### DIFF
--- a/spec/src/main/asciidoc/core/spi_full.asciidoc
+++ b/spec/src/main/asciidoc/core/spi_full.asciidoc
@@ -36,6 +36,7 @@ public interface BeanAttributes<T> {
     public boolean isAlternative();
     public boolean isReserve();
     public boolean isEager();
+    public boolean isAutoClose();
 }
 ----
 
@@ -43,6 +44,7 @@ public interface BeanAttributes<T> {
 * `isAlternative()` must return `true` if the bean is an alternative, and `false` otherwise.
 * `isReserve()` must return `true` if the bean is a reserve, and `false` otherwise.
 * `isEager()` must return `true` if the bean is eagerly initialized, and `false` otherwise.
+* `isAutoClose()` must return `true` if the bean is auto-closeable, and `false` otherwise.
 
 The interface `jakarta.enterprise.inject.spi.Bean` defines everything the container needs to manage instances of a certain bean.
 
@@ -182,11 +184,11 @@ public interface Producer<T> {
 }
 ----
 
-For a `Producer` that represents a class:
+For a `Producer` that represents a managed bean:
 
-* `produce()` calls the constructor annotated `@Inject` if it exists, or the constructor with no parameters otherwise, as defined in <<instantiation>>, and returns the resulting instance. If the class has interceptors, `produce()` is responsible for building the interceptors and decorators of the instance.
+* `produce()` calls the constructor annotated `@Inject` if it exists, or the constructor with no parameters otherwise, as defined in <<instantiation>>, and returns the resulting instance. If the bean has interceptors or decorators, `produce()` is responsible for building the interceptors and decorators of the instance.
 The instance returned by `produce()` may be a proxy.
-* `dispose()` does nothing.
+* `dispose()` only does one thing: if the bean is auto-closeable and the class of the contextual instance of the bean implements `AutoCloseable`, the `close()` method is called on the instance.
 * `getInjectionPoints()` returns the set of `InjectionPoint` objects representing all injected fields, bean constructor parameters and initializer method parameters.
 
 
@@ -194,6 +196,7 @@ For a `Producer` that represents a producer method or field:
 
 * `produce()` calls the producer method on, or accesses the producer field of, a contextual instance of the bean that declares the producer method, as defined in <<producer_or_disposer_methods_invocation>>.
 * `dispose()` calls the disposer method, if any, on a contextual instance of the bean that declares the disposer method, as defined in <<producer_or_disposer_methods_invocation>>, or performs any additional required cleanup, if any, to destroy state associated with a resource.
+Then, if the bean is auto-closeable and the class of the contextual instance of the bean implements `AutoCloseable`, the `close()` method is called on the instance.
 * `getInjectionPoints()` returns the set of `InjectionPoint` objects representing all parameters of the producer method.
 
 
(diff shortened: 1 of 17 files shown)


Ladicek:
Rebased. I left the commits intact for further discussion, but still intend to squash them before merging.

Azquelt (approved):

Ladicek:
Squashed the commits. This is now ready.

spec/src/main/asciidoc/core/spi_full.asciidoc
```diff
@@ -36,13 +36,15 @@ public interface BeanAttributes<T> {
     public boolean isAlternative();
     public boolean isReserve();
     public boolean isEager();
+    public boolean isAutoClose();
```
manovotn, line 39: We should update the `BeanAttributesConfigurator` section similarly

manovotn (commented):
One silly question - if I have a managed bean using this feature, is the `close()` method call intercepted? I.e. is it a business method invocation? I think it's not but we might want to make that clear?

And another one - what about exceptions coming from `close()`? For example `@PreDestroy` methods have some guidance (and even TCK test) in this regard. At least their javadoc says:

> If the method throws an unchecked exception it is ignored by the container.

Should we have something similar (obviously accounting for checked exceptions in this case)? To make sure the instance is destroyed properly and so are any dependents? Or do we want to leave it as-is and just propagate it?

Ladicek:
> One silly question - if I have a managed bean using this feature, is the `close()` method call intercepted? I.e. is it a business method invocation? I think it's not but we might want to make that clear?

Great point! I admit I didn't think of this. We currently don't specify that this should be a business method invocation, but the method is called by the container and so the fallback in [7.2. Container invocations and interception](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#biz_method) applies and the invocation is _not_ a business method invocation.

But I guess it should be, like most other things invoked by the container after the bean has been constructed. Except, hmm, at the point we call `close()`, the bean has been destroyed (although its dependent instances were not), so intercepting doesn't make too much sense...

In any case, we should specify it explicitly. Any ideas?

Similarly: is it possible to _decorate_ `close()`? :-)

> And another one - what about exceptions coming from `close()`? For example `@PreDestroy` methods have some guidance (and even TCK test) in this regard. At least their javadoc says:
> 
> > If the method throws an unchecked exception it is ignored by the container.
> 
> Should we have something similar (obviously accounting for checked exceptions in this case)? To make sure the instance is destroyed properly and so are any dependents? Or do we want to leave it as-is and just propagate it?

`Contextual.destroy()` must not throw per the specification (see [6.1. The `Contextual` interface](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#contextual)) and `AutoCloseable.close()` declares `throws Exception`, so the only possible implementation is to catch any possible exceptions thrown by `close()` and swallow them (possibly logging a warning).

I explain this in the newly added method `Bean.destroy()` (which overrides `Contextual.destroy()`), but we might want to add a note to the spec somewhere...? Or at least add something to the `@AutoClose` javadoc.

manovotn:
> But I guess it should be, like most other things invoked by the container after the bean has been constructed. Except, hmm, at the point we call close(), the bean has been destroyed (although its dependent instances were not), so intercepting doesn't make too much sense...

Exactly.

> In any case, we should specify it explicitly. Any ideas?

I agree and I have no particular idea in mind, just wanted to bring it up for now :)
Perhaps a note in javadoc would suffice if it is otherwise implied by the 7.2 section.

EDIT: and we should also add a TCK test that will assert an interception does not take place in this case.

That said, I am not sure I understand how it is implied by the fallback in:

> We currently don't specify that this should be a business method invocation, but the method is called by the container and so the fallback in [7.2. Container invocations and interception](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#biz_method) applies and the invocation is not a business method invocation.

Do you mean that it is a lifecycle callback? 

> Similarly: is it possible to decorate close()? :-)

Same as with interception, so no :)

manovotn:
> I explain this in the newly added method Bean.destroy() (which overrides Contextual.destroy()), but we might want to add a note to the spec somewhere...? Or at least add something to the `@AutoClose` javadoc.

Ah, you're right, I missed that.
Though I agree that javadoc note would be welcome :+1:

Ladicek:
>That said, I am not sure I understand how it is implied by the fallback in:
> 
> > We currently don't specify that this should be a business method invocation, but the method is called by the container and so the fallback in [7.2. Container invocations and interception](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#biz_method) applies and the invocation is not a business method invocation.
> 
> Do you mean that it is a lifecycle callback?

No. I mean that section says:

> When the container invokes a method of a bean, the invocation may or may not be treated as a business method invocation:
>
> [list of situations when an invocation is/isn't business method invocation]
>
> A method invocation passes through method interceptors if, and only if, it is a business method invocation.
>
> Otherwise, the invocation is treated as a normal Java method call and is not intercepted by the container.

We don't specify that `close()` is a business method invocation, so it isn't business method invocation, therefore it isn't intercepted. But specifying whether it is/isn't a business method invocation _explicitly_ would be much better.

Azquelt:
Allowing a business method to be called after the pre-destroy lifecycle callback has fired feels like it may break assumptions in existing code. In particular, interceptors may do clean-up in a `@PreDestroy` method which would not allow an `@AroundInvoke` method to be called afterwards.

So I think the container calling `close()` should not be a business method invocation, as long as it occurs after `@PreDestroy`, and I agree we should state this explicitly.

Ladicek:
Added

> Invocations of `AutoCloseable.close()` on auto-closeable beans during destruction are not business method invocations.

Also added a note to the `@AutoClose` javadoc that exceptions thrown by `close()` during destruction are swallowed.

spec/src/main/asciidoc/core/spi_full.asciidoc
Ladicek: Good point! I actually forgot to mention `eager()` there, so I'll do that as well.

Ladicek: Done.

manovotn:
> Added
> 
> > Invocations of `AutoCloseable.close()` on auto-closeable beans during destruction are not business method invocations.
> 
> Also added a note to the `@AutoClose` javadoc that exceptions thrown by `close()` during destruction are swallowed.

Thank you :+1:

manovotn (approved):

Azquelt (approved):

PR #924, opened 2026-01-12
improve performance of common annotation literals

Ladicek:
The `AnnotationLiteral` class provides reflective implementation of all useful methods: `annotationType()`, `equals()`, `hashCode()` and `toString()`. This implementation is generic, but also doesn't perform nearly as well as a direct implementation.

This commit provides direct implementation of these methods for all annotation literals that exist in this project. The new implementation is compatible with the old one in nearly all aspects, except:

- `equals()` for annotations that have members no longer checks for equality of `annotationType()`; instead, it uses `instanceof`
- return values of `toString()` may be slightly different, because they are aligned with the standard `Annotation.toString()` behavior (on JDK 17)

Note that for memberless annotations, `instanceof` is not used; the code still tests equality of `annotationType()`, similarly to how `AnnotationLiteral` works. This is because for memberless annotations, it is relatively common for `AnnotationLiteral` implementations that don't implement the annotation interface to exist, because they can easily be created as anonymous classes. It is technically possible to create an anonymous subclass for annotations with members, but the `AnnotationLiteral.getMembers()` method throws in such case, so the generic implementations are guaranteed to not work. Note that creating an `AnnotationLiteral` that doesn't implement the annotation interface breaks the `Annotation` contract and so should not be relied upon.

It would be posisble to break compatibility more (e.g. the new implementations no longer have to extend `AnnotationLiteral`), but this commit intentionally stays conservative.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/context/ApplicationScoped.java
+++ b/api/src/main/java/jakarta/enterprise/context/ApplicationScoped.java
@@ -19,6 +19,7 @@ import static java.lang.annotation.ElementType.METHOD;
 import static java.lang.annotation.ElementType.TYPE;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
+import java.lang.annotation.Annotation;
 import java.lang.annotation.Documented;
 import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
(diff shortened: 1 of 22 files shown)


Ladicek:
I'm proposing this because we (Quarkus) recently found that there's certain overhead in using the generic implementations. I'm fine with this being rejected -- we'll just carry over this patch ourselves. But I figured everyone could benefit :-)

PR #927, opened 2026-01-20
introduce global scopes and improve application scope

### LINKED ISSUE #925: when is the application scope active?
When looking at https://github.com/jakartaee/cdi/pull/911, I realized the spec should probably say that the application scope is active during eager initialization. Then, I noticed that Core CDI don't say _anything_ about when the application scope is active (see https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#builtin_contexts).

CDI SE says (see https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#builtin_contexts_se):

> The application scope is active:
>
> - during any method invocation
>
> The application context is shared between all method invocations that execute within the same container.
>
> The application context is destroyed when the container is shut down.
>
> The payload of the event fired when the application context is initialized or destroyed is:
>
> - any `java.lang.Object`.

This makes perfect sense; it's actually how I've been thinking about the application scope all the time.

CDI EE says (see https://jakarta.ee/specifications/platform/11/jakarta-platform-spec-11.0#builtin_contexts_ee in the platform spec):

> The application scope is active:
>
> - during the `service()` method of any servlet in the web application, during the `doFilter()` method of any servlet filter and when the container calls any `ServletContextListener`, `HttpSessionListener`, `AsyncListener` or `ServletRequestListener`,
> - during any Jakarta EE web service invocation,
> - during any asynchronous invocation of an event observer,
> - during any remote method invocation of any enterprise bean, during any asynchronous method invocation of any enterprise bean, during any call to an enterprise bean timeout method and during message delivery to any message-driven enterprise bean,
> - when the disposer method or `@PreDestroy` callback of any bean with any normal scope other than `@ApplicationScoped` is called, and
> - during `@PostConstruct` callback of any bean.
>
> The application context is shared between all servlet requests, web service invocations, asynchronous invocation of an event observer, enterprise bean remote method invocations, enterprise bean asynchronous method invocations, enterprise bean timeouts and message deliveries to message-driven beans that execute within the same application. The application context is destroyed when the application is shut down.
>
> The payload of the event fired when the application context is initialized or destroyed is:
>
> - the `ServletContext` if the application is a web application deployed to a Servlet container, or
> - any `java.lang.Object` for other types of application.

This is a full list of conditions, some of them being only relevant to EE and some of them fairly general purpose.

It's been like that since CDI 2.0, when the spec was split into Core CDI, CDI SE and CDI EE. I _think_ we should probably rectify that.

These conditions from the EE part are fairly general and could be moved to Core CDI:

> - when the disposer method or `@PreDestroy` callback of any bean with any normal scope other than `@ApplicationScoped` is called, and
> - during `@PostConstruct` callback of any bean.

But they are also insufficient. I think at the very least, it should also say:

> during notification of an asynchronous observer method

or perhaps

> during notification of an observer method

With this addition, this would be relatively close to what Core CDI says about when the request context is active. (I don't really understand why the request context should be active "during `@PostConstruct` callback of any bean"; that provision was added in CDI 1.1, but I'm not sure any implementation actually follows it...)

Then, we could add

> during eager initialization [link to the spec chapter]

which I'd say should be there _in addition_ to "during `@PostConstruct`", because to be able to create contextual instances of eagerly initialized app-scoped beans, the app scope must be active.

Ladicek:
This commit introduces the notion of _global scopes_, for which exactly one context object exists and is active on all application threads since startup until shutdown. Extensions may not register context objects for built-in
global scopes.

The dependent context was previously specified to be "always active", which is not very precise. The dependent context is now global, for which a more precise specification exists (per above).

The application scope is now also global. This makes the specification simpler, more useful (other specifications may rely on the application scope being "always" active) and more aligned with existing implementations (they already prevent registering custom application contexts).

The specification never talks about the singleton context, but it is naturally also global.

Extensions are now specified to cause a deployment problem if registration of a custom context for a built-in global scope is attempted.

Further, build compatible extensions are now specified to allow registering custom contexts for the same set of scopes as portable extensions. Previously, registering custom contexts for built-in scopes was forbidden, which is too limiting and was never enforced (by the TCK and by implementations).

Fixes #925

### DIFF
--- a/api/src/main/java/jakarta/enterprise/context/ApplicationScoped.java
+++ b/api/src/main/java/jakarta/enterprise/context/ApplicationScoped.java
@@ -31,41 +31,16 @@ import jakarta.enterprise.util.AnnotationLiteral;
  * <p>
  * Specifies that a bean is application scoped.
  * </p>
+ *
  * <p>
- * While <code>ApplicationScoped</code> must be associated with the built-in application context required by the specification,
- * third-party extensions are allowed to also associate it with their own context. Behavior described below is only related to
- * the built-in application context.
+ * The application scope is global. Extensions are not allowed to register a custom context for it.
  * </p>
  *
  * <p>
- * The application scope is active:
- * </p>
- *
- * <ul>
- * <li>during the <code>service()</code> method of any servlet in the web application, during the <code>doFilter()</code> method
- * of any servlet filter and when the container calls any <code>ServletContextListener</code>, <code>HttpSessionListener</code>,
- * <code>AsyncListener</code> or <code>ServletRequestListener</code>,</li>
- * <li>during any Java EE web service invocation,</li>
- * <li>during any remote method invocation of any EJB, during any asynchronous method invocation of any EJB, during any call to
- * an EJB timeout method and during message delivery to any EJB message-driven bean,</li>
- * <li>when the disposer method or <code>@PreDestroy</code> callback of any bean with any normal scope other than
- * <code>@ApplicationScoped</code> is called, and</li>
- * <li>during <code>@PostConstruct</code> callback of any bean.</li>
- * </ul>
- *
- * <p>
- * The application context is shared between all servlet requests, web service invocations, EJB remote method invocations, EJB
- * asynchronous method invocations, EJB timeouts and message deliveries to message-driven beans that execute within the same
- * application.
- * </p>
- * <p>
- * The application context is destroyed when the application is shut down.
- * </p>
- *
- * <p>
- * An event with qualifier <code>@Initialized(ApplicationScoped.class)</code> is fired when the application context is
- * initialized and an event with qualifier <code>@Destroyed(ApplicationScoped.class)</code> when the application context is
- * destroyed. The event payload is:
+ * An event with qualifier {@code @Initialized(ApplicationScoped.class)} is fired after the application context
+ * is initialized. An event with qualifier {@code @BeforeDestroyed(ApplicationScoped.class)} is fired before
+ * the application context is destroyed. An event with qualifier {@code @Destroyed(ApplicationScoped.class)} is fired
+ * after the application context is destroyed. In all cases, the event payload is:
  * </p>
  *
  * <ul>
--- a/spec/src/main/asciidoc/core/scopescontexts.asciidoc
+++ b/spec/src/main/asciidoc/core/scopescontexts.asciidoc
@@ -166,6 +166,17 @@ All pseudo-scopes must be explicitly declared `@Scope`, to indicate to the conta
 
 All scopes defined by this specification, except for the `@Dependent` pseudo-scope, are normal scopes.
 
+[[global_scope]]
+
+=== Global scopes
+
+Several scopes defined by this specification are _global_.
+A global scope has exactly one context object, which is:
+
+* initialized once, when the container starts up, before the event with qualifier `@Initialized(ApplicationScoped.class)` is fired,
+* destroyed once, when the container shuts down, after the event with qualifier `@BeforeDestroyed(ApplicationScoped.class)` is fired,
+* active on all application threads from the point the context object is initialized until it is destroyed.
+
 [[dependent_context]]
 
 === Dependent pseudo-scope
@@ -184,7 +195,7 @@ Every invocation of the `get()` operation of the `Context` object for the `@Depe
 
 Every invocation of the `get()` operation of the `Context` object for the `@Dependent` scope with no `CreationalContext` returns a null value.
 
-The `@Dependent` scope is always active.
+The `@Dependent` scope is global.
 
 [[dependent_objects]]
 
@@ -235,11 +246,11 @@ If there is exactly one active instance of `Context` associated with the scope t
 
 [[activating_builtin_contexts]]
 
-==== Activating Built In Contexts
+==== Activating Built-in Contexts
 
-Certain built in contexts support the ability to be activated and deactivated.  This allows developers to control built-in contexts in ways that they could also manage custom built contexts.
+Certain built-in contexts support the ability to be activated and deactivated.  This allows developers to control built-in contexts in ways that they could also manage custom built contexts.
 
-When activating and deactivating built in contexts, it is important to realize that they can only be activated if not already active within a given thread.
+When activating and deactivating built-in contexts, it is important to realize that they can only be activated if not already active within a given thread.
 
 [[activating_request_context]]
 
@@ -247,7 +258,7 @@ When activating and deactivating built in contexts, it is important to realize t
 
 Request contexts can be managed either programmatically or via interceptor.
 
-To programmatically manage request contexts, the container provides a built in bean that is `@Dependent` scoped and of type `RequestContextController` that allows you to activate and deactivate a request context on the current thread.  The object should be considered stateful, invoking the same instance on different threads may not work properly, non-portable behavior may occur.
+To programmatically manage request contexts, the container provides a built-in bean that is `@Dependent` scoped and of type `RequestContextController` that allows you to activate and deactivate a request context on the current thread.  The object should be considered stateful, invoking the same instance on different threads may not work properly, non-portable behavior may occur.
 
 [source, java]
 ----
@@ -390,11 +401,17 @@ The request context is destroyed:
 
 ==== Application context lifecycle
 
+The application scope is global.
+
 The _application context_ is provided by a built-in context object for the built-in scope type `@ApplicationScoped`.
 
 An event with qualifier `@Initialized(ApplicationScoped.class)` is synchronously fired when the application context is initialized.
 An event with qualifier `@BeforeDestroyed(ApplicationScoped.class)` is synchronously fired when the application context is about to be destroyed, i.e. before the actual destruction.
 An event with qualifier `@Destroyed(ApplicationScoped.class)` is synchronously fired when the application context is destroyed, i.e. after the actual destruction.
+
+The payload of the event fired when the application context is initialized or destroyed is:
+
+* any `java.lang.Object`.
 
 [[custom_contexts]]
 
@@ -410,4 +427,6 @@ where `X` is the scope type associated with the context.
 
 A suitable event payload should be chosen.
 
-Build compatible extensions may define custom context classes for custom scopes, but they may not define custom context classes for built-in scopes.
+Build compatible extensions may define custom context classes for non-global built-in scopes and for custom scopes.
+For example, a remoting framework might provide a request context object for the built-in request scope.
+If an attempt to register a custom context class for a global built-in scope is made, deployment problem occurs.
--- a/spec/src/main/asciidoc/javase/scopescontext_se.asciidoc
+++ b/spec/src/main/asciidoc/javase/scopescontext_se.asciidoc
@@ -11,27 +11,10 @@ SPDX-License-Identifier: Apache-2.0
 
 == Scopes and contexts in Java SE
 
-[[builtin_contexts_se]]
+[[global_scope_se]]
 
-=== Context management for built-in scopes in Java SE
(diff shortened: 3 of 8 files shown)


Ladicek:
When we merge this, I also need to submit a PR to the platform spec to update https://github.com/jakartaee/platform/blob/main/specification/src/main/asciidoc/platform/cdi-ee-spec/scopescontext_ee.adoc#context-management-for-built-in-scopes-in-jakarta-ee

Ladicek:
For the record, what I propose to submit to the Jakarta Platform spec is at https://github.com/Ladicek/jakartaee-platform/commits/improve-app-scope/

spec/src/main/asciidoc/javase/scopescontext_se.asciidoc
```diff
@@ -15,23 +15,14 @@ SPDX-License-Identifier: Apache-2.0
...
-When running in Java SE the container must extend the rules defined in <<application_context>> and is also required to ensure the following rules.
-
-The application scope is active:
-
-* during any method invocation
+When running in Java SE, the container must extend the rules defined in <<application_context>>, except of the two that are overridden by the following rules.
 
 The application context is shared between all method invocations that execute within the same container.
 
 The application context is destroyed when the container is shut down.
```
manovotn: I am not sure I can see the actual difference between what you stated in `spec/src/main/asciidoc/core/scopescontexts.asciidoc` and here:

> The application context is destroyed when the application is shut down.
> The application context is destroyed when the container is shut down.

Or rather, I am not sure the difference is worth spelling out here.
To me, in SE, the two are intertwined but perhaps I am missing something?

Ladicek: So I'm not sure if you can boot 2 containers in SE concurrently, but I guess it should be possible to boot one container, shut it down, then boot another. Which makes that difference important.

manovotn: I see, that makes sense :+1: 
FTR, I think in CDI as such you won't have the ability to bootstrap multiple SE containers. In Weld you can ([test](https://github.com/weld/core/blob/master/environments/se/core/src/test/java/org/jboss/weld/environment/se/test/container/isolation/ContainerIsolationTest.java)) but then doing anything like `CDI.current()` is a highway to hell :)

spec/src/main/asciidoc/core/scopescontexts.asciidoc
```diff
@@ -396,6 +396,16 @@ An event with qualifier `@Initialized(ApplicationScoped.class)` is synchronously
 An event with qualifier `@BeforeDestroyed(ApplicationScoped.class)` is synchronously fired when the application context is about to be destroyed, i.e. before the actual destruction.
 An event with qualifier `@Destroyed(ApplicationScoped.class)` is synchronously fired when the application context is destroyed, i.e. after the actual destruction.
 
+The application context is always active, since the event with qualifier `@Initialized(ApplicationScoped.class)` is fired and until the event with qualifier `@Destroyed(ApplicationScoped.class)` is fired.
```
manovotn: Personally I would rather frame the usability of the context by two events in which you can still use the context (`@Initialized` ->  `@BeforeDestroyed`) but I suppose I can live with `@Destroyed` as well :)

Ladicek: During `@BeforeDestroyed` notification, the application context is still guaranteed to be active. So I don't think you can say "until `@BeforeDestroyed` is fired", because that might be read as "during `@BeforeDestroyed`, the context is not active".

I think we could make it more precise by saying:

> The application context is always active, starting after the event with qualifier `@Initialized(ApplicationScoped.class)` is fired and ending before the event with qualifier `@Destroyed(ApplicationScoped.class)` is fired.

WDYT?

Ladicek: Though to be honest, we simply say

> The `@Dependent` scope is always active.

and don't bother specifying when it starts and when it ends, even though there clearly are such points in time.

spec/src/main/asciidoc/javase/scopescontext_se.asciidoc
Ladicek: > I think in CDI as such you won't have the ability to bootstrap multiple SE containers.

So the only thing the specification says about the possibility of having multiple containers is:

> An implementation does not need to support multiple calls of `SeContainerInitializer.initialize()` method when the `SeContainer` is running.

This is not super clear (is an implementation required to support _one_ call of `SeContainerInitializer.initialize()` when the `SeContainer` is running?), but it seems obvious that it's supposed to mean that the specification doesn't guarantee that multiple SE containers may be running at the same time.

This however in no way prevents the following:

```java
try (SeContainer container = SeContainerInitializer.newInstance().initialize()) {
    ...
}

try (SeContainer container = SeContainerInitializer.newInstance().initialize()) {
    ...
}
```

That is, you can have multiple containers running inside a single application, just not at the same time.

spec/src/main/asciidoc/core/scopescontexts.asciidoc
manovotn: > During `@BeforeDestroyed` notification, the application context is still guaranteed to be active. So I don't think you can say "until `@BeforeDestroyed` is fired", because that might be read as "during `@BeforeDestroyed`, the context is not active".
> 
> I think we could make it more precise by saying:
> 
> > The application context is always active, starting after the event with qualifier `@Initialized(ApplicationScoped.class)` is fired and ending before the event with qualifier `@Destroyed(ApplicationScoped.class)` is fired.
> 
> WDYT?

I was thinking something along the lines of:

> The application context is always active beginning with firing of event with qualifier `@Initialized(ApplicationScoped.class)` up to and including the notification of `@BeforeDestroyed(ApplicationScoped.class)` observers.

manovotn: > Though to be honest, we simply say
> 
> > The `@Dependent` scope is always active.
> 
> and don't bother specifying when it starts and when it ends, even though there clearly are such points in time.

The `@Dependent` scope doesn't fire the init/destroy events though. At least I think it doesn't as it won't make sense to observe those anyway 🤔

Ladicek: Correct, but that doesn't make a difference. The singleton context doesn't fire those events either, even though it would be meaningful to observe them. (But then, which one should come sooner? :-) ) Granted, we don't say anything about the singleton scope in the CDI spec, probably for the better :-)

Anyway, I'll reword the text slightly per above.

Ladicek: I'm leaning towards this:

> The application context is always active, starting before the event with qualifier `@Initialized(ApplicationScoped.class)` is fired and ending after the event with qualifier `@BeforeDestroyed(ApplicationScoped.class)` is fired.

This is because during notification of `@Initialized(AppScoped.class)`, the app scope is already active, so it must start before. For symmetry, I'm saying it ends after `@BeforeDestroyed(AppScoped.class)`. I could also say it ends "before `@Destroyed(AppScoped.class)`", but I don't like it as much and one could argue it would be somewhat less precise :-)

manovotn: I am sorry for nit picking but re-reading this section of the specification again, I am wondering if we aren't just re-stating the same thing.
The very few sentences prior to your addition already explain the events that mark the boundaries of app context lifecycle 🤔 

What I mean is we already have this:

```
==== Application context lifecycle

The _application context_ is provided by a built-in context object for the built-in scope type `@ApplicationScoped`.

An event with qualifier `@Initialized(ApplicationScoped.class)` is synchronously fired when the application context is initialized.
An event with qualifier `@BeforeDestroyed(ApplicationScoped.class)` is synchronously fired when the application context is about to be destroyed, i.e. before the actual destruction.
An event with qualifier `@Destroyed(ApplicationScoped.class)` is synchronously fired when the application context is destroyed, i.e. after the actual destruction.
```

Ladicek: So you're fine with this just saying

> The application scope is always active.

? :-)

That would be fine to me.

manovotn: > So you're fine with this just saying
> 
> > The application scope is always active.
> 
> ? :-)
> 
> That would be fine to me.

Yes :)

api/src/main/java/jakarta/enterprise/context/ApplicationScoped.java
```diff
@@ -30,33 +30,17 @@
...
- * <p>
- * While <code>ApplicationScoped</code> must be associated with the built-in application context required by the specification,
- * third-party extensions are allowed to also associate it with their own context. Behavior described below is only related to
- * the built-in application context.
- * </p>
  *
  * <p>
- * The application scope is active:
+ * The application scope is always active, since the event with qualifier {@code @Initialized(ApplicationScoped.class)}
+ * is fired and until the event with qualifier {@code @Destroyed(ApplicationScoped.class)} is fired.
```
Azquelt: Use of "since" here is wrong ("since" is usually for something that started in the past and is ongoing).

It also seems odd to say that the scope is always active and then place limits on when it's active.

```suggestion
 * The application scope is active from when the event with qualifier {@code @Initialized(ApplicationScoped.class)}
 * is fired until the event with qualifier {@code @Destroyed(ApplicationScoped.class)} is fired.
```

I'm still thinking about the change as a whole, and I'm not totally happy with it yet, but I think this suggestion to fix the wording here is necessary if we do keep the change.

manovotn: > It also seems odd to say that the scope is always active and then place limits on when it's active.

Well, yes, I agree. However, we need to somehow set the limits on when you can start and stop using it while at the same time explaining that within those bounds it is just "always active".
I will gladly accept help of a native speaker in that regard :)

Also note the discussion we had earlier - https://github.com/jakartaee/cdi/pull/927#discussion_r2712958828
I think we need to somehow merge that into the existing text...

spec/src/main/asciidoc/core/scopescontexts.asciidoc
Ladicek: Done.

api/src/main/java/jakarta/enterprise/context/ApplicationScoped.java
Ladicek: I just got rid of this complexity as agreed with @manovotn above :-) This part of javadoc now reads:

> The application context is always active. It is destroyed when the application is shut down.

manovotn: @Azquelt what is your take on the current state of the PR? Does it look better this way?

Azquelt: I still think "always" is too strong a requirement. At the very least the scope can only be active while the singular application context is active, and I would expect that some lifecycle events would occur before that.

In a tutorial or something, I think saying it's always active would be fine, but I'm not sure we can get away with that in the spec.

My second issue is that I'm still bothered by the fact that in an application server, there will probably be times where code is executing that's not part of an application and therefore there's no active application context on that thread, and therefore the scope is not active. Normally user code should not be running at this point, perhaps that's invisible to the user, but the reality of different applications being active on different threads makes the idea of saying that the scope is always active seem wrong.

I prefer the existing definition of when the application scope is active for Jakarta EE.

manovotn (approved):
LGTM, I have no objections 👍

arjantijms:
I'm late to the party, but +1 for this, and especially good that specific references to Servlet and EJB technologies could now be removed. Good work team!

PR #931, opened 2026-01-28
improve specification of `InjectionPoint`

### LINKED ISSUE #779: Define how `InjectionPoint` works when a bean is obtained via CDI.current()
The [InjectionPoint Javadoc](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/enterprise/inject/spi/injectionpoint) says this:

> Provides access to metadata about an injection point. May represent an [injected field](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/inject/inject) or a parameter of a [bean constructor](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/inject/inject), [initializer method](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/inject/inject), [producer method](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/enterprise/inject/produces), [disposer method](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/enterprise/inject/disposes) or [observer method](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/enterprise/event/observes).
> 
> If the injection point is a dynamically selected reference obtained then the metadata obtain reflects the injection point of the [Instance](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/enterprise/inject/instance), with the required type and any additional required qualifiers defined by [Instance.select()](https://jakarta.ee/specifications/platform/10/apidocs/jakarta/enterprise/inject/instance).

This doesn't cover the case where a bean is looked up using `CDI.current().select(...).get()` because there is no injection point for  the Instance.

For example if I have a qualifier:
```java
@Qualifier
@Retention(RUNTIME)
@Target({ TYPE, FIELD, METHOD, PARAMETER, ANNOTATION_TYPE })
public interface @MyQualifier {
    @Nonbinding String value()
    public class MyQualifierLiteral ...
}
```

and a managed bean class:

```java
@Dependent
@MyQualifier("")
public class MyBean {

    @Inject
    InjectionPoint ip;

    public String getQualifierValue() {
        return ip.getQualifiers().stream()
                        .filter(a -> a.annotationType() == MyQualifier.class)
                        .map(MyQualifier.class::cast)
                        .map(MyQualifier::value)
                        .findFirst()
                        .orElseThrow(() -> new IllegalStateException("No qualifier found"));
    }
}
```

I can inject this bean like this and get its qualifiers:
```java
@Inject @MyQualifier("foo") private MyBean fooBean;
@Inject @MyQualifier("bar") private Instance<MyBean> barBeanInstance;
@Inject @Any private Instance<Object> objectInstance;

void test() {
    assert fooBean.getQualifierValue().equals("foo");
    assert barBeanInstance.get().getQualifierValue().equals("bar");
    assert objectInstance.select(MyBean.class, new MyQualifier.MyQualifierLiteral("baz")).get().getQualifierValue().equals("baz");
}
```

However, I don't think it's defined what happens when I do this:
```java
void test() {
    MyBean bean = CDI.current().select(MyBean.class, new MyQualifier.MyQualifierLiteral("qux")).get();
    assert bean.getQualifierValue().equals("qux");
}
```

Would it make sense to define how the `InjectionPoint` should work here? I don't think there's any sensible value for `getAnnotated()`, `getBean()` or `getMember()`, but `getQualifiers()` and `getType()` would still be helpful in this scenario.

This came to my attention because the Jakarta Batch spec gave an example of looking up a bean like this but getting the injection point isn't supported on open web beans for a bean obtained via `CDI.select()`: eclipse-ee4j/batch-tck#69

Ladicek:
The specification of `InjectionPoint` contains a small number of holes that are arguably corner cases, but they should be specified nevertheless.

The most important one is dynamically obtained instance (`Instance`) that is not injected. This can happen through several means:

- `BeanContainer.createInstance()`
- `CDI.current()`
- `SeContainer`

The specification didn't say what multiple methods should return in this case; it didn't even explicitly say that an `InjectionPoint` should exist and indeed it doesn't on one existing implementation. With this commit, the behavior is clear: an `InjectionPoint` must exist and all methods besides `getType()` and `getQualifiers()` return `null` (or `false`).

Further, it is not clear what the `getBean()` method should return in case of a synthetic bean. Such injection points should, in portable extensions, be created using `BeanManager.createInjectionPoint()`, but these methods do not take a `Bean` and cannot figure out on their own which bean is the injection point created for. So on all existing implementations that support portable extensions, this method returns `null`. This is now explicitly allowed.

Finally, this commit allows synthetic beans to have injection points that do not correspond to a field or parameter. Currently, such injection points should never exist, because injection points of synthetic beans should be created using `BeanManager.createInjectionPoint()`, but it is not hard to imagine people implementing `InjectionPoint` manually and returning `null` from `getMember()` or `getAnnotated()`. We also plan to introduce specifying injection points for synthetic beans in the build compatible extensions API, where one is supposed to only specify the required type and required qualifiers and it is entirely conceivable that no field or parameter can possibly be determined upfront. So an `InjectionPoint` of a synthetic bean is now explicitly allowed to return `null` from the `getMember()` and `getAnnotated()` methods.

Fixes #779
Related to #833

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/se/SeContainer.java
+++ b/api/src/main/java/jakarta/enterprise/inject/se/SeContainer.java
@@ -23,7 +23,9 @@ import jakarta.enterprise.inject.spi.BeanManager;
  *
  * <p>
  * SeContainer implements {@link Instance} and therefore might be used to perform programmatic lookup.
- * If no qualifier is passed to {@link #select} method, the <code>@Default</code> qualifier is assumed.
+ * Initially, the required type of the {@code Instance} is {@code java.lang.Object} and there
+ * are no required qualifiers. If no qualifier is passed to {@code Instance.select()}, there
+ * is one required qualifier: {@code @Default}.
  * </p>
  *
  * <p>
--- a/spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
+++ b/spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
@@ -61,7 +61,8 @@ An object may obtain a reference to the current container by calling `CDI.curren
 If methods on `CDI` are called at any other time, non-portable behavior results.
 
 `CDI` implements `jakarta.enterprise.inject.Instance` and therefore might be used to perform programmatic lookup as defined in <<dynamic_lookup>>.
-If no qualifier is passed to `CDI.select()` method, the `@Default` qualifier is assumed.
+Initially, the required type of the `Instance` is `java.lang.Object` and there are no required qualifiers.
+If no qualifier is passed to `Instance.select()`, there is one required qualifier: `@Default`.
 
 When `CDI.current()` is called, `getCDI()` method is called on `jakarta.enterprise.inject.spi.CDIProvider`.
 
@@ -268,7 +269,8 @@ Instance<Object> createInstance();
 
 Instances of dependent scoped beans obtained with this `Instance` object must be explicitly released by calling `Instance.destroy()` method.
 
-If no qualifier is passed to `Instance.select()` method, the `@Default` qualifier is assumed.
+Initially, the required type of the `Instance` is `java.lang.Object` and there are no required qualifiers.
+If no qualifier is passed to `Instance.select()`, there is one required qualifier: `@Default`.
 
 [[bm_bean_event_assignability]]
 
--- a/api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/BeanContainer.java
@@ -258,7 +258,9 @@ public interface BeanContainer {
      * Instances of dependent scoped beans obtained with this <code>Instance</code> must be explicitly destroyed by calling
      * {@link Instance#destroy(Object)}
      * <p>
-     * If no qualifier is passed to {@link Instance#select} method, the <code>@Default</code> qualifier is assumed.
+     * Initially, the required type of the {@code Instance} is {@code java.lang.Object} and there
+     * are no required qualifiers. If no qualifier is passed to {@code Instance.select()}, there
+     * is one required qualifier: {@code @Default}.
      *
      * @return an {@link Instance} object to request beans instances
      * @throws IllegalStateException if called during application initialization, before the {@link AfterDeploymentValidation}
--- a/api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
@@ -28,7 +28,9 @@ import jakarta.enterprise.inject.Instance;
  *
  * <p>
  * CDI implements {@link Instance} and therefore might be used to perform programmatic lookup.
- * If no qualifier is passed to {@link #select} method, the <code>@Default</code> qualifier is assumed.
+ * Initially, the required type of the {@code Instance} is {@code java.lang.Object} and there
+ * are no required qualifiers. If no qualifier is passed to {@code Instance.select()}, there
+ * is one required qualifier: {@code @Default}.
  * </p>
  *
  *
--- a/spec/src/main/asciidoc/javase/bootstrap_se.asciidoc
+++ b/spec/src/main/asciidoc/javase/bootstrap_se.asciidoc
@@ -98,7 +98,7 @@ The `jakarta.enterprise.inject.se.SeContainer` interface provides access to the 
 
 [source, java]
 ----
-public interface SeContainer extends Instance<Object>,AutoCloseable {
+public interface SeContainer extends Instance<Object>, AutoCloseable {
     void close();
     boolean isRunning();
     BeanManager getBeanManager();
@@ -113,7 +113,8 @@ If it is called and the container was already shutdown, it throws an `IllegalSta
 If it is called and the container was already shutdown, it throws an `IllegalStateException`.
 
 `SeContainer` implements `jakarta.enterprise.inject.Instance` and therefore might be used to perform programmatic lookup as defined in <<dynamic_lookup>>.
-If no qualifier is passed to `SeContainer.select()` method, the `@Default` qualifier is assumed.
+Initially, the required type of the `Instance` is `java.lang.Object` and there are no required qualifiers.
+If no qualifier is passed to `Instance.select()`, there is one required qualifier: `@Default`.
 
 If any `Instance.select()` method is called and the container was already shutdown, the `IllegalStateException` is thrown.
 


Ladicek:
I will submit a TCK PR in the next few days, but the spec text should be done IMHO.

Ladicek:
For the record, I'm not entirely sure if this perhaps doesn't somehow interact with passivation. I looked at the relevant part of the spec and what implementations do and didn't find anything obvious, but I admittedly don't know much about the topic, so I might be missing something.

api/src/main/java/jakarta/enterprise/inject/se/SeContainer.java
```diff
@@ -23,7 +23,9 @@
  *
  * <p>
  * SeContainer implements {@link Instance} and therefore might be used to perform programmatic lookup.
- * If no qualifier is passed to {@link #select} method, the <code>@Default</code> qualifier is assumed.
+ * An initial required type of the {@code Instance} is {@code java.lang.Object} and there are
+ * initially no required qualifiers. If no qualifier is passed to {@code Instance.select()},
+ * there is one required qualifier: {@code @Default}.
```
Azquelt:
```suggestion
 * The initial required type of the {@code Instance} is {@code java.lang.Object} and there are
 * initially no required qualifiers. If no qualifier is passed to {@code Instance.select()},
 * there is one required qualifier: {@code @Default}.
```

Also in other places where this text is used.

api/src/main/java/jakarta/enterprise/inject/spi/EventMetadata.java
```diff
@@ -47,16 +47,14 @@ public interface EventMetadata {
...
      */
     public InjectionPoint getInjectionPoint();
 
     /**
-     * Get the type representing runtime class of the event object with type variables resolved.
-     *
-     *
      * @return the runtime type of the event object
      */
     public Type getType();
```
Azquelt, line 63: Is this a change correct?

The original text isn't great, but I would have guessed it meant the runtime type with any type variables inferred from the specified type of the event (as discussed in #884).

api/src/main/java/jakarta/enterprise/inject/spi/InjectionPoint.java
```diff
@@ -65,58 +66,68 @@
...
+     * whether the injection point is a field or a constructor/method parameter.
+     * <p>
+     * If the injection point belongs to a synthetic bean, may return {@code null}.
+     * If the injection point belongs to a dynamically obtained instance where the {@code Instance}
+     * is not injected, returns {@code null}.
+     * <p>
+     * May always return {@code null} when the container only supports CDI Lite.
      *
-     * @return an {@code AnnotatedField} or {@code AnnotatedParameter}
+     * @return the annotated member, may be {@code null}
```
Azquelt, line 116: I was initially surprised by this change, until I realised that this is what the spec [already says](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#injection_point).

api/src/main/java/jakarta/enterprise/inject/se/SeContainer.java
Ladicek: Makes sense, will change.

api/src/main/java/jakarta/enterprise/inject/spi/EventMetadata.java
Ladicek: Ah, I forgot we didn't merge any solution to #884 and #886. Too bad. I will resurrect the javadoc here, although in my eyes, it is really bad.

api/src/main/java/jakarta/enterprise/inject/spi/InjectionPoint.java
```diff
@@ -65,58 +66,68 @@
...
      */
     public Annotated getAnnotated();
 
     /**
-     * Determines if the injection point is a decorator delegate injection point.
+     * Returns whether the injection point is a decorator {@link jakarta.decorator.Delegate @Delegate} injection point.
```
manovotn, line 121: Nitpicker note - I am not a native speaker but `Returns whether` doesn't sound quite right.
I'd probably settle with the usual `Returns true if X, false otherwise`
Or, closer to your wording, `Determines whether ...` 🤷

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
```diff
@@ -61,7 +61,8 @@ An object may obtain a reference to the current container by calling `CDI.curren
 If methods on `CDI` are called at any other time, non-portable behavior results.
 
 `CDI` implements `jakarta.enterprise.inject.Instance` and therefore might be used to perform programmatic lookup as defined in <<dynamic_lookup>>.
-If no qualifier is passed to `CDI.select()` method, the `@Default` qualifier is assumed.
+An initial required type of the `Instance` is `java.lang.Object` and there are initially no required qualifiers.
```
manovotn:
```suggestion
Initially, the required type is `java.lang.Object` and there are no required qualifiers.
```

Just trying to avoid saying "initial" twice in one sentence.

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
```diff
@@ -487,8 +487,9 @@ The interface `jakarta.enterprise.inject.spi.InjectionPoint` provides access to
 An instance of `InjectionPoint` may represent:
 
 * an injected field or a parameter of a bean constructor, initializer method, producer method, disposer method or observer method, or
+* an injection into a synthetic bean (where an injected field or parameter does not necessarily have to exist), or
```
manovotn, line 490: I take it this is mainly for the sake of BCE?

Because in portable extensions it doesn't really make much sense to declare an injection point. They will only ever be used for validation; never an actual injection. Any "injection" is done via dynamic resolution later on.

To be frank I am thinking if there is any point in being able to declare IPs for beans added via portable extensions whatsoever. But I guess that's a discussion for another time and place.

spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
```diff
@@ -165,12 +165,10 @@ Therefore, the container must use a client proxy whenever a bean with normal sco
...
-If the injection point represents a dynamically obtained instance, then the `getAnnotated()` method returns an instance of `jakarta.enterprise.inject.spi.AnnotatedField` or `jakarta.enterprise.inject.spi.AnnotatedParameter` representing the `Instance` injection point, depending upon whether the injection point is an injected field or a constructor/method parameter.
 * The `isDelegate()` method returns `true` if the injection point is a decorator delegate injection point, and `false` otherwise.
-If the injection point represents a dynamically obtained instance then `isDelegate()` returns false.
-
-If the injection point represents a dynamically obtained instance then the `isTransient()` method returns `true` if the `Instance` injection point is a transient field, and `false` otherwise. If this injection point is declared as transient, after bean's passivation, the value will not be restored. Instance<> injection point is the preferred approach.
+If the injection point represents an injection into a synthetic bean or a dynamically obtained instance, the `isDelegate()` method returns `false`.
```
manovotn, line 169: In portable extension, you can technically take any existing injection point and register it with the synthetic bean (for no apparent reason but to shoot yourself in the foot) in which case this claim won't hold.

I am not sure we care to that extent though, just putting it up here as it crossed my mind :)

api/src/main/java/jakarta/enterprise/inject/spi/EventMetadata.java
manovotn: I agree, we need to keep the original wording...for now at least

api/src/main/java/jakarta/enterprise/inject/spi/InjectionPoint.java
Azquelt: I think "returns whether" is correct. It does sound a bit weird, it's the sort of thing I only expect to see in Javadoc.

I see both patterns used in the standard classes, I'd be happy with either.
- https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/Collection.html#isEmpty()
- https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/stream/Stream.html#anyMatch(java.util.function.Predicate)

manovotn: I see, then it is just my lack of familiarity with this collocation and we can keep it as-is :)

Ladicek: I find "returns whether" appropriate in nearly all javadocs of `boolean`-returning methods, so... :-)

spec/src/main/asciidoc/core/injectionandresolution.asciidoc
Ladicek: This is indeed mainly for BCEs, but PEs also support adding injection points and given BCE implemented in CDI Full are likely based on PEs, this applies to PEs as well.

spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
Ladicek: You are a very sneaky guy! :-D

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
Ladicek: OK, will use this.

api/src/main/java/jakarta/enterprise/inject/se/SeContainer.java
Ladicek: Done.

api/src/main/java/jakarta/enterprise/inject/spi/EventMetadata.java
Ladicek: Done.

spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
Ladicek: Done.

spec/src/main/asciidoc/core/injectionandresolution_full.asciidoc
Ladicek: Also, you would actually be violating the specification if you did that. https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#delegate_attribute says:

> If a bean class that is not a decorator has an injection point annotated `@Delegate`, the container automatically detects the problem and treats it as a definition error.

You could argue that a synthetic bean does not have a bean class and the injection point is not really "annotated", but it seems clear to me that delegate injection points may only occur in decorators and not elsewhere, and the situation you describe is clearly in violation of the spirit of the spec, if not perhaps of the text.

manovotn: I'd say it's just the grey area consisting of all the actions you can take to sabotage yourself through extensions.
Let me resolve the conversation and leave it at that :)

PR #937, opened 2026-02-04
allow specifying parameterized types in @Registration

### LINKED ISSUE #936: allow `@Registration.types()` to specify parameterized bean types
**Is your feature request related to a problem? Please describe.**

The `@Registration` annotation from Build Compatible Extensions provides means of specifying bean types (or observed event types) on which the annotated method should operate. Since `@Registration` is an annotation, the types can only be `Class<?>` literals, so there's no direct support for parameterized types.

It is possible to work around that by using `@Registration(types = Object.class)` and later check `bean.types()` against a parameterized type created using `Types`, but a straightforward solution would be preferrable.

**Describe the solution you'd like**

We can easily achieve that by adding a special case: if the class type present in `@Registration.types` has a direct superclass type whose erasure is `jakarta.enterprise.util.TypeLiteral`, we don't use the class type itself, but the type argument to the `TypeLiteral`. This looks horrible when written like this, but the actual code is relatively pleasant:

```java
@Registration(types = FooOfBar.class)
public void registration(BeanInfo bean) {
    ...
}

static class FooOfBar extends TypeLiteral<Foo<Bar>> {
}
```

There are 2 edge cases: `FooOfBar` could extend a raw type (`extends TypeLiteral`) or the argument to `TypeLiteral` could contain type variables (`class FooOfBar<T> extends TypeLiteral<Foo<T>>`). I propose we treat them as definition errors.

**Describe alternatives you've considered**

N/A

**Additional context**

This was originally reported in Quarkus: https://github.com/quarkusio/quarkus/issues/52352

I have a specification draft and a prototype implementation in ArC. They are both relatively small changes, so it feels fine to stuff it into CDI 5.0.

Ladicek:
This is done by specifying a class whose direct superclass type is a parameterization of `TypeLiteral`. In this case, the type argument to `TypeLiteral` is an expected type, not the class itself.

Fixes #936

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Registration.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Registration.java
@@ -60,7 +60,26 @@ import java.lang.annotation.Target;
 @Retention(RetentionPolicy.RUNTIME)
 public @interface Registration {
     /**
-     * Defines the set of <em>expected types</em>.
+     * Defines the set of <em>expected types</em>. Any type is directly included in the set of expected types,
+     * unless it is a class type whose direct superclass type's erasure is {@link jakarta.enterprise.util.TypeLiteral}.
+     * In that case, the type argument to {@code TypeLiteral} is included in the set of expected types.
+     * <p>
+     * If the type argument to {@code TypeLiteral} contains type variables, the container treats it as a definition error.
+     * (This restriction shall be lifted in the future, because {@code @Dependent} beans may have type variables
+     * in their bean types. What it means for a set of bean types to contain an expected type is not immediately
+     * obvious in case type variables are present. More specification work is required.)
+     * <p>
+     * If the direct superclass type is a raw {@code TypeLiteral}, the container treats it as a definition error.
+     * <p>
+     * For example, to observe all beans whose set of bean types contains the {@code CharSequence} type,
+     * you would write {@code @Registration(types = CharSequence.class)}. To observe all beans whose set of bean types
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/Registration.java
```diff
@@ -60,7 +60,10 @@
 @Retention(RetentionPolicy.RUNTIME)
 public @interface Registration {
     /**
-     * Defines the set of <em>expected types</em>.
+     * Defines the set of <em>expected types</em>. Any type is directly included in the set of expected types,
+     * unless it is a class type whose direct superclass type's erasure is {@link jakarta.enterprise.util.TypeLiteral}.
+     * In that case, the type argument to {@code TypeLiteral} is included in the set of expected types.
+     * If the direct superclass type is a raw {@code TypeLiteral}, the container treats it as a definition error.
```
Azquelt, line 72:
```suggestion
     * If the direct superclass type is a raw {@code TypeLiteral}, the container treats it as a definition error.
     * <p>
     * The type argument to {@code TypeLiteral} may include type variables. When considering whether a type is in the set of expected types, different type variables with the same bounds will be considered equal.
```

We could add something like this to resolve the [discussion around type variables in the TCK test PR](https://github.com/jakartaee/cdi-tck/pull/679#discussion_r2789449015)

manovotn (approved):
I think it's ready to go now 👍 

@Azquelt please confirm it is OK with you as well and if so, feel free to hit the merge button :)

PR #941, opened 2026-02-05
Change wording in ATD/BBD#addAnnotatedType() to correspond with how archive trimming is worded.

### LINKED ISSUE #939: Clarify how bean archive trimming works for synthetic annotated types
[Archive trimming](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#trimmed_bean_archive) is fairly straightforward for class-based types but lacks more in-depth information for annotated types added via extensions.

An extension can register additional AT via `BeforeBeanDiscovery#addAnnotatedType()` (and same in `AfterTypeDiscovery`) and can register a class with no bean defining annotation.
What should happen in this case? Should the container throw away such AT or keep processing it?

Some notes:
* ATs added this way still fire `ProcessSyntheticAnnotatedType` hence giving users type to modify them before the container would decide what to do during trimming.
* Current wording is very vague and IMO doesn't lean towards any option.
* I think it is better to throw away such types as that's consistent with "normally discovered" types and treats them the same way. Extension authors can easily register an AT and add any BDA while at it.
  * OTOH if we go the other direction, extensions would bypass trimming completely leaving users little maneuvering space to affect it.

I think we could add a sentence pointing out that trimming effects both, discovered ATs as well as synthetically added ones.

manovotn:
Fixes #939 

Current wording of bean archive trimming is (emphasis mine):

> If an explicit bean archive contains the <trim/> element in its beans.xml file, types that don’t have either a bean defining annotation (as defined in [Bean defining annotations](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#bean_defining_annotations)) or any scope annotation, **are removed from the set of discovered types**.

This PR suggest a slight change to wording of `BeforeBeanDiscovery/AfterTypeDiscovery#addAnnotatedType` so that it is clearly implied that even added synthetic types are part of the 'set of discovered types' and hence subject to archive trimming.

Note that when it comes to BCE, there is [`ScannedClasses`](https://github.com/jakartaee/cdi/blob/ff86ca5cc3d86c219a65f3dcf22b6b85a56ac214/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/ScannedClasses.java#L14) which already uses this kind of wording:

> Allows adding additional classes to the set of types discovered during type discovery.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/spi/AfterTypeDiscovery.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/AfterTypeDiscovery.java
@@ -70,7 +70,7 @@ public interface AfterTypeDiscovery {
 
     /**
      * <p>
-     * Adds a given {@link AnnotatedType} to the set of types which will be scanned during bean discovery.
+     * Adds a given {@link AnnotatedType} to the set of types discovered during type discovery.
      * </p>
      *
      * <p>
@@ -85,16 +85,16 @@ public interface AfterTypeDiscovery {
      * </p>
      *
      * @param type The {@link AnnotatedType} to add for later scanning
-     * @param id the identifier used to distinguish this AnnotatedType from an other one based on the same underlying type
+     * @param id the identifier used to distinguish this AnnotatedType from another one based on the same underlying type
      * @throws IllegalStateException if called outside of the observer method invocation
      */
     public void addAnnotatedType(AnnotatedType<?> type, String id);
 
     /**
      * <p>
-     * Obtains a new {@link AnnotatedTypeConfigurator} to configure a new {@link AnnotatedType} and add it to the set of types
-     * which will be scanned during bean discovery at the end of the observer invocation. Calling this method multiple times
-     * will return a new AnnotatedTypeConfigurator.
+     * Obtains a new {@link AnnotatedTypeConfigurator} to configure a new {@link AnnotatedType} and add it to the set
+     * of types discovered during type discovery at the end of the observer invocation.
+     * Calling this method multiple times will return a new {@link AnnotatedTypeConfigurator}.
      * </p>
      *
      * <p>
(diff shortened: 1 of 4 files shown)


api/src/main/java/jakarta/enterprise/inject/spi/AfterTypeDiscovery.java
```diff
@@ -70,7 +70,7 @@ public interface AfterTypeDiscovery {
 
     /**
      * <p>
-     * Adds a given {@link AnnotatedType} to the set of types which will be scanned during bean discovery.
+     * Adds a given {@link AnnotatedType} to the set of discovered types.
```
Ladicek: I would prefer

```suggestion
     * Adds a given {@link AnnotatedType} to the set of types discovered during type discovery.
```

This makes it super clear and also aligns with what `ScannedClasses` says.

Ladicek: (Obviously, not just here but on all other places as well.)

manovotn: Done

Ladicek (approved):

api/src/main/java/jakarta/enterprise/inject/spi/AfterTypeDiscovery.java
```diff
@@ -70,7 +70,8 @@ public interface AfterTypeDiscovery {
 
     /**
      * <p>
-     * Adds a given {@link AnnotatedType} to the set of types which will be scanned during bean discovery.
+     * Adds a given {@link AnnotatedType} to the set of types discovered during type discovery.
+     * Added type will therefore be scanned during bean discovery.
```
Azquelt: Suggestion:
```suggestion
     * The added type will therefore be scanned during bean discovery.
```

I would also make this change everywhere else this sentence is used.

It is slightly funny to be adding this clarifying wording when the goal is to align with the wording on trimming which would in turn _prevent_ an added type from being scanned during bean discovery, but I think on balance that it's more helpful than not.

manovotn: TODO - we agreed to remove this sentence (all occurrences + `ScannedClasses`) during the CDI call.

PR #961, opened 2026-04-13
improve async handlers specification

Ladicek:
The situation where explicit configuration is needed is now clarified: it is not just when a target method matches multiple async handlers, but when multiple async handlers have the same async type.

Further, this commit makes the wording about service providers consistent across the spec: it's always "service provider declared in `META-INF/services`". This used to make no sense before Java 9, because there was no other way to declare service providers, but it actually saves us now in Java 9+. This is because there are _two_ ways to declare a service provider: in `META-INF/services` or in the module descriptor. We acknowledge that we currently require `META-INF/services` and declaring a service provider in the module descriptor is nonportable. We'll need to get back to it later.

With this, the section on async handlers gets rid of the sentence about the difference between provider method and provider constructor, because that only applies to service providers declared in the module descriptor. Service providers declared in `META-INF/services` always have a provider constructor.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildCompatibleExtension.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildCompatibleExtension.java
@@ -11,10 +11,8 @@
 package jakarta.enterprise.inject.build.compatible.spi;
 
 /**
- * Build compatible extensions are service providers for this interface, as defined in {@link java.util.ServiceLoader}.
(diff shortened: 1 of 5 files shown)


Ladicek:
The commit in this PR, as well as the PR description, used to mention a situation where more than one handler matches the target method. This was a mistake on my side. The behavior in this situation is actually already specified:

> If the target method matches exactly one async handler, the invoker becomes asynchronous.

So if the target method matches more than one async handler, it is _not_ asynchronous.

As mentioned previously, this is purely a theoretical concern that doesn't occur in practice.

manovotn (approved):
LGTM

PR #962, opened 2026-04-14
allow supporting Kotlin `suspend` functions in async handlers

Ladicek:
This is a draft for 2 reasons:

1. It includes the commit from #961, just because it's easier.
2. I'm not sure we want to do this. The async handler interface becomes more complex only to be able to accommodate Kotlin `suspend` functions, which is a trade off I'm not sure we're ready to make. Also I'm not terribly fond of the API design. In other words, there are downsides that perhaps outweigh the upside.

I'd be interested to hear your opinions.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildCompatibleExtension.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/BuildCompatibleExtension.java
@@ -11,10 +11,8 @@
 package jakarta.enterprise.inject.build.compatible.spi;
 
 /**
- * Build compatible extensions are service providers for this interface, as defined in {@link java.util.ServiceLoader}.
- * This means: they are classes that implement this interface, provide a {@code META-INF/services} file,
- * and satisfy all other service provider constraints. Additionally, build compatible extensions must not be beans
- * and must not be referred to by application code.
+ * Build compatible extensions are service providers for this interface declared in {@code META-INF/services}.
+ * Additionally, build compatible extensions must not be beans and must not be referred to by application code.
  * <p>
  * Build compatible extensions may define arbitrary {@code public}, non-{@code static}, {@code void}-returning methods
  * without type parameters, annotated with one of the extension annotations. Such methods are called extension methods.
--- a/api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
+++ b/api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
@@ -72,6 +72,25 @@ public interface AsyncHandler<T> {
     T transform(T original, Runnable completion);
 
     /**
+     * Called only on {@link ParameterType @ParameterType} async handlers, after the target
+     * method returns normally. Allows the async handler to detect synchronous completion
+     * of the asynchronous action based on the return value of the target method.
+     * <p>
+     * In some async protocols, the target method may complete synchronously without invoking
+     * the callback that was wrapped by {@link #transform(Object, Runnable) transform()}.
+     * This method should detect such synchronous completion and call {@code completion.run()}.
+     * <p>
+     * The default implementation does nothing. This is the correct behavior for most
+     * {@link ParameterType @ParameterType} async handlers; the only known situation when
+     * this method needs to be implemented is invoking Kotlin {@code suspend} functions.
+     *
+     * @param returnValue the return value of the target method
+     * @param completion the same completion action that was passed to {@code transform()}
+     */
+    default void checkReturnValue(Object returnValue, Runnable completion) {
+    }
+
+    /**
      * If an async handler is annotated {@code @ReturnType}, the target method matches when
      * the erasure of its return type is identical to the async type of the async handler.
      * <p>
--- a/spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
+++ b/spec/src/main/asciidoc/core/beanmanager_lite.asciidoc
@@ -67,7 +67,7 @@ If no qualifier is passed to `Instance.select()`, there is one required qualifie
 When `CDI.current()` is called, `getCDI()` method is called on `jakarta.enterprise.inject.spi.CDIProvider`.
 
 The `CDIProvider` to use may be set by the application or container using the `setCDIProvider()` method.
-If the `setCDIProvider()` has not been called, the service provider with highest priority of the service `jakarta.enterprise.inject.spi.CDIProvider` declared in META-INF/services is used.
+If the `setCDIProvider()` has not been called, the service provider with highest priority of the service `jakarta.enterprise.inject.spi.CDIProvider` declared in `META-INF/services` is used.
 The order of more than one `CDIProvider` with the same priority is undefined.
 If no provider is available an `IllegalStateException` is thrown.
 
--- a/spec/src/main/asciidoc/core/invokers.asciidoc
+++ b/spec/src/main/asciidoc/core/invokers.asciidoc
@@ -237,8 +237,7 @@ public interface AsyncHandler<T> {
 }
 ----
 
-An _async handler_ is a service provider of `jakarta.enterprise.invoke.AsyncHandler`.
-An async handler may not declare a provider method; it must declare a provider constructor.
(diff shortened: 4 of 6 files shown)


Azquelt:
Kotlin suspend functions feel like something it would be worth supporting, though I agree that I don't much like the API design either.

I'm not familiar with Kotlin suspend functions, but based entirely on [Coroutines under the hood](https://kt.academy/article/cc-under-the-hood), a Kotlin suspend function
- has a `Continuation` parameter
- is complete when it returns a result that isn't `COROUTINE_SUSPENDED`

So I assume to make this work with Invoker you'd want to:
- wrap the `Continuation` when the method is called
    - make `resumeWith` call the `completion` function when the method either returns a result or throws an exception
- if the method returns a value, it's complete and you need to run the `completion` function
- if the method returns `COROUTINE_SUSPENDED`, whatever you did earlier with the `Continuation` takes care of calling `completion`
- if the coroutine is resumed, the method gets called again and continues from where it was suspended but the invoker doesn't get called again

Based on that, I think your API change is sufficient to support Kotlin suspend functions, but it's very specific to their requirements. If you needed to pass some data from when the invoker was called to when it returns, for example, you couldn't do that.

Another option would be to go back to making the async handler effectively in interceptor as in https://github.com/jakartaee/cdi/issues/859#issuecomment-3397034146 which is a bit more flexible at the cost of complexity. We could keep the current design of how handlers are discovered and chosen. I would argue that if someone is writing an integration with an async framework, they can probably handle a bit of complexity.

Azquelt:
The more I think about it, the more I think the "problem" is that `Continuation.resumeWith` effectively wraps the function call when the suspended function is resumed, and our async handler design doesn't allow the original call to be wrapped in the same way.

Ladicek:
The "problem" with Kotlin coroutines is that, unlike any other async API, they can signal completion using 2 different means:

- if the `suspend` function returns `COROUTINE_SUSPENDED`, completion is later signaled through the `Continuation` parameter (asynchronous completion)
- if the `suspend` function returns something else, it becomes completed instantly _and the `Continuation` parameter is ignored_ (synchronous completion)

The funny thing is that we already support synchronously thrown exceptions, because that is a possibility that may (although shouldn't) occur with other async APIs as well. We just don't have a way to handle the two possibilities of completion.

Interceptor-like solution is certainly an option, although 1. it has higher overhead, 2. it looks exactly like invoker wrapper which I previously proposed and we agreed to postpone it :-)

The biggest issue I see with this PR is that the `checkReturnValue()` clearly feels like a bolt-on and not a natural part of the API. Maybe we should have:

```java
interface AsyncHandler<RETURN_TYPE, PARAMETER_TYPE> {
    PARAMETER_TYPE transformParameter(PARAMETER_TYPE original, Runnable completion);

    RETURN_TYPE transformReturnValue(RETURN_TYPE original, Runnable completion);
}
```

But it looks like unnecessary pain for `@ReturnValue` handlers, which the majority case.

So maybe we really should have 2 types, say:

```java
interface ReturnTypeAsyncHandler<T> {
    T transform(T original, Runnable completion);
}

interface ParameterTypeAsyncHandler<P, R> {
    P transformParameter(P original, Runnable completion);

    default R transformReturnValue(R original, Runnable completion) {
        return original;
    }
}
```

It is exactly the same API we already have (including the proposal in this PR), just more cohesive.

We can certainly go full interceptors, too, I'm just not a big fan.

Azquelt:
> Interceptor-like solution is certainly an option, although 1. it has higher overhead,

Fair, and if we give the user free reign to write their interceptor, it can't easily be rolled into a `MethodHandle` either.

> 2. it looks exactly like invoker wrapper which I previously proposed and we agreed to postpone it :-)

Kind of. The key difference is that we're giving the user a way to control exactly when dependent beans are destroyed, which I don't think was part of the invoker wrapper.

----

Thinking through other issues with having an interceptor-style handler:

- If the handler needs to process a parameter, the handler needs to be given enough information to know which parameter it will be modifying
  - In the general case it probably wants the `AnnotatedMethod` or `MethodInfo`
    - so that immediately makes writing a handler more difficult if you need to deal with both cases
  - You want to calculate whatever you need to know up front, so I think you need an instance per invoker, rather than a singleton
  - You then need to sort out providing this information to the handler in a build-time-friendly way
    - I can't remember enough about what can and can't be done at build time to know whether this would be painful

There's probably some things I'm missing. My feeling is that it could be done, but the API wouldn't end up being simple.

Given that, I'm happy to drop this suggestion, though I won't be surprised if we end up needing to add something like it in a future version. (I just feel like at some point we're going to encounter another issue like this where whatever more limited solution we come up with now won't quite meet the requirements.)

----

Back to your suggestion

> So maybe we really should have 2 types, say:
> 
> ```java
> interface ReturnTypeAsyncHandler<T> {
>     T transform(T original, Runnable completion);
> }
> 
> interface ParameterTypeAsyncHandler<P, R> {
>     P transformParameter(P original, Runnable completion);
> 
>     default R transformReturnValue(R original, Runnable completion) {
>         return original;
>     }
> }
> ```
> 
> It is exactly the same API we already have (including the proposal in this PR), just more cohesive.

I agree this seems more cohesive than the current PR and I do prefer this. We'd need to explain the lifecycle of the `ParameterTypeAsyncHandler` so that the user knows they can't store any state.

I'm unsure whether we should use both the parameter and return types to decide whether to apply a `ParameterTypeAsyncHandler`, or whether you'd still match just on the parameter type and require the handler to handle any return type.

Ladicek:
> I agree this seems more cohesive than the current PR and I do prefer this. We'd need to explain the lifecycle of the `ParameterTypeAsyncHandler` so that the user knows they can't store any state.

I'll try to rewrite the specification in this manner.

> I'm unsure whether we should use both the parameter and return types to decide whether to apply a `ParameterTypeAsyncHandler`, or whether you'd still match just on the parameter type and require the handler to handle any return type.

We should certainly still match only on the parameter type. Most implementations won't care about the return type/value at all, so I'm not sure we should even have a second type parameter for the return type.

We might even have something like:

```java
interface ReturnTypeAsyncHandler<T> {
    T transform(T original, Runnable completion);
}

interface ParameterTypeAsyncHandler<T> {
    T transformParameter(T original, Runnable completion);
}

interface ExtendedParameterTypeAsyncHandler<P, R> extends ParameterTypeAsyncHandler<P> {
    R transformReturnValue(R original, Runnable completion);
}
```

But in that case, we could even leave the `AsyncHandler` interface intact and have

```java
public interface AsyncHandler<T> {
    T transform(T original, Runnable completion);

    // only for `@ParameterType`, deployment error for `@ReturnType`
    public interface Extended<P, R> extends AsyncHandler<P> {
        R transformReturnValue(R original, Runnable completion);
    }
}
```

But that's almost the same as this PR :-)

Considering that vast majority of async handlers apply to return types, I think my original proposal of

```java
interface ReturnTypeAsyncHandler<T> {
    T transform(T original, Runnable completion);
}

interface ParameterTypeAsyncHandler<P, R> {
    P transformParameter(P original, Runnable completion);

    default R transformReturnValue(R original, Runnable completion) {
        return original;
    }
}
```

is reasonable and as mentioned above, I'll try to rewrite the spec in this manner.

Azquelt:
As discussed on the call, I think it the parameter handler interface should be defined like this, with `R` as a type variable of the method, not the interface:

```java
interface ReturnTypeAsyncHandler<T> {
    T transform(T original, Runnable completion);
}

interface ParameterTypeAsyncHandler<P> {
    P transformParameter(P original, Runnable completion);

    default <R> R transformReturnValue(R original, Runnable completion) {
        return original;
    }
}
```

Otherwise an implementation of the interface can declare that it only supports certain return types so we would have to match on the return type as well when deciding whether an async handler applies to a method.

Ladicek:
That's a good point, but adding a type parameter to the `transformReturnValue()` doesn't add much value. It feels like just using `Object` is simplest:

```java
interface ReturnTypeAsyncHandler<T> {
    T transform(T original, Runnable completion);
}

interface ParameterTypeAsyncHandler<T> {
    T transformParameter(T original, Runnable completion);

    default Object transformReturnValue(Object original, Runnable completion) {
        return original;
    }
}
```

Ladicek:
Superseded by #967, closing.

PR #966, opened 2026-04-21
improve the specification of `SyntheticInjections`

Ladicek:
This clarifies that `SyntheticBeanCreator` and `SyntheticBeanDisposer` behave similarly to producer/disposer methods also when it comes to injected `@Dependent` beans.

When a `@Dependent` bean is injected into a producer, it becomes a dependent object of the produced instance and so is destroyed when the produced instance is destroyed. The same occurs here: when a `@Dependent` bean is created by `SyntheticInjections` in the `SyntheticBeanCreator`, it becomes a dependent object of the synthetic bean instance and so is destroyed when the instance is destroyed.

When a `@Dependent` bean is injected into a disposer, it is destroyed when the disposer invocation completes. The same occurs here: when a `@Dependent` bean is created by `SyntheticInjections` in the `SyntheticBeanDisposer`, it is destroyed when the `SyntheticBeanDisposer.dispose()` invocation completes.

Producer and disposer are independent and so don't share injection points. The same occurs here: `SyntheticBeanCreator` and `SyntheticBeanDisposer` don't share `SyntheticInjections` and don't share instances for the same injection points.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanCreator.java
+++ b/api/src/main/java/jakarta/enterprise/inject/build/compatible/spi/SyntheticBeanCreator.java
@@ -38,6 +38,10 @@ public interface SyntheticBeanCreator<T> {
      * All injectable references looked up from {@code SyntheticInjections} have to previously be registered using
      * {@code SyntheticBeanBuilder.withInjectionPoint()}.
      * <p>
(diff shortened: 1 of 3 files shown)


manovotn (approved):
LGTM

Azquelt (approved):
:+1:

PR #967, opened 2026-04-23
respecify async handlers

Ladicek:
The API is quite different now: instead of annotations to specify whether a return type or parameter type should be inspected, there are nested interfaces in `AsyncHandler` that need to be implemented.

The `AsyncHandler.ReturnType` is the same as previous `AsyncHandler`, but `AsyncHandler.ParameterType` is more advanced and allows inspecting the return value for cases (like Kotlin `suspend` functions) that use two ways of signaling completion: return value in case of synchronous completion and parameter in case of asynchronous completion.

### DIFF
--- a/api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
+++ b/api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
@@ -10,98 +10,113 @@
 
 package jakarta.enterprise.invoke;
 
-import java.lang.annotation.ElementType;
-import java.lang.annotation.Retention;
-import java.lang.annotation.RetentionPolicy;
-import java.lang.annotation.Target;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
-import java.util.concurrent.Flow;
-
 /**
- * Includes logic for determining when an asynchronous action started by an invocation
- * of an asynchronous method completes. Service providers of this interface are called
- * <em>async handlers</em>. For the purpose of method invokers, a method is called
- * <em>asynchronous</em> if it matches an async handler.
+ * Determines when an asynchronous action started by an invocation of an asynchronous method
+ * completes. For the purpose of method invokers, a method is called <em>asynchronous</em> if it matches
+ * exactly one async handler. An <em>async handler</em> is either a {@linkplain ReturnType return type async handler}
+ * or a {@linkplain ParameterType parameter type async handler}. For an asynchronous method, the matching async handler
+ * is called during {@link Invoker#invoke(Object, Object[]) Invoker.invoke()}.
  * <p>
- * An <em>async type</em> of an async handler is the erasure of the type argument
- * to the {@code AsyncHandler} direct superinterface type.
+ * Async handlers must be stateless. No guarantees about their lifecycle are made.
  * <p>
- * An async handler must be annotated either {@link ReturnType} or {@link ParameterType}.
- * This annotation determines how methods are matched and how the async handler is used.
- * <p>
- * The {@link #transform(Object, Runnable) transform()} method of a matching async handler
- * is called exactly once by {@link Invoker#invoke(Object, Object[]) Invoker.invoke()}.
- * It takes an {@code original} asynchronous action and returns a transformation of it.
- * This transformation calls {@code completion.run()} exactly once, when the asynchronous
- * action completes.
- * <p>
- * For example, an async handler for {@link java.util.concurrent.CompletionStage} may look like:
- *
- * <pre>
- * &#064;ReturnType
- * public class CompletionStageAsyncHandler&lt;T&gt; implements AsyncHandler&lt;CompletionStage&lt;T&gt;&gt; {
- *     public CompletionStage&lt;T&gt; transform(CompletionStage&lt;T&gt; original, Runnable completion) {
- *         return original.whenComplete((value, error) -> {
- *             completion.run();
- *         });
- *     }
- * }
- * </pre>
- *
- * The CDI container must provide an async handler for the return types of
+ * The CDI container provides built-in return type async handlers for
  * {@link java.util.concurrent.CompletionStage CompletionStage},
- * {@link java.util.concurrent.CompletableFuture CompletableFuture} and
- * {@link Flow.Publisher Flow.Publisher}.
+ * {@link java.util.concurrent.CompletableFuture CompletableFuture}
+ * and {@link java.util.concurrent.Flow.Publisher Flow.Publisher}.
  *
- * @param <T> the type that represents the asynchronous action
  */
-public interface AsyncHandler<T> {
+public interface AsyncHandler {
     /**
-     * Takes the {@code original} asynchronous action and returns a variant of it, transformed
-     * such that {@code completion.run()} is called on completion. Note that {@code completion.run()}
-     * must be called exactly once.
+     * A <em>return type async handler</em> is a service provider for this interface declared
+     * in {@code META-INF/services}. The <em>async type</em> of a return type async handler is the erasure
+     * of the type argument to the {@code AsyncHandler.ReturnType} direct superinterface type.
      * <p>
-     * It is recommended that {@code completion.run()} is called <em>before</em> completion
-     * is propagated to the caller.
+     * A target method matches a return type async handler when the erasure of the method's return type
+     * is identical to the async type of the async handler.
      *
-     * @param original the original asynchronous action
-     * @param completion action that must be executed when the asynchronous action completes
-     * @return the transformed async action
+     * @see #transform(Object, Runnable)
+     * @param <T> the async type of the handler
      */
-    T transform(T original, Runnable completion);
-
-    /**
-     * If an async handler is annotated {@code @ReturnType}, the target method matches when
-     * the erasure of its return type is identical to the async type of the async handler.
-     * <p>
-     * In such case, the {@code AsyncHandler.transform()} method is called after the target
-     * method returns. The original return value is passed to {@code transform()} as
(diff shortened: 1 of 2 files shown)


Ladicek:
I don't have an updated TCK just yet, but I do have an implementation (which passes all tests I have in ArC), so I feel fine with submitting this for review. I will follow up on the TCK front early next week.

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -265,5 +291,5 @@ The CDI container must include async handlers for the following return types:
 If two async handlers have the same async type, by default a deployment problem occurs.
 Containers are required to provide implementation-defined means of configuring an async handler that should be used for any given async type.
```
Azquelt:
```suggestion
Containers are required to provide an implementation-defined means of configuring the async handler that should be used for a particular async type.
```

The original wording could be read as configuring an async handler that should be used for _all_ async types.

api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
```diff
@@ -10,98 +10,108 @@
...
+         * Called once by {@link Invoker#invoke(Object, Object[]) Invoker.invoke()}, before the target method
+         * is invoked.
+         * <p>
+         * Takes the {@code original} argument value for the async parameter and returns a variant of it,
+         * transformed such that {@code completion.run()} is called on completion. Note that
+         * {@code completion.run()} must be called exactly once. The result is passed to the target method
+         * instead of the {@code original}.
+         * <p>
+         * It is recommended that {@code completion.run()} is called <em>before</em> completion
+         * is propagated to the caller.
```
Azquelt: I suggest changing it like this because there are now two methods which may both work together to ensure `completion.run()` is run.

```suggestion
         * Takes the {@code original} argument value for the async parameter and returns a variant of it,
         * transformed such that {@code completion.run()} is called on completion. The result is passed to the
         * target method instead of the {@code original}.
         * <p>
         * The net effect of this method and {@link #transformReturnValue(Object, Runnable)} must be that
         * {@code completion.run()} is called exactly once, after the asynchronous action completes. It is
         * recommended that {@code completion.run()} is called <em>before</em> completion is propagated to the
         * caller.
```

api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
```diff
@@ -10,98 +10,108 @@
...
+        /**
+         * Called once by {@link Invoker#invoke(Object, Object[]) Invoker.invoke()}, after the target method
+         * returns. When the target method throws an exception synchronously, this method is not called.
+         * <p>
+         * Takes the {@code original} return value and returns a variant of it, transformed
+         * such that {@code completion.run()} is called on completion. Note that {@code completion.run()}
+         * must be called exactly once. The result is returned to the invoker caller instead of the {@code original}.
+         * <p>
+         * It is recommended that {@code completion.run()} is called <em>before</em> completion
+         * is propagated to the caller.
```
Azquelt: - clarify that the two methods work together to run `completion.run()` once
- indicate that a transformation may or may not be required here
- state the default implementation behaviour explicitly

```suggestion
         * Takes the {@code original} return value and returns a variant of it, performing any transformation
         * necessary such that that {@code completion.run()} is called on completion. The result is returned
         * to the invoker caller instead of the {@code original}.
         * <p>
         * The net effect of this method and {@link #transformArgument(Object, Runnable)} must be that
         * {@code completion.run()} is called exactly once, after the asynchronous action completes. It is
         * recommended that {@code completion.run()} is called <em>before</em> completion is propagated to the
         * caller.
         * <p>
         * The default implementation returns {@code original} directly.
```

api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
```diff
@@ -10,98 +10,108 @@
...
-
 /**
- * Includes logic for determining when an asynchronous action started by an invocation
- * of an asynchronous method completes. Service providers of this interface are called
- * <em>async handlers</em>. For the purpose of method invokers, a method is called
- * <em>asynchronous</em> if it matches an async handler.
- * <p>
- * An <em>async type</em> of an async handler is the erasure of the type argument
- * to the {@code AsyncHandler} direct superinterface type.
+ * Contains logic that determines when an asynchronous action started by an invocation of an asynchronous method
```
Azquelt:
```suggestion
 * Determines when an asynchronous action started by an invocation of an asynchronous method
```

api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
```diff
@@ -10,98 +10,108 @@
...
- * @param <T> the type that represents the asynchronous action
  */
-public interface AsyncHandler<T> {
+public interface AsyncHandler {
     /**
-     * Takes the {@code original} asynchronous action and returns a variant of it, transformed
-     * such that {@code completion.run()} is called on completion. Note that {@code completion.run()}
-     * must be called exactly once.
+     * A <em>return type async handler</em> is a service provider for this interface declared
+     * in {@code META-INF/services}. An <em>async type</em> of a return type async handler is the erasure
```
Azquelt:
```suggestion
     * in {@code META-INF/services}. The <em>async type</em> of a return type async handler is the erasure
```

api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
```diff
@@ -10,98 +10,108 @@
...
-     * It is recommended that {@code completion.run()} is called <em>before</em> completion
-     * is propagated to the caller.
+     * A target method matches a return type async handler when the erasure of the method's return type
+     * is identical to the async type of the async handler.
      *
-     * @param original the original asynchronous action
-     * @param completion action that must be executed when the asynchronous action completes
-     * @return the transformed async action
+     * @see #transform(Object, Runnable)
+     * @param <T> the type that represents the asynchronous action
```
Azquelt:
```suggestion
     * @param <T> the async type of the handler
```

api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
```diff
@@ -10,98 +10,108 @@
...
-     * <p>
-     * If an async handler is annotated {@code @ParameterType}, it must not be annotated
-     * {@code @ReturnType}, otherwise deployment problem occurs.
+     * A target method matches a parameter type async handler when it declares exactly one parameter
+     * whose type's erasure is identical to the async type of the async handler. This parameter is called
+     * the <em>async parameter</em>.
+     *
+     * @see #transformArgument(Object, Runnable)
+     * @see #transformReturnValue(Object, Runnable)
+     * @param <T> the type that represents the asynchronous action
```
Azquelt:
```suggestion
     * @param <T> the async type of the handler
```

spec/src/main/asciidoc/core/invokers.asciidoc
```diff
@@ -232,31 +232,57 @@ Method invokers have special support for asynchronous methods when it comes to d
...
+The target method matches a parameter type async handler if it declares exactly one parameter whose type's erasure is identical to the async type of the async handler; this parameter is called the _async parameter_.
 
-Each async handler must be annotated either `@AsyncHandler.ReturnType` or `@AsyncHandler.ParameterType`, otherwise deployment problem occurs.
-If an async handler is annotated `@AsyncHandler.ReturnType`, the target method matches the async handler if the erasure of its return type is identical to the async type of the async handler.
-If an async handler is annotated `@AsyncHandler.ParameterType`, the target method matches the async handler if it declares exactly one parameter whose type's erasure is identical to the async type of the async handler; this parameter is called an _async parameter_.
+NOTE: The provider-configuration file in `META-INF/services` must be named `jakarta.enterprise.invoke.AsyncHandler$ReturnType` or `\...AsyncHandler$ParameterType`, because the `ServiceLoader` requires the file name to be a _fully qualified binary name_ of the service.
```
Azquelt:
```suggestion
NOTE: The provider-configuration file in `META-INF/services` must be named `jakarta.enterprise.invoke.AsyncHandler$ReturnType` or `jakarta.enterprise.invoke.AsyncHandler$ParameterType`, because the `ServiceLoader` requires the file name to be a _fully qualified binary name_ of the service.
```

It isn't many more characters to spell it out and I think it's clearer.

Azquelt:
This looks good :+1: I've added some suggestions that I think improve clarity.

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: Fair point.

api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
Ladicek: Makes sense.

api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
Ladicek: Makes sense.

api/src/main/java/jakarta/enterprise/invoke/AsyncHandler.java
Ladicek: Nice!

spec/src/main/asciidoc/core/invokers.asciidoc
Ladicek: It looked uglier when rendered, but I agree it's a little better.

Azquelt (approved):
Thanks

Ladicek:
Squashed the commits.

PR #973, opened 2026-05-14
fix one corner case of the async handlers specification

Ladicek:
A parameter type async handler can, in theory, signal completion while the method is still running. Before this commit, the specification required that instances of `@Dependent` looked up beans are destroyed at that moment in time, which would be wrong, especially when the method is called on a looked up instance of a `@Dependent` bean.

This commit requires that instances of `@Dependent` looked up beans are destroyed only after the target method returns, even if completion has already been signaled.


manovotn (approved):

Azquelt (approved):
Thanks.

PR #995, opened 2026-07-02
Fix race condition in calculating configuredProvider in CDI.current

### LINKED ISSUE #994: Race condition in CDI.getCDIProvider()
**Describe the bug**
In `api/src/main/java/jakarta/enterprise/inject/spi/CDI.java`, the method `getCDIProvider()` has a potential race condition when `configuredProvider` is `null` (or unusable).

The code currently has a lock that is used to synchronize calculation of the `discoveredProviders` collection, but before it enters that lock, it overwrites `configuredProvider` with a `null` value.  Then it calculates discoveredProviders (if necessary) and then does a stream/filter/findFirst over the result to calculate a new `configuredProvider` value. No locking is done around checking/setting this `configuredProvider` value when a new one is being calculated.

If the `configuredProvider` is null and multiple threads call into `CDI.current()` simultaneously, it's possible that one thread will erroneously return `null` because the result it calculated has been overwritten by the second thread.

Copy/pasting the relevant code:
```
        configuredProvider = null;
        // Discover providers and cache
        if (discoveredProviders == null) {
            synchronized (lock) {
                if (discoveredProviders == null) {
                    findAllProviders();
                }
            }
        }
        configuredProvider = discoveredProviders.stream()
                .filter(CDI::checkProvider)
                .findFirst().orElseThrow(() -> new IllegalStateException("Unable to access CDI"));
        return configuredProvider;
```

Long-story-short, if thread 1 sets its configured provider in the 2nd-last statement (the stream/filter operation), then the 2nd thread runs its `configuredProvider = null;` at the top, the first thread will then return `null` when it runs its next line of code.

**Expected behavior**
CDI.current() should be safe for multiple threads to enter at the same time. If `configuredProvider` is null or needs to be recalculated, a lock is used to prevent multiple threads interfering with each other's processing. If not a lock, then at least don't temporarily null-out the configuredProvider variable while it's being calculated.  Just directly assign the value with the new calculated provider after a result is available.

**Additional context**
We ran into this in Wildfly 39 in a complex EAR deployment.  We have a JDBC proxy driver that does a lookup with `CDI.current().select(A.class).get()` to look up the current user that executed a query. We're making changes to the code to avoid `CDI.current()`, but this race condition was leading to random failures in application code triggered by a NullPointerException when `getCDIProvider().getCDI()` was called (because `getCDIProvider()` had returned a null value).

aaronwatry-advarrapublic:
Moves the lock acquisition out to and around where configuredProvider is set/reset.

Since the lock is acquired unconditionally, change the double checking of discoveredProviders being null to a single check since it's inside the syncronized section now.

Fixes #994

### DIFF
--- a/api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
+++ b/api/src/main/java/jakarta/enterprise/inject/spi/CDI.java
@@ -85,7 +85,6 @@ public abstract class CDI<T> implements Instance<T> {
                 throw e;
             }
         }
-        configuredProvider = null;
         // Discover providers and cache
         if (discoveredProviders == null) {
             synchronized (lock) {


Ladicek:
Thanks for the bug report! The proper fix is, I think, just a CAS loop. I'll submit a PR. Thanks again, though!

